PR #51, opened 2022-09-12
Enable override attribute on Netty dependencies

sdeleuze:
@gradinac Could you please review this PR? We should also probably update the repository version to `0.2.0-SNAPSHOT` in order to express that the format has changed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,7 @@ It should contain the following entries:
     ]
     ```
     `metadata-version` key specifies the subdirectory where metadata for tested versions "lives".
+    `override` flag allows to express the intent to exclude outdated builtin metadata when set to `true`.
     So, the metadata for `org.example:library:0.0.1` and `org.example:library:0.0.2` is located at `metadata/org.example/library/0.0.1`.
 
    Make sure that each supported version is listed in `tested-version`, as that value is used in build tools to match metadata to a specific library.
--- a/metadata/index.json
+++ b/metadata/index.json
@@ -40,6 +40,10 @@
     "module": "org.apache.tomcat.embed:tomcat-embed-core"
   },
   {
+    "directory": "io.netty/netty-common",
+    "module": "io.netty:netty-common"
+  },
+  {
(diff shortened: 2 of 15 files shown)


sdeleuze:
I had to include empty metadata to override Netty configurations, should I add empty tests or do we have a way to exclude those for the checks?

lazar-mitrovic:
I would personally prefer if we had explicit tests for each part of Netty (especially since we are now excluding existing config, and we should check for regressions).

Alternatively we could link it to the existing `netty-transport` test or some dummy test, but I'm afraid that would set a bad precedent.

lazar-mitrovic:
- [ ] Prior to merging this we need to implement `override` support for TCK in order to be able to properly test metadata.
This shouldn't be hard, fix shouldn't be more than couple of lines long. I'll take a look at that this week.

sdeleuze:
To add some tests, don't we need NBT 0.9.14 to be released first? Isn't it a kind of chicken and egg problem?

lazar-mitrovic:
> To add some tests, don't we need NBT 0.9.14 to be released first? Isn't it a kind of chicken and egg problem? If so we can maybe more forward with empty tests for now and add proper testing as a second step after NBT 0.9.14 and metadata repo 0.2.0 releases.

Well, we already have a different code path for metadata location detection [here](https://github.com/oracle/graalvm-reachability-metadata/blob/bb36403180544e182008266edc9340f37c1a4fcb/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle#L72). We aren't using metadata support artifact here since it isn't what we are testing - we are just using standard primitives and relying on TCK to inject the right values for us, and thus we don't rely on NBT version at all.

sdeleuze:
Ok please let me know when you have done the required update, I can then add related tests (unless you want to do it).

sdeleuze:
Rebased on top of the commit mentioned in [this comment](https://github.com/oracle/graalvm-reachability-metadata/pull/52#issuecomment-1250327355) and after various tries, I decided to skip tests for empty metadata.

Issue #62, opened 2022-09-23
Introduce a `default-for` attribute


sdeleuze:
`latest` attribute seems not robust enough since it does not manage any notion of semantic versioning.

For example,  I am going to add metadata for `ch.qos.logback:logback-classic 1.4.1` with the `latest` flag set to `true`. Once Logback `1.2.12` will be out and used by somebody, native compilation will be broken due to the default to the latest if we just have metadata for `1.2.11`.

As discussed with @melix, we could introduce a `default-for` attribute with a regexp value used to match the version of the dependency. `latest` would translate to `"default-for": ".*"`, and we could document and use regexp for matching the same major + major version, which would be the most common use case.

sdeleuze:
Related NBT PR is https://github.com/graalvm/native-build-tools/pull/432.

PR #90, opened 2022-10-26
JSON schema for the Native Libraries Support repository

dnestoro:
## What does this PR do?
Specifies schema for JSON that will be used as an input for website libraries list


supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+      "artifact": "mysql",
+      "minimal_version": "8.0.29",
+      "metadata_location": "https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata/mysql",
+      "support_level": "Community Supported"
+    },
+    {
+      "artifact": "io.netty:netty-transport",
+      "minimal_version": "4.1.76.Final",
+      "maximal_version": "4.1.80.Final",
+      "metadata_location": "https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata/io.netty/netty-transport",
```
vjovanov, line 60: Should we use a list here? Also, how does it look when it is not in the repo?

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
+{
```
vjovanov, line 1: What do we do for libraries whose configuration depends on the usage? For example, in netty channel handlers written by the user need to be annotated.

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+      "minimal_version": "8.0.29",
+      "metadata_location": "https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata/mysql",
+      "support_level": "Community Supported"
+    },
+    {
+      "artifact": "io.netty:netty-transport",
+      "minimal_version": "4.1.76.Final",
+      "maximal_version": "4.1.80.Final",
+      "metadata_location": "https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata/io.netty/netty-transport",
+      "support_level": "Community Supported"
```
vjovanov, line 61: I would compute this field based on the metadata location and the level of testing. And introduce a field for the level of testing (untested, partially, fully).

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+      "artifact",
+      "minimal_version",
+      "metadata_location",
+      "support_level"
+    ],
+    "properties": {
+      "artifact": {
+        "type": "string",
+        "default": "",
+        "title": "The name of the metadata"
```
vjovanov, line 19: ?

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+      },
+      "minimal_version": {
+        "type": "string",
+        "default": "",
+        "title": "Minimal supported version"
```
vjovanov, line 24: supported -> for which this entry applies.

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+      "maximal_version": {
+        "type": "string",
+        "default": "",
+        "title": "Maximal supported version",
+        "description": "If not defined, it is assumed to be supported for all versions starting from the minimum_version"
+      },
+      "metadata_location": {
+        "type": "string",
+        "default": "",
+        "title": "Web location of provided metadata"
```
vjovanov, line 35: Web location -> Web URL

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+  "items": {
+    "type": "object",
+    "required": [
+      "artifact",
+      "minimal_version",
+      "metadata_location",
+      "support_level"
+    ],
+    "properties": {
+      "artifact": {
```
fniephaus, line 16: I wonder whether we can (re)use some ID/format here that can be used to identify the artifact in both Maven and Gradle. Maybe @melix has an idea for this?

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+          "Unsupported",
+          "Community Supported",
+          "Supported"
+        ],
+        "default": "Unsupported",
+        "title": "Metadata support level"
+      }
+    }
+  },
+  "examples": [
```
fniephaus, line 49: I don't think the example should include metadata from this repo because metadata from this repo should not be mentioned in this file at all (otherwise, we'll have to keep it in sync with the actual metadata under `metadata/`).

Can we use something representative like [netty](https://github.com/netty/netty/tree/main/common/src/main/resources/META-INF/native-image)?

supported-libraries-schema.json
fniephaus: Do you mean libraries/frameworks that generate configuration based on their annotations? I don't see a reason why those cannot be listed in here?

fniephaus, line 1: As a user seeing `supported-libraries-schema.json`, I'd think this is about the libraries supported by this repo. Also, I think we should include frameworks as well. How about we rename this to `library-and-framework-support-schema.json` and the data is stored in `library-and-framework-support.json`?

supported-libraries-schema.json
```diff
@@ -0,0 +1,64 @@
...
+      "metadata_location",
+      "support_level"
+    ],
+    "properties": {
+      "artifact": {
+        "type": "string",
+        "default": "",
+        "title": "The name of the metadata"
+      },
+      "minimal_version": {
```
fniephaus, line 21: I'd add a `description` key inbetween `artifact` and `minimal_version`, so that we can add a short text that says what the library or framework can do.

supported-libraries-schema.json
fniephaus: Agreed. What if we add a `tests` key (or something like that) after `metadata_url` like this:
```js
// ...
"tests": {
  "level": "fully",
  "url": "https://github.com/open-telemetry/opentelemetry-java/blob/21d55971b76c8f73343094270db0704c0b8ae4a3/.github/workflows/build-graal.yml",
},
//...
```

The calculation could be as simple as this:
```python
def calculate_support_level(entry):
  if entry['tests']['level'] == 'fully':
    return SUPPORTED
  if entry['metadata_url'] != '':
      return COMMUNITY_SUPPORTED
  else:
    return UNSUPPORTED
```

supported-libraries-schema.json
fniephaus: > Those libraries can be Supported without testing I believe.

Even for libraries and frameworks that don't need metadata, I think testing against GraalVM NI should be a requirement to be listed in here. If it's not tested and has no metadata, it's not supported.

supported-libraries-schema.json
dnestoro: We will update examples once we have clear idea of how to use all properties

supported-libraries-schema.json
sdeleuze: Can't we just use `$groupid:$artifactId`?

supported-libraries-schema.json
sdeleuze: Maybe additional metadata like a name that make sense to display can be retrieve from Maven Central based on the groupId, artifactId and version?

supported-libraries-schema.json
sdeleuze: `maximal_version` will be hardly maintainable. Regexp as suggested in https://github.com/oracle/graalvm-reachability-metadata/issues/62 could be an option.

library-and-framework-support-schema.json
```diff
@@ -0,0 +1,76 @@
...
+          "Partially tested",
+          "Fully tested"
+        ],
+        "default": "Untested",
+        "title": "Library level of testing"
+      },
+      "support_level": {
+        "type": "string",
+        "enum": [
+          "Unsupported",
```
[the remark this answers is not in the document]
vjovanov: Yeah, I would ditch the whole field.

library-and-framework-support-schema.json
```diff
@@ -0,0 +1,111 @@
+{
```
fniephaus, line 1: Maybe we should rename the file to `library-and-framework-list-schema.json` or `library-and-framework-overview-schema.json`. @vjovanov @wirthi?

library-and-framework-support-schema.json
```diff
@@ -0,0 +1,111 @@
...
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
+  "$id": "This will be populated with actual link once uploaded",
+  "title": "Schema for the supported libraries list JSON",
```
fniephaus, line 4:
```suggestion
  "title": "Schema for the 'library-and-framework-list.json'",
```
or whatever filename we agree on.

library-and-framework-support-schema.json
```diff
@@ -0,0 +1,111 @@
...
+      "artifact",
+      "details"
+    ],
+    "properties": {
+      "artifact": {
+        "type": "string",
+        "pattern": "^[a-zA-Z0-9.-]+:[a-zA-Z0-9.-]+$",
+        "default": "",
+        "title": "The name of the artifact",
+        "description": "The artifact name should be provided in following format: groupId:artifactId"
```
fniephaus, line 19:
```suggestion
        "description": "The artifact name in the groupId:artifactId format"
```

library-and-framework-support-schema.json
```diff
@@ -0,0 +1,111 @@
...
+      },
+      "description": {
+        "type": "string",
+        "default": "",
+        "title": "Short text that says what the library or framework can do"
```
fniephaus, line 24:
```suggestion
        "title": "Short description of the library or framework"
```

library-and-framework-support-schema.json
```diff
@@ -0,0 +1,111 @@
...
+            "minimal_version": {
+              "type": "string",
+              "default": "",
+              "title": "Minimal version for which this entry applies"
+            },
+            "maximal_version": {
+              "type": "string",
+              "default": "",
+              "title": "Maximal version for which this entry applies",
+              "description": "If not defined, it is assumed to be supported for all versions starting from the minimum_version"
```
fniephaus, line 48:
```suggestion
              "description": "Maximum version for which this entry applies. If not defined, it is assumed to be supported for all versions starting from the minimum_version"
```

library-and-framework-support-schema.json
```diff
@@ -0,0 +1,111 @@
...
+          "tests_locations": [
+            "https://github.com/netty/netty/actions"
+          ],
+          "test_level": "Fully tested"
+        },
+        {
+          "minimal_version": "2.0",
+          "maximum_version": "4.0",
+          "metadata_locations": [
+            "https://github.com/netty/netty/tree/main/common/src/main/resources/META-INF/native-image"
```
fniephaus, line 101:
```suggestion
            "https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata/io.netty"
```

fniephaus (changes requested):

library-and-framework-support-schema.json
vjovanov: LGTM

wirthi: or `tested-libraries-schema.json`?

library-and-framework-list-schema.json
```diff
@@ -0,0 +1,111 @@
...
+              "title": "Web URLs to tests (sources, CI dashboards/configurations, ...)"
+            },
+            "test_level": {
+              "type": "string",
+              "enum": [
+                "Untested",
+                "Community tested",
+                "Fully tested"
+              ],
+              "default": "Untested",
```
fniephaus, line 75: Considering the JSON file is meant for non-human consumption, maybe these should look more like symbols (`untested`, `community-tested`, `fully-tested`). At least, they should probably be lower-cased, and don't forget to update the example below.

fniephaus (approved):
LGTM.
Please feel free to incorporate https://github.com/oracle/graalvm-reachability-metadata/pull/90/files#r1011293794, but I'll leave that up to you :)

PR #92, opened 2022-10-27
Rework TCK build plugins

melix:
The current implementation was heavily relying on static state to share data between build scripts. While this works, this isn't a good practice because it relies on accidental state being available, and not Gradle idiomatic.

This also solves a number of issues with paths being eagerly resolved, and the (ab)use of `Path` in APIs (when in Gradle we should use the directory/file providers).

This also makes it quite difficult to follow what is setting what, and what is an input of a task. Instead, plugins should be applied, exposing data in a reliable way via project extensions.

There are still occurrences left in some plugins, which may be fixed in a later PR. For example, all tested projects use the same kind of build scripts, but incorrectly eagerly resolve the `libraryVersion` property.

Therefore this is _only_ a small step in cleaning things up.

## Checklist before merging


melix:
Failure seems unrelated to those changes

gradinac (approved):
I'm a bit rusty with Gradle, but overall the changes look okay to me :)

melix:
Thanks, let me try to merge the changes with master.

Issue #133, opened 2022-12-01
In `CONTRIBUTING.md` point out that PR needs to remove GraalVM reachability metadata related to MBeans


linghengqian:
**Is your feature request related to a problem? Please describe.**

- Yes, refer to 
    - https://github.com/oracle/graalvm-reachability-metadata/pull/113 
    - https://github.com/oracle/graalvm-reachability-metadata/pull/124
    - https://github.com/oracle/graalvm-reachability-metadata/pull/122

**Describe the solution you'd like**

- In `CONTRIBUTING.md` point out that PR needs to remove GraalVM reachability metadata related to MBeans. The following classes seem to cause a lot of misunderstanding, and do not affect nativeTest. The problem is that GraalVM Native Build Tools' Conditional Agent Mode picks up these classes easily.
    - `java.lang.management.**`
    - `jdk.management.**`
    - `com.sun.management.**`
    - `sun.management.**`
    - `javax.management.**`

**Describe alternatives you've considered**

- Null.

**Additional context**

- I did some possible related tests at https://github.com/linghengqian/graalvm-trace-metadata-smoketest/issues/1 .

linghengqian:
- I work on this issue. 


- Of course I only work with documents, because to be honest I am a Gradle novice and not sure how to write regular tasks for Github Action.

linghengqian:
- According to https://github.com/oracle/graalvm-reachability-metadata/pull/135, let's move to https://github.com/oracle/graal/issues/6322 to continue exploring.

PR #135, opened 2022-12-02
Point out that PR needs to remove GraalVM reachability metadata related to MBeans

linghengqian:
## What does this PR do?

- Fixes https://github.com/oracle/graalvm-reachability-metadata/issues/133 .

## Checklist before merging


CONTRIBUTING.md
```diff
@@ -122,6 +122,19 @@ output if necessary:
 ./gradlew check
 ```
 
+All submitted GraalVM reachability metadata, if the MBean-related part is unnecessary, should actively remove the
```
vjovanov, line 125: We should fix the root cause in Native Image. Could you be more precise about what is failing if we introduce MBean metadata?

linghengqian: - As an additional topic, can I bring the following reference links to this document, namely https://github.com/oracle/graalvm-reachability-metadata/pull/162 , https://github.com/oracle/graalvm-reachability-metadata/pull/161 and https://github.com/oracle/graalvm-reachability-metadata/pull/113 ?

linghengqian: - I tried to describe this process in the latest commit. But since this is currently only happening on the Spring Boot side, I'm not too sure about the more specific behavior. 

- I haven't personally investigated AOT on Spring Boot 3.x yet. Because I'm blocked by other behavior at https://github.com/oracle/graalvm-reachability-metadata/pull/154.

linghengqian (commented):
- @vjovanov I never reproduced this issue reported by the Spring team, which seems to have to use Spring Boot to write a relatively complex Example. 

- @wilkinsona I wonder if there are some reproducible unit tests on the Spring side?

wilkinsona:
I don't think there's anything Spring or Spring Boot-specific about the problem. It will be harder to reproduce now after the corrections to the metadata. Please open the requested GraalVM issue and I'll help out there as time permits. IMO, linking to https://github.com/spring-projects/spring-boot/issues/33210 would be a good start.

wilkinsona:
I should also have said that, while I agree that this should be fixed on the Graal side, I don't think metadata contributions should include anything for types like `com.sun.management.GarbageCollectorMXBean` which they never even touch irrespective of whether or not that cause a failure. If this PR has been declined, what alternative measures can we put in place to prevent future metadata updates causing problems or being too broad in scope?

vjovanov:
We have a check now that makes sure that all conditions in the metadata are related to the library in question. This way, all unintentional metadata additions of this kind should be avoided.

wilkinsona:
Great stuff! Thank you.

linghengqian:
- I still can't create a minimal reproducible Example, so I just opened https://github.com/oracle/graal/issues/6322.

PR #146, opened 2022-12-08
Add currently supported and tested Spring libraries

bclozel:
This commit adds the Spring Boot starters that are known to be
compatible with GraalVM Native and are tested in the Spring CI with
dedicated smoke tests.

Instead of listing all Spring artifacts that are known to work, we are
only listing the Spring Boot starters and optional artifacts that
developers add to their Spring Boot projects.

Spring Cloud projects are not listed in this commit as they are not
released yet on Maven Central.


library-and-framework-list.json
```diff
@@ -138,5 +138,537 @@
...
       }
     ]
+  },
+  {
+    "artifact": "org.springframework.boot:spring-boot-starter-actuator",
+    "description": "Starter for using Spring Boot's Actuator which provides production ready features to help you monitor and manage your application",
+    "details": [
+      {
+        "minimum_version": "3.0.0",
+        "metadata_locations": [],
```
dnestoro, line 148: Why the metadata locations are not provided here? I believe we need them, since our schema requires it.

bclozel: I've just updated the PR to add `"medata_locations"` entries. I've also rebased on top of the current master.

PR #148, opened 2022-12-12
Add support for grpc-netty

dsyer:
- [x] I have properly formatted metadata files (see [CONTRIBUTING](/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md) document)
- [x] I have added thorough tests. (see [this](/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#Tests))


amitbansal26:
When this will be merged ?

gradinac:
Hey @dsyer! :) Sorry for a late answer on https://github.com/netty/netty/pull/13039#issuecomment-1348464623 - I'll continue the posts here as it's directly related to this PR.
I don't think we currently have a way to use existing metadata from other libraries in tests for a particular library. In this case, it would be best if we could use the existing metadata, but I think we can have the required netty metadata manually specified for now.
Regarding https://github.com/netty/netty/pull/13039#issuecomment-1348465721, that's a very good question. It would not make the reflection registration redundant - if a type X is reachable, we do not register anything from type X for reflection by default. In this way, we tell native-image the following: if this type does end up in the image, also register the following for it.

gradinac (approved):

dsyer:
So how do we get this merged? We need another approval (from whom)? Or something else?

dnestoro:
@dsyer sorry for the delay, we were on holidays so we didn't merge anything these days.

PR #172, opened 2023-01-13
Add null check in diffCordinates function

dnestoro:
## What does this PR do?

As reported in [this](https://github.com/oracle/graalvm-reachability-metadata/actions/runs/3911262846/jobs/6685813930) gate there is an attempt to create a stream over the element which is null(we don't have null check for element we are trying to get from map). @melix please check if the fix can break something else in your logic here, since you are more familiar with this part of the code.


melix (approved):
I am not particularly familiar with this code, actually (it was written by Lazar) but the change LGTM

PR #203, opened 2023-01-25
Fix Not Supported Java Version Bug

dnestoro:
## What does this PR do?
Fix Ci `not supported Java version` error.

### DIFF
--- a/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
+++ b/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
@@ -86,8 +86,10 @@ if (project.hasProperty("baseCommit")) {
 }
 
 def matrixDefault = [
-        "version"     : ["22.3.0", "dev"], // TODO: Add other supported versions
-        "java-version": ["17"], // TODO: Add "17", "19"
+        "versions" : [
+//                [ "graalvm": "dev", "java": "dev" ], // TODO: uncomment this once we are ready to run tests on JDK19
+                [ "graalvm": "latest", "java": "17"]
+        ],
         "os"          : ["ubuntu-latest"] // TODO: Add support for "windows-latest", "macos-latest"
 ]
 
--- a/.github/workflows/test-all-metadata.yml
+++ b/.github/workflows/test-all-metadata.yml
@@ -32,7 +32,7 @@ jobs:
           ./gradlew generateMatrixMatchingCoordinates -Pcoordinates=all
 
   test-all-metadata:
-    name: "🧪 ${{ matrix.coordinates }} (GraalVM ${{ matrix.version }} ${{ matrix.java-version }} @ ${{ matrix.os }})"
+    name: "🧪 ${{ matrix.coordinates }} (GraalVM ${{ matrix.versions.graalvm }} ${{ matrix.versions.java }} @ ${{ matrix.os }})"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 20
     needs: get-all-metadata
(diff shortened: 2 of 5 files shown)


alvarosanchez (changes requested):
It seems this is not that easy.

If this is later used for `graalvm/setup-graalvm` action, the valid combinations are:
- `version: latest`, java-version: 17` (should remove hardcoded 22.3.0  version and use latest)
- `version: dev`, java-version: dev` (see https://github.com/graalvm/setup-graalvm/pull/31).

dnestoro:
> It seems this is not that easy.
> 
> If this is later used for `graalvm/setup-graalvm` action, the valid combinations are:
> 
> * `version: latest`, java-version: 17` (should remove hardcoded 22.3.0  version and use latest)
> * `version: dev`, java-version: dev` (see [graalvm/setup-graalvm#31](https://github.com/graalvm/setup-graalvm/pull/31)).

I have to explore how to implement that. Thanks for the suggestion

linghengqian (commented):
- I think https://github.com/oracle/graalvm-reachability-metadata/blob/master/gradle/wrapper/gradle-wrapper.jar should be removed instead of keeping a copy of the old 7.4 binary in Git.  

- We should also add `**/gradle/wrapper/gradle-wrapper.jar` entry in `.gitignore` and remove `!gradle-wrapper.jar` entry.

linghengqian (commented):
- According to https://github.com/oracle/graalvm-reachability-metadata/actions/runs/4016627730/jobs/6900250113 , it seems that when `org.jetbrains.kotlin-kotlin-reflect-1.7.10` detects itself running on JDK19  , will automatically block tests, it seems that need add `＠Disabled` annotations to all its unit tests temporarily?  Cc @mhalbritter , the original committer of this unit test.  

- https://github.com/oracle/graalvm-reachability-metadata/actions/runs/4016627730/jobs/6900237366 shows that `org.eclipse.jetty-jetty-server-11.0.12` is calling the JDK19 Virtual Thread class, which it is so weird, I can't see the relevant operations in the unit test.

- It looks like we have to temporarily change the `build.gradle` plugin settings in https://github.com/oracle/graalvm-reachability-metadata/blob/master/tests/src/org.eclipse.jetty/jetty-server/11.0.12/build.gradle .

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -86,8 +86,10 @@ if (project.hasProperty("baseCommit")) {
 }
 
 def matrixDefault = [
-        "version"     : ["22.3.0", "dev"], // TODO: Add other supported versions
-        "java-version": ["17"], // TODO: Add "17", "19"
+        "graal-java-versions" : [
+//                [ graal: "dev", java: "dev" ], // TODO: uncomment this once we are ready to run tests on JDK19
+                [ graal: "latest", java: "17"]
+        ],
```
fniephaus, line 92: I think it makes sense to simplify this a bit:

```suggestion
        "versions" : [
//                [ "graalvm": "dev", "java": "dev" ], // TODO: uncomment this once we are ready to run tests on JDK19
                [ "graalvm": "latest", "java": "17"]
        ],
```

dnestoro: I will do this change locally, since this requires one more change.

fniephaus (approved):
LGTM, thanks!

dnestoro:
As discussed with @fniephaus we will disable JDK19 builds temporarily, to unblock CI. Builds will be enabled again, as soon as we define the best solution for the problem.

linghengqian:
> As discussed with @fniephaus we will disable JDK19 builds temporarily, to unblock CI. Builds will be enabled again, as soon as we define the best solution for the problem.

- Once this PR is merged, I can try to fix JDK 19 related unit tests in another PR. I have a PR addressing issues with Jetty Server, Jetty Client and Kotlin Reflect.

alvarosanchez (approved):

PR #152, opened 2022-12-16
Add metadata for graphql-java-extended-validation

bclozel:
## What does this PR do?

This PR builds on the existing graphql-java metadata and adds the relevant content for the graphql-java-extended-validation artifact.

## Checklist before merging
- [x] I have properly formatted metadata files (see [CONTRIBUTING](/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md) 
document)
- [x] I have added thorough tests. (see [this](/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#Tests))

PR #214, opened 2023-02-01
Fix "requires" clauses being ignored

### LINKED ISSUE #211: TCK does not use metadata from declared dependencies
**Describe the bug**
The TCK does not include metadata from declared dependencies. See [this PR comment](https://github.com/oracle/graalvm-reachability-metadata/pull/152#issuecomment-1359537797) for context.

**To Reproduce**
Remove the `configurationFileDirectories.from(tck.metadataRoot.dir...` lines from the graphql-java-extended-validation TCK `build.gradle` and run:

```
./gradlew test -Pcoordinates=com.graphql-java:graphql-java-extended-validation:19.1
```

The test suite should fail on missing metadata related to jboss-logging, hibernate-validator and graphql-java - all libraries with actual metadata in this repository. 

**Expected behavior**
The test should pass without manually adding the metadata for known dependencies listed in the `"requires"` attribute of the `index.json` document.

melix:
This commit fixes the metadata repository not being used in the TCK. Before, it was assumed that each test was working in isolation, but in practice, #211 showed that it is not the case.

So instead of the TCK figuring out by itself which metadata config to include, and completely ignoring the "requires" clauses, we will now simply configure the tests to use the metadata from the current state of the repository. It makes things easier, at the cost of possible side effects: it is possible that the metadata from a transitive dependency is actually what makes a build pass.

However this drawback should be covered by regular code reviews, as it is unexpected that metadata for a transitive "talks" about the current module under test.

Fixes #211

## Checklist before merging

### DIFF
--- a/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
+++ b/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
@@ -49,13 +49,6 @@ String metadataPath = Objects.requireNonNullElse(
 		providers.gradleProperty('metadata.dir').get()
 )
 
-File metadataFile
-if (!metadataPath.startsWith("/")) {
-	metadataFile = file(tck.metadataRoot).toPath().resolve(metadataPath).toFile()
-} else {
-    metadataFile = file(metadataPath)
-}
-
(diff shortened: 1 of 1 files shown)


bclozel (approved):
I've tested this change with #152 (without the workaround) and I confirm that this change makes the native tests pass. Thanks!

dnestoro (approved):

Issue #163, opened 2023-01-06
Add support for `org.apache.curator:curator-*:5.4.0`


linghengqian:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

- Add support for `org.apache.curator:curator-framework:5.4.0`.
- Add support for `org.apache.curator:curator-client:5.4.0`.
- Add support for `org.apache.curator:curator-recipes:5.4.0`.

**Describe alternatives you've considered**

- Null.

**Additional context**

- Metadata generated at https://github.com/linghengqian/graalvm-trace-metadata-smoketest/issues/1 .
- Linkes to https://github.com/apache/shardingsphere/discussions/12594 and https://github.com/apache/shardingsphere/issues/21347 .

linghengqian:
- I'm working on this issue.

linghengqian:
- We need to wait for the zookeeper version release that includes https://github.com/apache/zookeeper/pull/1942 . NativeTest on JDK17 fails at all. Refer to https://github.com/linghengqian/graalvm-trace-metadata-smoketest/tree/main/curator-client .
```shell
Error: Found 1 violations of @Uninterruptible usage:
method java.lang.Thread.getId() is annotated but org.apache.zookeeper.server.quorum.QuorumPeer.getId() is not
com.oracle.svm.core.util.UserError$UserException: Found 1 violations of @Uninterruptible usage:
method java.lang.Thread.getId() is annotated but org.apache.zookeeper.server.quorum.QuorumPeer.getId() is not
        at org.graalvm.nativeimage.builder/com.oracle.svm.core.util.UserError.abort(UserError.java:73)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.code.UninterruptibleAnnotationChecker.reportViolations(UninterruptibleAnnotationChecker.java:100)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.code.UninterruptibleAnnotationChecker.checkBeforeCompilation(UninterruptibleAnnotationChecker.java:91)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.code.CompileQueue.finish(CompileQueue.java:433)
        at org.graalvm.nativeimage.builder/com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:651)
[...]

PR #220, opened 2023-02-06
Increase the timeout period of CI to solve the unit test timeout problem of some libraries

linghengqian:
## What does this PR do?

- For https://github.com/oracle/graalvm-reachability-metadata/issues/163#issuecomment-1418723038 .
- Increase the timeout period of CI to solve the unit test timeout problem of some libraries.

## Checklist before merging
- [x] I have properly formatted metadata files (see [CONTRIBUTING](/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md) document)
- [x] I have added thorough tests. (see [this](/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#Tests))

### DIFF
--- a/.github/workflows/test-all-metadata.yml
+++ b/.github/workflows/test-all-metadata.yml
@@ -34,7 +34,7 @@ jobs:
   test-all-metadata:
     name: "🧪 ${{ matrix.coordinates }} (GraalVM ${{ matrix.versions.graalvm }} ${{ matrix.versions.java }} @ ${{ matrix.os }})"
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 20
+    timeout-minutes: 80
     needs: get-all-metadata
     strategy:
       fail-fast: false
--- a/.github/workflows/test-changed-metadata.yml
+++ b/.github/workflows/test-changed-metadata.yml
@@ -37,7 +37,7 @@ jobs:
     name: "🧪 ${{ matrix.coordinates }} (GraalVM ${{ matrix.versions.graalvm }} ${{ matrix.versions.java }} @ ${{ matrix.os }})"
     if: needs.get-changed-metadata.result == 'success' && needs.get-changed-metadata.outputs.none-found != 'true'
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 20
+    timeout-minutes: 80
     needs: get-changed-metadata
 
     strategy:


.github/workflows/test-all-metadata.yml
```diff
@@ -34,7 +34,7 @@ jobs:
   test-all-metadata:
     name: "🧪 ${{ matrix.coordinates }} (GraalVM ${{ matrix.versions.graalvm }} ${{ matrix.versions.java }} @ ${{ matrix.os }})"
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 20
+    timeout-minutes: 80
```
fniephaus, line 37: Why do we need to increase timeout by 4x? How long does it take to build https://github.com/oracle/graalvm-reachability-metadata/issues/163#issuecomment-1418723038 on GitHub Actions? I'd prefer if we could selectively bump the timeout for metadata that requires it.

linghengqian: - Completing nativeTest for `org.apache.curator:curator-recipes:5.4.0` took 62 minutes on my device. I've created the smallest possible subset of unit tests for this dependency on my local device, but the full test still takes so long.

>  I'd prefer if we could selectively bump the timeout for metadata that requires it.

-  I don't quite understand, how to do this?

fniephaus: > Completing nativeTest for `org.apache.curator:curator-recipes:5.4.0` took 62 minutes on my device.

That is quite a lot. Were these 64min dominated by Native Image builds or by actually running the tests?

> I don't quite understand, how to do this?

Right now, we don't have a mechanism for this, but maybe we can come up with something that allows overriding the timeout if necessary. We could implement that in [`org.graalvm.internal.tck-harness.gradle`](https://github.com/linghengqian/graalvm-reachability-metadata/blob/985f371e5dc9fc5a460619359166ed07ebe5d38d/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle) somehow, and then use:

```yml
timeout-minutes: ${{ matrix.timeout-minutes }}
```

fniephaus: Do we need to run the unit tests of `curator-recipes` on the JVM or can we skip this step?

linghengqian: - I don't think it's necessary to run unit tests on the JVM, it's definitely something to consider. But, what I see in https://graalvm.github.io/native-build-tools/latest/gradle-plugin.html#testing-support says
```
Currently, this feature requires the execution of the tests in the classic "JVM" mode prior to the execution of tests in native mode. 
```
- I honestly don't see an option to disable nativeTest from executing unit tests on the JVM, is there something I'm overlooking here?

fniephaus: >I honestly don't see an option to disable nativeTest task from executing unit tests on the JVM, is there something I'm overlooking here?

You are right, I'm afraid tests currently need to run on the JVM so that we can reliably detect all test ids.

linghengqian: - So should I close this PR now?

fniephaus: You can unless you want to look into making timeouts adjustable per metadata entry, which I believe we will need to support `curator-recipes`.

PR #248, opened 2023-03-13
Update metadata for `ch.qos.logback:logback-classic:1.4.1`

msupic:
## What does this PR do?
Adds metadata to `ch.qos.logback:logback-classic:1.4.1` for external logback config file usage.

## Checklist before merging


tests/src/ch.qos.logback/logback-classic/1.4.1/generate-metadata.sh
dnestoro, line 1: Hey, can you just explain me why do we need this script? It looks like it is not executed anywhere in the code? If it was used just for your local testing, then it probably doesn't need to be pushed.

tests/src/ch.qos.logback/logback-classic/1.4.1/src/test/java/org/graalvm/logback/LogbackTests.java
```diff
@@ -44,6 +46,27 @@ public void setUp() {
...
+  void traceMessageSkippedViaExternalConfig() {
+    Logger logger = context.getLogger(Logger.ROOT_LOGGER_NAME);
+    logger.trace("test message");
+    assertThat(outputStreamCaptor.toString()).isBlank();
+  }
+
+  @Test
+  void consoleAppenderPropertySetter() {
+    ConsoleAppender consoleAppender = new ConsoleAppender();
+    PropertySetter propertySetter = new PropertySetter(new BeanDescriptionCache(null), consoleAppender);
```
dnestoro, line 66: What does this `BeanDescriptionCache` creates?

tests/src/ch.qos.logback/logback-classic/1.4.1/generate-metadata.sh
msupic: The script was used to create metadata that were copied to the `metadata/ch.qos.logback/logback-classic/1.4.1` directory. The script is needed because it executes an additional custom gradle task `updateGeneratedMetadata`. Without the script, metadata won't be valid. Also, the README.md file contains info how metadata have been created. If more metadata need to be added in future, it is easier to add and maintain when the exact procedure of creating metadata is explained and automatized.

PR #286, opened 2023-05-15
Fix compatibility with Hibernate 6.2.2

sdeleuze:
This commit adds additional reflection entries for logger that are required due to https://hibernate.atlassian.net/browse/HHH-16507 and https://hibernate.atlassian.net/browse/HHH-16441 to fix compatibility with Hibernate 6.2.2.

I chose to not create a dedicated new version for maintanablity reasons and because those hints are triggered by `typeReachable` condition that allow to support both 6.2.0 and 6.2.2.

### DIFF
--- a/metadata/org.hibernate.orm/hibernate-core/6.2.0.Final/reflect-config.json
+++ b/metadata/org.hibernate.orm/hibernate-core/6.2.0.Final/reflect-config.json
@@ -882,5 +882,15 @@
   {
     "condition":{"typeReachable":"org.hibernate.dialect.PgJdbcHelper"},
     "name":"org.postgresql.util.PGobject"
+  },
+  {
+    "condition":{"typeReachable":"org.hibernate.dialect.DialectLogging"},
+    "name":"org.hibernate.dialect.DialectLogging_$logger",
+    "methods":[{"name":"<init>","parameterTypes":["org.jboss.logging.Logger"] }]
+  },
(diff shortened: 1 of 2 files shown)


fniephaus (approved):
LGTM, thanks!

matthenry87:
Hi @sdeleuze,

I still had to add the below hints with 3.0.1 today to make native work (using Spring Framework 6.0.10-SNAPSHOT)

```json
[
    {
        "name": "org.hibernate.dialect.DialectLogging_$logger",
        "methods":
        [
            {
                "name": "<init>",
                "parameterTypes":
                [
                    "org.jboss.logging.Logger"
                ]
            }
        ]
    },
    {
        "name": "org.hibernate.metamodel.mapping.MappingModelCreationLogging_$logger",
        "methods":
        [
            {
                "name": "<init>",
                "parameterTypes":
                [
                    "org.jboss.logging.Logger"
                ]
            }
        ]
    }
]
```

I do see hints for these in the current reachability release - but still would not work unless I added them.

sdeleuze:
@matthenry87 I guess you did not configure `0.3.1` metadata version explicitly so `0.3.0` is used by default.

matthenry87:
> @matthenry87 I guess you did not configure `0.3.1` metadata version explicitly so `0.3.0` is used by default.

Yup you are correct, my apologies.

PR #289, opened 2023-05-16
Increase checkstyle timeout

dnestoro:
Currently actions are failing because of short timeout set in checkstyle.yml workflow.


fniephaus (approved):
LGTM

PR #296, opened 2023-05-22
Add allowed-packages to scaffold task

dnestoro:
Extends scaffold task with newly added entry `allowed-packages`


fniephaus (approved):
LGTM

PR #316, opened 2023-06-12
Added metadata for commons-logging:commons-logging:1.2

msupic:
## What does this PR do?
Adds metadata for commons-logging:commons-logging:1.2

## Checklist before merging
- [x] I have considered including reachability metadata directly in the library or the framework (see [this](./CONTRIBUTING.md))
- [x] I have properly formatted metadata files (see [this](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#format-metadata-files))
- [x] I have added thorough tests (see [this](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#Tests))

### DIFF
--- a/metadata/commons-logging/commons-logging/index.json
+++ b/metadata/commons-logging/commons-logging/index.json
@@ -1,0 +1,10 @@
+[
+  {
+    "latest": true,
+    "metadata-version": "1.2",
+    "module": "commons-logging:commons-logging",
+    "tested-versions": [
+      "1.2"
+    ]
+  }
+]
--- a/metadata/commons-logging/commons-logging/1.2/index.json
+++ b/metadata/commons-logging/commons-logging/1.2/index.json
@@ -1,0 +1,3 @@
+[
+  "reflect-config.json"
+]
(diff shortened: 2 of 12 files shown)


metadata/commons-logging/commons-logging/index.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "latest": true,
```
melix, line 3: Should we use the new `default-for` flag too?

msupic: If we have metadata for only one version, how we would set `default-for`regexp? To which versions we should limit the  metadata? The existing latest metadata might be appropriate for future versions too, so I think we should not limit versions for latest metadata until metadata for a new version is created. @vjovanov @fniephaus - your thoughts?

msupic: I will merge these changes to the master branch since I need it for other tests, but I guess we can continue discussion here for future PRs.

PR #322, opened 2023-06-22
Upgrade Native Build Tools and update to GraalVM for JDK 17.

fniephaus:
## What does this PR do?
This PR upgrade the NBT to `0.9.23` and updates all uses of the `setup-graalvm` action to use the new GraalVM for JDK 17 release.

## Checklist before merging
n/a

### DIFF
--- a/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
+++ b/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
@@ -90,10 +90,8 @@ if (project.hasProperty("baseCommit")) {
 }
 
 def matrixDefault = [
-        "versions" : [
-            [ "java": "17" ]
-        ],
-        "os"       : ["ubuntu-latest"] // TODO: Add support for "windows-latest", "macos-latest"
+        "version": ["17"],
+        "os"     : ["ubuntu-latest"] // TODO: Add support for "windows-latest", "macos-latest"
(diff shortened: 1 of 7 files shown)


tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -91,10 +91,9 @@ if (project.hasProperty("baseCommit")) {
 
 def matrixDefault = [
         "versions" : [
-                [ "graalvm": "dev",    "java": "17" ],
-                [ "graalvm": "latest", "java": "17" ]
+            [ "java": "17" ]
```
dnestoro, line 94: I think we don't need this extra subsection. Since we don't use pairs anymore, we can remove "java", and use simple array as it is the case for [os](https://github.com/oracle/graalvm-reachability-metadata/pull/322/files#diff-1152174d2b5a90761fb50a4ed8aa633feb9e97a726bded72e1c3ac4ed16f0737R96). And then change `matrix.versions.java` -> `matrix.versions` [here](https://github.com/oracle/graalvm-reachability-metadata/pull/322/files#diff-5f5232eedb6d9df7c3be178e6b1150b9d6a4d975dda714ef84d7c74feb2df241R38) as we do for `matrix.os`

dnestoro (approved):

melix (approved):

PR #347, opened 2023-07-24
Narrow the search in getAllowedPackages function

dnestoro:
## What does this PR do?
Currently we are looking for `allowedPackages` field in `metdata/index.json` by searching for the `groupId` in the module name of the corresponding library. It turns out that this approach misses corner case when two libraries have the same `groupId` but different `artifactId` (Example: `org.eclipse.paho:org.eclipse.paho.client.mqttv3` and `org.eclipse.paho:org.eclipse.paho.mqttv5.client`). This PR narrows the search by looking for both `groupId` and `artifactId` in the module name.

### DIFF
--- a/metadata/index.json
+++ b/metadata/index.json
@@ -130,7 +130,10 @@
   },
   {
     "directory" : "org.apache.tomcat/tomcat-jdbc",
-    "module" : "org.apache.tomcat:tomcat-jdbc"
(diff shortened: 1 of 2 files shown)


fniephaus (approved):
LGTM, thanks!

msupic (approved):

PR #351, opened 2023-07-25
Set up dependabot for automated Dockerfile updates.

fniephaus:
## What does this PR do?

This PR adds a `dependabot.yml` config that enables automatic Dockerfile version updates, and adds Dockerfiles for all allowed Docker images.

## Checklist before merging

n/a


dnestoro:
LGTM! Next steps would be changing DockerTask and GrypeTask so we can get rid of the list. I can do that. What do you think @fniephaus should we do that in a follow-up PR or I should contribute to this PR so we can merge everything together?

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+
+version: 2
+
+updates:
+  - package-ecosystem: "docker"
+    directory: "/tests/tck-build-logic/src/main/resources/allowed-docker-images"
+    reviewers:
+      - "matneu"
+    schedule:
+      interval: "weekly"
```
matneu, line 12: Any reason we pick weekly? I think this can be a daily pass, shouldn't take long?

fniephaus: Sure, can be daily.

fniephaus:
> What do you think @fniephaus should we do that in a follow-up PR or I should contribute to this PR so we can merge everything together?

I don't think we need to do everything in this PR. Let's get rid of the `AllowedDockerImages.txt` in a follow-up PR.

dnestoro (approved):

matneu (approved):

PR #356, opened 2023-07-25
Http client connection issue

msupic:
Trying to troubleshooting occasional http client connection issues

### DIFF
--- a/tests/src/org.apache.httpcomponents/httpclient/4.5.14/build.gradle
+++ b/tests/src/org.apache.httpcomponents/httpclient/4.5.14/build.gradle
@@ -15,3 +15,11 @@ dependencies {
     testImplementation "org.apache.httpcomponents:httpclient:$libraryVersion"
     testImplementation 'org.assertj:assertj-core:3.22.0'
 }
+
+graalvmNative {
+    binaries {
+        test {
+            buildArgs.add('--enable-url-protocols=http')
+        }
+    }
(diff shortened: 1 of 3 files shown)


melix (approved):

tests/src/org.apache.httpcomponents/httpclient/4.5.14/build.gradle
```diff
@@ -15,3 +15,11 @@ dependencies {
     testImplementation "org.apache.httpcomponents:httpclient:$libraryVersion"
     testImplementation 'org.assertj:assertj-core:3.22.0'
 }
+
+graalvmNative {
+    binaries {
+        test {
+            buildArgs.add('--enable-url-protocols=http')
```
dnestoro, line 22: Why do we need this?

msupic: The test is using an http request to check whether nginx container is started. The http protocol is not enabled by default in graalvm https://www.graalvm.org/latest/reference-manual/native-image/dynamic-features/URLProtocols/, so the test fails if there is no `--enable-url-protocols=http`

dnestoro: Thanks for the explanation!

dnestoro (approved):

PR #361, opened 2023-08-03
Disable dev builds since they are blocking github actions

dnestoro:


fniephaus (approved):
LGTM

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -90,7 +90,7 @@ if (project.hasProperty("baseCommit")) {
 }
 
 def matrixDefault = [
-        "version": ["17", "dev"],
+        "version": ["17"],
```
fniephaus, line 93: Oh, maybe just comment it out, so we know how to turn it back on.

dnestoro: It is a trivial change, but okay, we can comment it (instead of deleting it) as well.

fniephaus: It's trivial for you and me, but maybe not for others 😉

PR #402, opened 2023-10-27
Make docker images scanner more flexible

dnestoro:
## What does this PR do?
Make docker images scanner:

-  weekly job for checking vulnerabilities in existing docker images
-  triggered on allowed docker images list update


.github/workflows/scan-docker-images-skip.yml
```diff
@@ -1,14 +0,0 @@
-name: "Scan docker images from the allowed docker images list"
-
-on:
-  pull_request:
-    paths:
```
fniephaus, line 5: I think we'll still need the skip job, because otherwise, there is no job with that name that matches the requirements on a PR. Remember?

fniephaus (approved):
LGTM, thanks!

PR #407, opened 2023-10-27
Test CODEOWNERS and docker scanner PR

dnestoro:
## What does this PR do?

Check if: 

- adequate code owner will be added 
- docker image scanner runs when list of allowed is changed images


dnestoro:
Works as expected. Codeowner is added correctly and adequate action is triggered

PR #416, opened 2023-10-31
Add abillity to run docker task only for changed images

dnestoro:
## What does this PR do?
Run docker images scanner job:

-  for changed docker images only (on pull request)
- for all docker images (weekly)


tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/ScaffoldTask.java
```diff
@@ -28,7 +28,7 @@
  *
  * @author Moritz Halbritter
  */
-class ScaffoldTask extends DefaultTask {
+public class ScaffoldTask extends DefaultTask {
```
dnestoro, line 31: This change was made because IDE reported following warning: `Access to 'ScaffoldTask' exceeds its access rights` in `tck-harness.gradle`.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/DockerUtils.java
```diff
@@ -7,22 +7,25 @@
...
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 public class DockerUtils {
+    private static String getDockerfileDirectory() {
+        return Paths.get("./tests/tck-build-logic/src/main/resources/allowed-docker-images").toString();
```
fniephaus, line 17: Will this always resolve the correct directory? Should we make it relative to `DockerUtils.java`?

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/ScaffoldTask.java
fniephaus: Does it actually cause a problem?

.github/workflows/scan-docker-images.yml
```diff
@@ -23,7 +23,14 @@ jobs:
           java-version: '17'
           distribution: 'graalvm'
           github-token: ${{ secrets.GITHUB_TOKEN }}
-      - name: "🔎 Check docker images"
+      - name: "🔎 Check changed docker images"
+        if: github.event_name == 'pull_request'
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sudo sh -s -- -b /usr/local/bin
+          sudo apt-get install jq
```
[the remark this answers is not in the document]
dnestoro: Yes, I even planned to do so, but somehow I missed to do that. Good catch!

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/ScaffoldTask.java
dnestoro: I reverted the change since it is only a warning and doesn't cause a problem

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/DockerUtils.java
dnestoro: I changed whole handling to use resources instead of Paths and Files

fniephaus (approved):
LGTM

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
```diff
@@ -2,31 +2,74 @@
...
+    private String newCommit;
+    private String baseCommit;
+
     private final String jqMatcher = " | jq -c '.matches | .[] | .vulnerability | select(.severity | (contains(\"High\") or contains(\"Critical\")))'";
 
+    private List<URL> getChangedImages(String base, String head){
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        getExecOperations().exec(spec -> {
+            spec.setStandardOutput(baos);
+            spec.commandLine("git", "diff", "--name-only", "--diff-filter=ACMRT", base, head);
```
fniephaus, line 43: Hang on, isn't this only to check whether a file in `tests/tck-build-logic/src/main/resources/allowed-docker-images` has changed? If so, I'd be for duplicating the workflow: one for PRs and one for weekly checks. Or why do we need to maintain this code here?

fniephaus (changes requested):

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
dnestoro: No, this is actually to get all files that are changed or added. With workflows we only know that "something" changed on specified path, but we don't know what files have changed. This way we will get a list of changed/added files from allowed-docker-images and execute check only for them. We don't need to execute for example `nats` docker image scan if we only bumped `mysql` image version. On the other hand, we will execute scan of all docker images once a week to check if new vulnerabilities are discovered (but that won't block PRs that change only one image).

fniephaus (approved):
LGTM

Issue #151, opened 2022-12-16
Adding large SDKs to the library-and-framework-list.json


msailes:
**Is your feature request related to a problem? Please describe.**
Adding the AWS SDK for Java 2.0 to the library-and-framework-list.json is difficult because it is made up of 418 separate artifacts. Adding each one individually would be annoying and bloat the file for other users.

**Describe the solution you'd like**
Allow the artifact field to include a `*` `"artifact": "software.amazon.awssdk:*"`

**Describe alternatives you've considered**
Writing out 418 separate enteries

**Additional context**

This was my first draft at the addition

```
  {
    "artifact": "software.amazon.awssdk:*",
    "description": "AWS SDK for Java 2.0",
    "details": [
      {
        "minimum_version": "2.16.1",
        "metadata_locations": [
          "https://repo1.maven.org/maven2/software/amazon/awssdk/"
        ],
        "tests_locations": [
          "https://github.com/aws/aws-sdk-java-v2/tree/master/test/sdk-native-image-test"
        ],
        "test_level": "community-tested"
      }
    ]
  }
```

fniephaus:
Hi Mark,
Let's follow up on this. So far, the JSON is used to generate [the table on this website](https://www.graalvm.org/native-image/libraries-and-frameworks/), so I don't see a reason why we can't allow wildcards for the artifact ID. If you could open a PR with the entry you'd like to add, we can adjust the schema accordingly. Thanks!

PR #417, opened 2023-11-02
Adding the AWS SDK v2.0

msailes:
## What does this PR do?
PR for https://github.com/oracle/graalvm-reachability-metadata/issues/151

## Code sections where the PR accesses files, network, docker or some external service
N/A
- (example link to code section)

## Checklist before merging
- [ ] I have considered including reachability metadata directly in the library or the framework (see [our contributing document](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md))
- [ ] I am the original author of all content provided in the pull request, and I did not copy the content from any other source (see [the tests section from our contributing document](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#tests))
- [ ] For all tests where I am not the sole author, I have added a comment that proves I may publish them under the specified license (see [the tests section from our contributing document](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#tests))
- [ ] I have properly formatted metadata files (see [our formatting guide](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#format-metadata-files))
- [ ] I have added thorough tests (see [the tests section from our contributing document](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#Tests))
- [ ] I have filled all places where my pull request accesses files, network, docker, or any other external service (see the section above)

PR #418, opened 2023-11-03
Add Asterix char as valid part of library or framework name pattern

dnestoro:
## What does this PR do?

Fixes https://github.com/oracle/graalvm-reachability-metadata/issues/151


fniephaus (changes requested):
I don't think we want to allow wildcards in the group ID. I'm not even sure we can allow wildcards in the artifact ID.

dnestoro:
Closing this PR because changes are applied [here](https://github.com/oracle/graalvm-reachability-metadata/pull/417#pullrequestreview-1712352704)

PR #466, opened 2024-03-20
Run latest ea builds in GitHub actions

dnestoro:
## What does this PR do?

- Runs ci actions with `latest-ea` builds
- Fixes kotlin, undertow and protobuf tests to pass on `latest-ea`
- Updates `actions/checkout` and `actions/setup-java` versions

### DIFF
--- a/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
+++ b/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
@@ -90,7 +90,10 @@ if (project.hasProperty("baseCommit")) {
 }
 
 def matrixDefault = [
-        "version": [ "latest-ea" ],
+        "version": [
+            "17",
+            "latest-ea"
+        ],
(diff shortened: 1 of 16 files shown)


tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -124,7 +124,11 @@ Provider<Task> generateMatrixDiffCoordinates = tasks.register("generateMatrixDif
             def matrix = [
                 "coordinates": ["No matches found!"]
             ]
+
             matrix.putAll(matrixDefault)
+            matrix.version.add("21");
```
fniephaus, line 129: Could you add a comment here why we do this? Also, I think we want to add `22` now that it's out :)

fniephaus (approved):
LGTM, one more comment.

dnestoro:
@fniephaus please review again, because I introduced more changes since your approval

PR #485, opened 2024-04-16
Improve MariaDB client test

mhalbritter:
- Uses an ephemeral port for docker
- Uses `127.0.0.1` instead of `localhost`
- Updates MariaDB from `10.8` to `11` (which hopefully passes the vulnerability scanners)
- Fails fast if docker process exits early
- print stdout and stderr if docker process exits early


melix (approved):

dnestoro:
This PR can't be merged since it is using vulnerable docker image

PR #527, opened 2024-08-26
Remove confusing example in library-and-framework-list-schema.json

eregon:
## What does this PR do?

That 2nd details entry in the example was confusing, because if the Native Image config is tracked in this repo vs upstream then that would be tracked in files such as https://github.com/oracle/graalvm-reachability-metadata/blob/master/metadata/io.netty/netty-common/index.json and not in https://github.com/oracle/graalvm-reachability-metadata/blob/master/library-and-framework-list.json which seems to only be when it is in upsteam.

## Checklist before merging

### DIFF
--- a/library-and-framework-list-schema.json
+++ b/library-and-framework-list-schema.json
@@ -15,17 +15,6 @@
             "https://github.com/netty/netty/actions"
           ],
           "test_level": "fully-tested"
-        },
-        {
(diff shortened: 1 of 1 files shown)


eregon:
@fniephaus Could you review this?

fniephaus (approved):
LGTM, thanks!

PR #537, opened 2024-10-03
Add GitHub Action for Scheduled Releases

dnestoro:
## What does this PR do?
This PR introduces github workflow that creates automatic releases. The github action defined in workflow runs every two weeks.  

To make sure this PR works, we must have fixed `SNAPSHOT` version on master. Otherwise the process can't be fully automated and some steps may require manual maintainer inputs (since the master branch has protection rules that disallows direct committing onto it). With the fixed `SNAPSHOT` version on master, we are avoiding direct committing onto the protected master branch.

### DIFF
--- a/.github/workflows/create-scheduled-release.yml
+++ b/.github/workflows/create-scheduled-release.yml
@@ -59,7 +59,7 @@ jobs:
           echo "CURRENT_RELEASE_TAG=$CURRENT_RELEASE_TAG" >> ${GITHUB_ENV}
       - name: "⬆️ Update version"
         run: |
-          sed -i "s/project.version(\"0.3.10-SNAPSHOT\")/project.version(\"${{ env.CURRENT_RELEASE_TAG }}\")/g" build.gradle
+          sed -i "s/project.version(\"1.0.0-SNAPSHOT\")/project.version(\"${{ env.CURRENT_RELEASE_TAG }}\")/g" build.gradle
       - name: "🔍 Run spotless check"
         run: |
           ./gradlew spotlessCheck
--- a/.github/workflows/create-scheduled-release.yml
+++ b/.github/workflows/create-scheduled-release.yml
@@ -1,0 +1,78 @@
+name: "Create scheduled release"
+
+on:
+  schedule:
+    - cron: "0 0 1 * *"
+    - cron: "0 0 15 * *"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  get-changed-metadata:
+    name: "📋 Get a list of changed metadata"
+    runs-on: "ubuntu-20.04"
+    timeout-minutes: 5
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+      none-found: ${{ steps.set-matrix.outputs.none-found }}
+    steps:
+      - name: "☁️ Checkout repository"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: "🔧 Setup java"
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'oracle'
+          java-version: '17'
+      - name: "🕸️ Get changed metadata matrix"
+        id: set-matrix
+        run: |
+          LATEST_TAG=$(ls -1 .git/refs/tags/ | sort -V | tail -1)
+          ./gradlew generateMatrixDiffCoordinates -PbaseCommit=$(git show-ref -s $LATEST_TAG) -PnewCommit=$(git rev-parse HEAD)
+
+  release:
+    needs: get-changed-metadata
+    if: needs.get-changed-metadata.result == 'success' && needs.get-changed-metadata.outputs.none-found != 'true'
+    name: "🚀 Create a release"
+    runs-on: "ubuntu-20.04"
+    env:
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      - name: "☁️ Checkout repository"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: "🔧 Setup java"
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'graalvm'
+          java-version: '17'
+      - name: "⬆️ Update version"
+        run: |
+          PREVIOUS_RELEASE_TAG=$(ls -1 .git/refs/tags/ | sort -V | tail -1)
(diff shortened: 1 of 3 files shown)


.github/workflows/create-scheduled-release.yml
```diff
@@ -0,0 +1,78 @@
...
+          java-version: '17'
+      - name: "⬆️ Update version"
+        run: |
+          PREVIOUS_RELEASE_TAG=$(ls -1 .git/refs/tags/ | sort -V | tail -1)
+          echo "PREVIOUS_RELEASE_TAG=$PREVIOUS_RELEASE_TAG" >> ${GITHUB_ENV}
+          
+          CURRENT_RELEASE_TAG=$(sed -E 's/^([0-9]+\.)([0-9]+\.)([0-9]+)/echo \1\2$((\3+1))/e' <<< $PREVIOUS_RELEASE_TAG)
+          echo "CURRENT_RELEASE_TAG=$CURRENT_RELEASE_TAG" >> ${GITHUB_ENV}
+          
+          sed -i "s/project.version(\"0.3.10-SNAPSHOT\")/project.version(\"$CURRENT_RELEASE_TAG\")/g" build.gradle
```
dnestoro, line 61: `0.3.10-SNAPSHOT` must be replaced with new, fixed version once we agree how it should look like.

From that point on, we will always have the same `SNAPSHOT` version on master, and concrete version in each "release commit".

.github/workflows/create-scheduled-release.yml
```diff
@@ -0,0 +1,78 @@
...
+          fetch-depth: 0
+      - name: "🔧 Setup java"
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'oracle'
+          java-version: '17'
+      - name: "🕸️ Get changed metadata matrix"
+        id: set-matrix
+        run: |
+          LATEST_TAG=$(ls -1 .git/refs/tags/ | sort -V | tail -1)
```
fniephaus, line 33: Fiddling with `git` internals is not great. What about something like this:
```suggestion
          LATEST_TAG=$(git describe --tags --abbrev=0)
```

.github/workflows/create-scheduled-release.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - name: "☁️ Checkout repository"
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: "🔧 Setup java"
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'graalvm'
+          java-version: '17'
```
fniephaus, line 52: Can this run on 21?
```suggestion
          java-version: '21'
```

.github/workflows/create-scheduled-release.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: "⬆️ Update version"
+        run: |
+          PREVIOUS_RELEASE_TAG=$(ls -1 .git/refs/tags/ | sort -V | tail -1)
```
fniephaus, line 55:
```suggestion
          PREVIOUS_RELEASE_TAG=$(git describe --tags --abbrev=0)
```

.github/workflows/create-scheduled-release.yml
fniephaus: Is this the only place that needs to be updated? the tests won't need to be updated because they will need to match the version on the main branch, which currently is `0.3.10-SNAPSHOT`, right?

.github/workflows/create-scheduled-release.yml
```diff
@@ -0,0 +1,78 @@
...
+          sed -i "s/project.version(\"0.3.10-SNAPSHOT\")/project.version(\"$CURRENT_RELEASE_TAG\")/g" build.gradle
+      - name: "📄 Commit changes"
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "Github Actions"
+          git add .
+          git commit -m "Release version ${{ env.CURRENT_RELEASE_TAG }}"
+          git tag ${{ env.CURRENT_RELEASE_TAG }}
+          git push origin ${{ env.CURRENT_RELEASE_TAG }}
+      - name: "🔍 Run spotless check"
```
fniephaus, line 70: I would run `spotlessCheck` and `package` before pushing the changes. Otherwise, we potentially push changes that fail those steps.

fniephaus (changes requested):

.github/workflows/create-scheduled-release.yml
dnestoro: Yes, this is the only place where we have to update project version.

.github/workflows/create-scheduled-release.yml
dnestoro: I already tried that, but the `git` command you mentioned returns the last tag from the current branch, and since our "release tags" won't be on the master, this won't work (it will always return `0.3.9` which is currently the latest tag on master, and after we create the next release (`0.3.10`) with the tag not on master, the release after that will again see `0.3.9` as the latest tag if we use the git command from above)

.github/workflows/create-scheduled-release.yml
dnestoro: Again, the same [explanation as above](https://github.com/oracle/graalvm-reachability-metadata/pull/537#discussion_r1787497591).

.github/workflows/create-scheduled-release.yml
dnestoro: Good catch! Thanks!

.github/workflows/create-scheduled-release.yml
fniephaus: Right, how about this?
```suggestion
          LATEST_TAG=$(git tag --list | sort -V | tail -1)
```

dnestoro: That works!

.github/workflows/create-scheduled-release.yml
dnestoro: Version will be changed to `1.0.0-SNAPSHOT` in [this PR](https://github.com/oracle/graalvm-reachability-metadata/pull/541). Mentioned version will only act as a placeholder version that will be changed in actual releases (through CI)

fniephaus: Why do it in a separate PR though?

melix: +1, let's do it in the same PR to make it easier

melix (approved):

.github/workflows/create-scheduled-release.yml
dnestoro: Done!

PR #543, opened 2024-10-07
Create Latest Release for Every Merged Pull Request

dnestoro:
## What does this PR do?
This PR enables automatic `latest` releases. Whenever a new metadata is merged, we will update current `latest` tag (delete previous and create a new one) to contain that metadata.

To make this work, we must create initial latest tag with the corresponding release.

### DIFF
--- a/.github/workflows/create-latest-release.yml
+++ b/.github/workflows/create-latest-release.yml
@@ -1,10 +1,9 @@
-name: "Create scheduled release"
+name: "Create latest release"
 
 on:
-  schedule:
-    - cron: "0 0 1 * *"
-    - cron: "0 0 15 * *"
-  workflow_dispatch:
+  push:
+    branches:
+      - master
 
 permissions:
   contents: write
@@ -25,13 +24,12 @@ jobs:
       - name: "🔧 Setup java"
         uses: actions/setup-java@v4
         with:
-          distribution: 'graalvm'
+          distribution: 'oracle'
           java-version: '21'
       - name: "🕸️ Get changed metadata matrix"
         id: set-matrix
         run: |
-          LATEST_TAG=$(git tag --list | grep -v latest | sort -V | tail -1)
-          ./gradlew generateMatrixDiffCoordinates -PbaseCommit=$(git show-ref -s $LATEST_TAG) -PnewCommit=$(git rev-parse HEAD)
+          ./gradlew generateMatrixDiffCoordinates -PbaseCommit=$(git show-ref -s "latest") -PnewCommit=$(git rev-parse HEAD)
 
   release:
     needs: get-changed-metadata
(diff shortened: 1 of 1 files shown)


.github/workflows/create-scheduled-release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: "🕸️ Get changed metadata matrix"
         id: set-matrix
         run: |
-          LATEST_TAG=$(git tag --list | sort -V | tail -1)
+          LATEST_TAG=$(git tag --list | grep -v latest | sort -V | tail -1)
```
dnestoro, line 33: This additional step skips `latest` tag as potential base for next standard release tag name (as we always bump patch version in the version string, we want to do so from the latest **standard** release).

fniephaus:
Why is this needed? Are you sure `latest` is even an accepted value for the version string? If we need more frequent releases, I'd vote for increasing the number of builds per month, for example to once every week.

dnestoro:
> Why is this needed? Are you sure `latest` is even an accepted value for the version string? If we need more frequent releases, I'd vote for increasing the number of builds per month, for example to once every week.

This kind of release can give users multiple benefits. For example:
- Users don't need to wait two weeks to be able to use their metadata
- Also users may want always new metadata in their projects in the development
- Besides that if we make releases too frequent (weekly or even more frequent) we may unnecessarily spend github resources for  storing many releases with just 1-2 libraries difference

For the `latest` as a version string I am not sure. I tried it on a testing repo and Github accepts it. We can discuss a different name for this kind of releases. Maybe we can use something with `SNAPSHOT` annotation (or even only `SNAPSHOT`)?

fniephaus:
If someone wants to you the latest metadata, wouldn't it be better if the consumer, the thing that downloads the metadata, uses this link rather than a latest tag that we need to manage manually: https://github.com/oracle/graalvm-reachability-metadata/archive/refs/heads/master.zip

melix:
So I think having a `head`/`master`/`snapshot` release is indeed useful. For example a user can download the archive and add it to their sources, then configure the plugin to use that zip.

The question that Fabio raises is whether we actually need to do that, given that there's already something similar provided out of the box by GitHub. The difference is that the content of the zip file is different: a metadata release contains directly, at the root of the zip file, the content of the [metadata](https://github.com/oracle/graalvm-reachability-metadata/tree/master/metadata) directory only (the rest of the repository is irrelevant for users). The zip provided by GitHub contains a top level directory named `graalvm-reachability-metadata-master` then the whole repository. Therefore it cannot be exploited directly by the user (or the plugins).

.github/workflows/create-scheduled-release.yml
vjovanov: Maybe write that comment in the file?

vjovanov:
I would call it `<name>-SNAPSHOT.zip`, as this is what is the maven community uses. Since we are providing this for the JVM I feel it would immediately ring a bell to all the users.

.github/workflows/create-latest-release.yml
```diff
@@ -0,0 +1,73 @@
...
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "Github Actions"
+          git add .
+          git commit -m "Update latest release"
+          git tag "latest"
+          git push origin "latest"
+      - name: "📝 Publish a release"
+        run: |
+          gh release create "latest" build/graalvm-reachability-metadata-*.zip --title "Latest"
```
vjovanov, line 73: Missing line

vjovanov:
LGTM

vjovanov:
We need this more than ever as we release 50 libs a day now. @jormundur00 can you handle this in the next couple of days?

PR #552, opened 2024-10-25
Scan new versions for existing libraries

dnestoro:
## What does this PR do?
This PR introduces a new GitHub workflow that scans for newer versions of libraries that currently exist in the repo. 
When there is a newer version available, we are trying to run existing tests with existing metadata to verify if tests can pass.

- In case when existing tests can pass using the newer version of the library, the workflow will automatically update `tested-versions` list in the corresponding `index.json` file. Accumulated results will be provided through an automatically created PR.
- Otherwise, when tests cannot pass with existing metadata for newer library version, workflow will store this information and create an issue with accumulated information about all unsupported versions.

NOTE: we must allow github actions to create PRs automatically.

### DIFF
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -103,14 +103,12 @@ jobs:
           sudo bash -c "cat ./.github/workflows/dockerd.service > /etc/systemd/system/docker.service.d/http-proxy.conf"
           sudo systemctl daemon-reload
           sudo systemctl restart docker
-      - name: "Replace version"
-        run: |
-          CURRENT_VERSION=$(echo ${{ matrix.coordinates }} | cut -d ":" -f3)
-          sed -i "s/library.version\s*=\s*${{ env.LATEST_VERSION }}/library.version = $CURRENT_VERSION/g" tests/src/${{ env.TEST_PATH }}/gradle.properties
       - name: "🧪 Run '${{ env.TEST_COORDINATES }}' tests"
         run: |
+          TESTING_VERSION=$(echo ${{ matrix.coordinates }} | cut -d ":" -f3)
+          export GVM_TCK_LV=TESTING_VERSION
+
           ./gradlew test -Pcoordinates=${{ env.TEST_COORDINATES }}
-          git stash
       - name: "New library is supported"
         if: success()
         run: |
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -38,7 +38,7 @@ jobs:
       - name: "🕸️ Populate matrix"
         id: set-matrix
         run: |
-          ./gradlew getExistingLibrariesWithNewVersions -PlibraryMatrixLimit=200
+          ./gradlew fetchExistingLibrariesWithNewerVersions --matrixLimit=200
       - name: "🔨 Create branch"
         run: |
           git config --local user.email "actions@github.com"
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -43,8 +43,8 @@ jobs:
         run: |
           git config --local user.email "actions@github.com"
           git config --local user.name "Github Actions"
-          git switch -C libraryVersionsDependabot/$(date '+%Y-%m-%d')
-          git push origin libraryVersionsDependabot/$(date '+%Y-%m-%d')
+          git switch -C check-new-library-versions/$(date '+%Y-%m-%d')
+          git push origin check-new-library-versions/$(date '+%Y-%m-%d')
       - name: "🔨 Create issue"
         id: set-issue
         run: |
@@ -137,6 +137,6 @@ jobs:
         run: |
           git config --local user.email "actions@github.com"
           git config --local user.name "Github Actions"
-          git fetch origin libraryVersionsDependabot/$(date '+%Y-%m-%d')
-          git checkout libraryVersionsDependabot/$(date '+%Y-%m-%d')
+          git fetch origin check-new-library-versions/$(date '+%Y-%m-%d')
+          git checkout check-new-library-versions/$(date '+%Y-%m-%d')
           gh pr create --title "Update supported library versions" --body "This pull request updates supported versions of the existing libraries in the repo"
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -1,5 +1,7 @@
 name: "Check new library versions"
 
+# The workflow runs bi-weekly alternating with the scheduled release workflow. This way we have enough time to provide metadata for failing tests.
+# In case we need more scans, there is a possibility to trigger the workflow manually.
 on:
   schedule:
     - cron: "0 0 8 * *"
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -38,7 +38,7 @@ jobs:
       - name: "🕸️ Populate matrix"
         id: set-matrix
         run: |
-          ./gradlew getExistingLibrariesWithNewVersions
+          ./gradlew getExistingLibrariesWithNewVersions -PlibraryMatrixLimit=200
       - name: "🔨 Create branch"
         run: |
           git config --local user.email "actions@github.com"
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -17,7 +17,7 @@ concurrency:
 jobs:
   get-all-libraries:
     if: github.repository == 'oracle/graalvm-reachability-metadata'
-    name: "📋 Get list of all supported libraries"
+    name: "📋 Get list of all supported libraries with newer versions"
     permissions: write-all
     runs-on: "ubuntu-20.04"
     timeout-minutes: 5
@@ -39,13 +39,13 @@ jobs:
         id: set-matrix
         run: |
           ./gradlew getExistingLibrariesWithNewVersions
-      - name: "Create branch"
+      - name: "🔨 Create branch"
         run: |
           git config --local user.email "actions@github.com"
           git config --local user.name "Github Actions"
           git switch -C libraryVersionsDependabot/$(date '+%Y-%m-%d')
           git push origin libraryVersionsDependabot/$(date '+%Y-%m-%d')
-      - name: "Create issue"
+      - name: "🔨 Create issue"
         id: set-issue
         run: |
           git config --local user.email "actions@github.com"
@@ -109,11 +109,11 @@ jobs:
           export GVM_TCK_LV=$TESTING_VERSION
 
           ./gradlew test -Pcoordinates=${{ env.TEST_COORDINATES }}
-      - name: "New library is supported"
+      - name: "✔️ New library is supported"
         if: success()
         run: |
           sh ./.github/workflows/tryPushVersionsUpdate.sh ${{ matrix.coordinates }} ${{ env.LATEST_VERSION }}
-      - name: "New library is not supported"
+      - name: "❗ New library is not supported"
         if: failure()
         run: |
           git config --local user.email "actions@github.com"
@@ -133,7 +133,7 @@ jobs:
     steps:
       - name: "☁️ Checkout repository"
         uses: actions/checkout@v4
-      - name: "PR for supported versions"
+      - name: "✏️ PR for supported versions"
         run: |
           git config --local user.email "actions@github.com"
           git config --local user.name "Github Actions"
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -106,13 +106,13 @@ jobs:
       - name: "🧪 Run '${{ env.TEST_COORDINATES }}' tests"
         run: |
           TESTING_VERSION=$(echo ${{ matrix.coordinates }} | cut -d ":" -f3)
-          export GVM_TCK_LV=TESTING_VERSION
+          export GVM_TCK_LV=$TESTING_VERSION
 
           ./gradlew test -Pcoordinates=${{ env.TEST_COORDINATES }}
       - name: "New library is supported"
         if: success()
         run: |
-          sh ./.github/workflows/tryPushVersionsUpdate.sh ${{ matrix.coordinates }}
+          sh ./.github/workflows/tryPushVersionsUpdate.sh ${{ matrix.coordinates }} ${{ env.LATEST_VERSION }}
       - name: "New library is not supported"
         if: failure()
         run: |
--- a/.github/workflows/check-new-library-versions.yml
+++ b/.github/workflows/check-new-library-versions.yml
@@ -1,0 +1,144 @@
+name: "Check new library versions"
+
+on:
+  schedule:
+    - cron: "0 0 8 * *"
+    - cron: "0 0 22 * *"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  actions: write
+
+concurrency:
+  group: "workflow = ${{ github.workflow }}, ref = ${{ github.event.ref }}, pr = ${{ github.event.pull_request.id }}"
+  cancel-in-progress: true
+
+jobs:
+  get-all-libraries:
+    if: github.repository == 'oracle/graalvm-reachability-metadata'
+    name: "📋 Get list of all supported libraries"
+    permissions: write-all
+    runs-on: "ubuntu-20.04"
+    timeout-minutes: 5
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+      issue: ${{ steps.set-issue.outputs.issue }}
+    steps:
+      - name: "☁️ Checkout repository"
+        uses: actions/checkout@v4
+      - name: "🔧 Prepare environment"
+        uses: graalvm/setup-graalvm@v1
+        with:
+          java-version: '21'
+          distribution: 'graalvm'
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: "🕸️ Populate matrix"
+        id: set-matrix
+        run: |
+          ./gradlew getExistingLibraries
+      - name: "Create branch"
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "Github Actions"
+          git switch -C libraryVersionsDependabot/$(date '+%Y-%m-%d')
+          git push origin libraryVersionsDependabot/$(date '+%Y-%m-%d')
+      - name: "Create issue"
+        id: set-issue
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "Github Actions"
+          
+          issue_url=$(gh issue create --title "List unsupported library versions" --body "This issue lists unsupported versions of the existing libraries in the repo")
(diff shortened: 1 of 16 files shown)


.github/workflows/check-new-library-versions.yml
```diff
@@ -0,0 +1,142 @@
+name: "Check new library versions"
+
+on:
+  schedule:
+    - cron: "0 0 8 * *"
```
vjovanov, line 5: Is this due to the release schedule? 

Would it be possible to run more often? That way we would have time to react before the release.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,51 @@ def matrixDefault = [
...
+    String group = library.split(":")[0].replace(".", "/")
+    String artifact = library.split(":")[1]
+    def data = new URL(baseUrl + "/" + group + "/" + artifact + "/" + "maven-metadata.xml").getText()
+
+    return tck.getNewerVersionsFromLibraryIndex(data, startingVersion, library)
+}
+
+// gradle getExistingLibrariesWithNewVersions
+Provider<Task> getExistingLibrariesWithNewVersions = tasks.register("getExistingLibrariesWithNewVersions", DefaultTask) { task ->
+    task.setDescription("Returns list of all libraries coordinates")
```
vjovanov, line 117: with the latest versions? Or all coordinates possible?

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,51 @@ def matrixDefault = [
...
+List<String> getNewerVersionsFor(String library, String startingVersion) {
+    def baseUrl = "https://repo1.maven.org/maven2"
+    String group = library.split(":")[0].replace(".", "/")
+    String artifact = library.split(":")[1]
+    def data = new URL(baseUrl + "/" + group + "/" + artifact + "/" + "maven-metadata.xml").getText()
+
+    return tck.getNewerVersionsFromLibraryIndex(data, startingVersion, library)
+}
+
+// gradle getExistingLibrariesWithNewVersions
```
vjovanov, line 115: Would be useful to explain how this function works in comments.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,51 @@ def matrixDefault = [
 
 final String METADATA_GROUP = "Metadata"
 
+List<String> getNewerVersionsFor(String library, String startingVersion) {
+    def baseUrl = "https://repo1.maven.org/maven2"
```
vjovanov, line 107: Are we maven only?

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,51 @@ def matrixDefault = [
...
+            if (!libraryName.contains("samples") && !libraryName.contains("example")) {
+                List<String> versions = getNewerVersionsFor(libraryName, tck.getLatestLibraryVersion(libraryName))
+                versions.forEach {
+                    newerVersions.add(libraryName.concat(":").concat(it))
+                }
+            }
+        }
+
+        // save github resources as workflows has limitation to only accept up to 255 parallel jobs
+        if (newerVersions.size() > 200) {
```
vjovanov, line 137: Please put 200 as a parameter. Don't link the body of this function with current limits on github.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,51 @@ def matrixDefault = [
...
+    task.doFirst {
+        Set<String> libraries = []
+        matchingCoordinates.forEach {
+            libraries.add(it.substring(0, it.lastIndexOf(":")))
+        }
+
+        List<String> newerVersions = new ArrayList<>()
+        libraries.forEach {
+            String libraryName = it
+            if (!libraryName.contains("samples") && !libraryName.contains("example")) {
```
vjovanov, line 128: Hmm, where does this heuristic come from?

vjovanov:
This PR requires a detailed explanation (in the form of docs) on how the new version is computed.

.github/workflows/check-new-library-versions.yml
dnestoro: The idea is to have library updates exactly in the middle between two releases. This way, we are giving some time after the release to aggregate more new library versions but still to have one week to react before the next release.

In my opinion, having more frequent scans (lets say every week) won't help us much, because we won't have time to react in those weeks when we have a scheduled release.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
dnestoro: I thought the name of the task `getExistingLibrariesWithNewVersions` is explainable enough and therefore didn't add more detailed description. Do you think we should extend the description?

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
dnestoro: You mean like a javadoc style comment (before function) or to add comments inside the function to explain step by step?

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
dnestoro: We have tests for the repo infrastructure inside `samples` and `org.example` directories. They are only used to test repository functions like docker image pull, and simple test execution. Therefore, there is no point to search for any "version" of those tests.

.github/workflows/check-new-library-versions.yml
```diff
@@ -0,0 +1,142 @@
...
+      - name: "🧪 Run '${{ env.TEST_COORDINATES }}' tests"
+        run: |
+          TESTING_VERSION=$(echo ${{ matrix.coordinates }} | cut -d ":" -f3)
+          export GVM_TCK_LV=$TESTING_VERSION
+
+          ./gradlew test -Pcoordinates=${{ env.TEST_COORDINATES }}
+      - name: "✔️ New library is supported"
+        if: success()
+        run: |
+          sh ./.github/workflows/tryPushVersionsUpdate.sh ${{ matrix.coordinates }} ${{ env.LATEST_VERSION }}
```
fniephaus, line 115: Is there a reason for this being in a separate file? Also, the shebang of the file says it should run on `bash`, but you are invoking it with `sh`. Not sure this is a good idea.

.github/workflows/check-new-library-versions.yml
```diff
@@ -0,0 +1,142 @@
...
+      - name: "🕸️ Populate matrix"
+        id: set-matrix
+        run: |
+          ./gradlew getExistingLibrariesWithNewVersions -PlibraryMatrixLimit=200
+      - name: "🔨 Create branch"
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "Github Actions"
+          git switch -C libraryVersionsDependabot/$(date '+%Y-%m-%d')
+          git push origin libraryVersionsDependabot/$(date '+%Y-%m-%d')
```
fniephaus, line 47: I see you are using such branch names yourself, but typically, we use kebab-case instead of camelCase. Also, why do you reference dependabot? Does GH's dependabot really do anything here? I guess using the date, also only really works because the jobs are scheduled at midnight. Otherwise, it's possible that the branch is created on a different date before it is checked out by the next job(s).

.github/workflows/tryPushVersionsUpdate.sh
```diff
@@ -0,0 +1,31 @@
...
+set -x
+
+git config --local user.email "actions@github.com"
+git config --local user.name "Github Actions"
+
+BRANCH="libraryVersionsDependabot/$(date '+%Y-%m-%d')"
+git fetch origin "$BRANCH"
+git checkout "$BRANCH"
+
+while [ true ]
```
fniephaus, line 11: while-true-loops make me nervous, especially when I cannot easily understand the steps in this script. I think it would be nice if you add some comments that explain the intention of the git commands.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,59 @@ def matrixDefault = [
 
 final String METADATA_GROUP = "Metadata"
 
+List<String> getNewerVersionsFor(String library, String startingVersion) {
+    def baseUrl = "https://repo1.maven.org/maven2"
+    String group = library.split(":")[0].replace(".", "/")
+    String artifact = library.split(":")[1]
```
fniephaus, line 109: I'd extract the split operation:
```suggestion
    String[] libraryParts = library.split(":")
    String group = libraryParts[0].replace(".", "/")
    String artifact = libraryParts[1]
```

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -351,4 +353,62 @@ List<String> getMetadataFileList(Path directory) throws IOException {
             return foundFiles;
         }
     }
+
+    String getLatestLibraryVersion(String libraryModule) {
+        try {
+            List<String> coordinates = List.of(libraryModule.split(":"));
+            String group = coordinates.get(0);
+            String artifact = coordinates.get(1);
```
fniephaus, line 361: Does this really need to be a List?
```suggestion
            String[] coordinates = libraryModule.split(":");
            String group = coordinates[0];
            String artifact = coordinates[1];
```

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -351,4 +353,62 @@ List<String> getMetadataFileList(Path directory) throws IOException {
...
+            for (MetadataVersionsIndexEntry entry : entries) {
+                allTested.addAll(entry.testedVersions());
+            }
+
+            if (allTested.isEmpty()) {
+                throw new IllegalStateException("Cannot find any tested version for: " + libraryModule);
+            }
+
+            allTested.sort(Comparator.comparing(VersionNumber::parse));
+            return allTested.get(allTested.size() - 1);
```
fniephaus, line 381: this is just `getLast()`, right?
```suggestion
            return allTested.getLast();
```

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
```diff
@@ -0,0 +1,73 @@
...
+    @TaskAction
+    void run() throws IllegalStateException, IOException {
+        List<String> GAVCoordinates = Arrays.stream(coordinates.split(":")).toList();
+        if (GAVCoordinates.size() != 3) {
+            throw new IllegalArgumentException("Maven coordinates should have 3 parts");
+        }
+
+        String group = GAVCoordinates.get(0);
+        String artifact = GAVCoordinates.get(1);
+        String version = GAVCoordinates.get(2);
```
fniephaus, line 51: You found another way to do the same thing ;) 
```suggestion
        String[] coordinatesParts = coordinates.split(":");
        if (coordinatesParts.length != 3) {
            throw new IllegalArgumentException("Maven coordinates should have 3 parts");
        }

        String group = coordinatesParts[0];
        String artifact = coordinatesParts[1];
        String version = coordinatesParts[2];
```

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
dnestoro: `getLast` method on the `ArrayList` was not present in `JDK17`. Therefore, if we ever run something on `JDK17` (and [we do that for our tests](https://github.com/oracle/graalvm-reachability-metadata/blob/master/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle#L97)) we can't use `getLast()`.

dnestoro:
> This PR requires a detailed explanation (in the form of docs) on how the new version is computed.

Maybe we should start some kind of infrastructure changelog?

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,60 @@ def matrixDefault = [
...
+    String[] libraryParts = library.split(":")
+    String group = libraryParts[0].replace(".", "/")
+    String artifact = libraryParts[1]
+    def data = new URL(baseUrl + "/" + group + "/" + artifact + "/" + "maven-metadata.xml").getText()
+
+    return tck.getNewerVersionsFromLibraryIndex(data, startingVersion, library)
+}
+
+// gradle getExistingLibrariesWithNewVersions
+Provider<Task> getExistingLibrariesWithNewVersions = tasks.register("getExistingLibrariesWithNewVersions", DefaultTask) { task ->
```
melix, line 117: :x: Would be cleaner to move this task code to a proper task in the `tck-build-logic`. There's a lot of build logic in this and it's not a good practice to mix build logic with scripts.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -102,6 +103,60 @@ def matrixDefault = [
...
+    task.setDescription("Returns list of all libraries coordinates")
+    task.setGroup(METADATA_GROUP)
+    task.doFirst {
+        if (!project.hasProperty("libraryMatrixLimit")) {
```
melix, line 121: :x: Do not access `project` at task execution time. Instead, define a task and declare inputs/outputs.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
melix: nit: the name `getExistingLibrariesWithNewVersions` is not very Gradle-idiomatic, it sounds like a getter rather than a task name. maybe `fetch` instead of `get`?

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
```diff
@@ -0,0 +1,71 @@
...
+import org.gradle.api.tasks.options.Option;
+import org.gradle.util.internal.VersionNumber;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+
+public abstract class TestedVersionUpdaterTask extends DefaultTask {
```
melix, line 21: :x: This task doesn't declare any input or output. It must declare them properly.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private final ObjectMapper objectMapper = new ObjectMapper()
+            .enable(SerializationFeature.INDENT_OUTPUT)
+            .setSerializationInclusion(JsonInclude.Include.NON_NULL);
+
+    @Option(option = "coordinates", description = "GAV coordinates of the library")
+    void setCoordinates(String coordinates) {
+        this.coordinates = coordinates;
```
melix, line 29: :x: Use an abstract class and declare a `Property<String> getCoordinates()` instead.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+        String group = coordinatesParts[0];
+        String artifact = coordinatesParts[1];
+        String version = coordinatesParts[2];
+        Coordinates c = new Coordinates(group, artifact, version);
+        addToMetadataIndexJson(c);
+    }
+
+    private void addToMetadataIndexJson(Coordinates c) throws IOException {
+        File coordinatesMetadataIndex = getProject().file(CoordinateUtils.replace("metadata/$group$/$artifact$/index.json", c));
```
melix, line 55: :x: Tasks must not access `project` at execution time. Instead the files should be declared as inputs.

melix (changes requested):
Functionally-wise, sounds good to me. The implementation however needs a couple reworks to be Gradle idiomatic (and not break up-to-date checking/configuration).

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/FetchExistingLibrariesWithNewerVersionsTask.groovy
```diff
@@ -0,0 +1,131 @@
...
+import org.gradle.util.internal.VersionNumber
+
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+
+
+abstract class FetchExistingLibrariesWithNewerVersionsTask extends DefaultTask {
+
+    @Input
+    List<String> allLibraryCoordinates
```
melix, line 23: :x: Use `@Input abstract ListProperty<String> getAllLibraryCoordinates()` instead

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/FetchExistingLibrariesWithNewerVersionsTask.groovy
```diff
@@ -0,0 +1,131 @@
...
+
+    @Input
+    @Option(option = "matrixLimit", description = "Sets the maximum number of coordinates in the final matrix")
+    Integer matrixLimit
```
melix, line 27: :x: Use` public abstract Property<Integer> getMatrixLimit()` instead

melix (changes requested):
Almost there, almost there!

melix (approved):

.github/workflows/check-new-library-versions.yml
```diff
@@ -0,0 +1,142 @@
...
+    steps:
+      - name: "☁️ Checkout repository"
+        uses: actions/checkout@v4
+      - name: "✏️ PR for supported versions"
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "Github Actions"
+          git fetch origin check-new-library-versions/$(date '+%Y-%m-%d')
+          git checkout check-new-library-versions/$(date '+%Y-%m-%d')
+          gh pr create --title "Update supported library versions" --body "This pull request updates supported versions of the existing libraries in the repo"
```
vjovanov, line 142: It is nice to have an EOL before EOF.

.github/workflows/check-new-library-versions.yml
vjovanov: Write this down as a comment.

vjovanov (approved):

fniephaus (approved):

PR #642, opened 2025-06-16
Add metadata for upcoming netty-common 4.2.3 build time flag

yawkat:
https://github.com/netty/netty/pull/15325 introduced code that works around a bug in GraalVM 19-25 where native image would fail with a linker error. The fix requires the class CleanerJava25 to be initialized at build time, but the native-image.properties to do this in the netty repo is overridden by graalvm-reachability-metadata.

This PR adds the same flag to this repo for the upcoming 4.2.3 release. Because the flag is actually counterproductive on the current netty release – with it, the build will fail even on GraalVM versions like 17 that don't have the bug in the first place – I can't add it to the existing metadata, and need to create a new version instead.

Unfortunately, the native-image.properties don't seem to be applied by the build plugin, so this is a draft. Any suggestions, @melix?


yawkat:
@melix then we have a problem. This issue makes netty 4.2.2 stop working for all users of the reachability metadata.

violetagg:
Adding @sdeleuze

yawkat:
@violetagg GraalVM has agreed to backport a fix to 21 that will make this patch unnecessary

violetagg:
thanks

Issue #540, opened 2024-10-04
Make vulnerability scanner less strict when updating existing docker images


dnestoro:
We should keep docker images scanner as it is for new docker images but change it when we update existing docker images in the allowed images list. Improved solution would mean that we accept updates for docker images if the updated image has less vulnerabilities than the existing one.

PR #669, opened 2025-09-05
Make vulnerabilities scanner less strict

dnestoro:
## What does this PR do?

Make vulnerabilities scanner less strict in order to accept slightly vulnerable images that are less vulnerable than the currently existing allowed images.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/540

### DIFF
--- a/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
+++ b/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
@@ -51,13 +51,7 @@ public abstract class GrypeTask extends DefaultTask {
         }
 
         public boolean isLessVulnerable(DockerImage other) {
-            // first check number of critical vulnerabilities
-            if (this.vulnerabilities.critical() < other.vulnerabilities().critical()) {
-                return true;
-            }
-
-            // if number of critical vulnerabilities is the same => check number of high vulnerabilities
-            return this.vulnerabilities.critical() == other.vulnerabilities().critical() && this.vulnerabilities.high() < other.vulnerabilities().high();
+            return this.vulnerabilities.critical() < other.vulnerabilities().critical() && this.vulnerabilities.high() < other.vulnerabilities().high();
         }
 
         public void printVulnerabilityStatus() {
--- a/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
+++ b/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
@@ -1,5 +1,13 @@
 package org.graalvm.internal.tck;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.type.TypeReference;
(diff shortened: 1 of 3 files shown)


tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
```diff
@@ -33,88 +36,190 @@ void setNewCommit(String newCommit) {
...
+        }
 
-        if (diffFiles.isEmpty()) {
-            throw new RuntimeException("There are no changed or new docker image founded. " +
-                    "This task should be executed only if there are changes in allowed-docker-images directory.");
+        public boolean isVulnerableImage() {
+            return vulnerabilities.critical() > 0 || vulnerabilities.high() > 0;
+        }
+
+        public boolean isLessVulnerable(DockerImage other) {
```
matteoldani, line 64: I do agree that, in general, a critical vulnerability has more priority that any number of high ones. However, since we are not just comparing two different scenarios but doing an upgrade, how about introducing a sanity check that prevents us from introducing a container with 1 less critical and thousands more high-scored vulns without noticing it? 

After all, if we decide on making such a tradeoff, we can always adapt the heuristic.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
```diff
@@ -33,88 +36,190 @@ void setNewCommit(String newCommit) {
...
+        getExecOperations().exec(execSpec -> {
+            execSpec.setExecutable("/bin/sh");
+            execSpec.setArgs(List.of(command));
+            execSpec.setStandardOutput(execOutput);
+        });
 
-                if (numberOfHigh > 4 || numberOfCritical > 0) {
-                    shouldFail = true;
+        // parse Grype output
+        ByteArrayInputStream inputStream = new ByteArrayInputStream(execOutput.toByteArray());
```
matteoldani, line 157: The grype output is in json format. Can't we parse it using a json parse? It would make the vulns detection more reliable to output changes.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
dnestoro: I think the only solution that is correct in 100% cases is that we only accept a new image if it contains less critical **and** less high vulnerabilities. In all other combinations we can't determine programmatically if the image is acceptable candidate for update or not (we will still rely on a manual check of the image).
What do you think?

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
dnestoro: I refactored that part :heavy_check_mark: 
Please take a look again

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
matteoldani: It is definitely better from security point of view if we can enforce that and we have the time for manual checking the one that break the policy 👍🏼 

I would hope that most, if not all, the cases will pass the new policy. Have you ever checked manually?

matteoldani (approved):

Issue #672, opened 2025-09-09
Run all jobs for testing new libraries versions


dnestoro:
At the moment we can run only 250 jobs per workflow at the same time because of the GitHub job matrix limitations. With this restriction [processing of newer versions of libraries](https://github.com/oracle/graalvm-reachability-metadata/blob/master/.github/workflows/check-new-library-versions.yml#L1) that we have in the repository is slow and we are processing 250 new versions on a bi-weekly base.

Idea: because the limitation applies on a workflow level, we can have:
- one workflow that knows the full list of libraries versions that we want to test, and orchestrates (spawns) other workflows to run the batches
- workflow that would accept batch of 250 jobs and execute them in parallel

This way, the orchestration workflow can dynamically spawn many worker workflows that will execute batches of tests (which will run in parallel).

PR #675, opened 2025-09-10
Run all tests for new libraries versions checking at once

dnestoro:
## What does this PR do?

Implements: https://github.com/oracle/graalvm-reachability-metadata/issues/672


.github/workflows/new-libraries-orchestrator.yml
```diff
@@ -0,0 +1,66 @@
...
+            echo "Items for this batch: $current_batch"
+          
+            branch_name="check-new-library-versions/batch-id-$batch_id/$(date '+%Y-%m-%d')"
+          
+            git checkout master
+            git switch -C "$branch_name"
+            git push origin "$branch_name"
+
+            gh workflow run check-new-library-versions-in-batch.yml --ref main -f items="$current_batch" -f batch_id="$batch_id"
+            sleep 120
```
vjovanov: Why sleep?

dnestoro: During some experiments on my fork of the repository, I noticed that if I spawn all jobs at once, some of the jobs appear to be frozen. With this small delay, I am allowing GitHub some time to organize the jobs properly. 

Also, if two workflows run the same group of tests (for example, Hibernate tests with different versions), it can create a race condition when creating GitHub issues. 

Note that this might be because my fork is a private repository, and we may not see this behavior on this repo since it is public and under the organization. 

Overall, I think this change will not make us wait much longer, and it can be beneficial in many ways.

vjovanov: Would 30 or 60 seconds work?

.github/workflows/check-new-library-versions-in-batch.yml
```diff
@@ -1,92 +1,100 @@
...
+  group: "workflow=${{ github.workflow }},ref=${{ github.event.ref }},pr=${{ github.event.pull_request.id }}"
   cancel-in-progress: true
 
 jobs:
   get-all-libraries:
-    if: github.repository == 'oracle/graalvm-reachability-metadata'
     name: "📋 Get list of all supported libraries with newer versions"
-    permissions: write-all
-    runs-on: "ubuntu-22.04"
+    if: github.repository == 'oracle/graalvm-reachability-metadata'
```
vjovanov, line 19: Why do we need this? What if I fork this repo?

dnestoro: Previously this workflow was scheduled for execution every two weeks, and therefore we didn't want this to be executed every two weeks on user's forks (we would spend their repo resources that way). We can remove it now if necessary, but I don't see a reason why this job should be executed anywhere else than in this repo.

vjovanov: Interesting, just leave it as a comment. Is this standard practice?

Will this prevent me from running this task manually on my fork, and should it?

.github/workflows/run-consecutive-tests.sh
```diff
@@ -0,0 +1,47 @@
...
+# Parse versions with jq
+readarray -t VERSIONS < <(echo "$VERSIONS_JSON" | jq -r '.[]')
+
+for VERSION in "${VERSIONS[@]}"; do
+  echo "Running test with GVM_TCK_LV=$VERSION and coordinates=$TEST_COORDINATES"
+  GVM_TCK_LV="$VERSION" ./gradlew test -Pcoordinates="$TEST_COORDINATES"
+  RESULT=$?
+  if [ "$RESULT" -eq 0 ]; then
+    PASSED_VERSIONS+=("$VERSION")
+    echo "PASSED:$VERSION"
```
vjovanov, line 37: It would be interesting to retry the runs in case we have flaky tests. Maybe try up to 3 times?

vjovanov:
Why not run this every week, after the initial manual run?

.github/workflows/check-new-library-versions-in-batch.yml
dnestoro: Yes, that will prevent you from running this workflow on your fork (that is the purpose of this). As long as this is not a scheduled, but manually triggered job, we can remove this line. If we decide to make it scheduled at some point, I will keep this line.

.github/workflows/run-consecutive-tests.sh
dnestoro: Is it worth doing that? In most cases we will repeat actually failing test for 3 times just to make sure that it really fails. On the other hand, a flaky test will fail rarely/occasionally, and if that really happens the next iteration of this job will run the test properly.

dnestoro:
> Why not run this every week, after the initial manual run?

Once we establish a quick way for fixing many tests at once, we can make this scheduled. Until then, running tests every week, while we only fixed 1 or 2 tests, will just tell us that we have same failures every week (while we spend so much resources doing that). So in my opinion, that is something for TODO list here. Once we have proper way of bulk fixing tests, we can easily make this job scheduled.

vjovanov:
Continued in https://github.com/oracle/graalvm-reachability-metadata/pull/708

PR #713, opened 2025-11-07
Bump Gradle version to 9.1.0 to allow using GraalVM 25

kimeta:
## What does this PR do?

Implements: #709 

Upgrade Gradle to 9.1.0 to enable support for GraalVM 25.

### DIFF
--- a/gradle/checkstyle.xml
+++ b/gradle/checkstyle.xml
@@ -89,7 +89,6 @@
             <property name="ignoreConstructorParameter" value="true"/>
             <metadata name="net.sf.eclipsecs.core.lastEnabledSeverity" value="inherit"/>
         </module>
-        <module name="FinalClass"/>
         <module name="HideUtilityClassConstructor">
             <property name="severity" value="ignore"/>
             <metadata name="net.sf.eclipsecs.core.lastEnabledSeverity" value="inherit"/>
--- a/tests/src/com.google.protobuf/protobuf-java-util/3.21.12/build.gradle
+++ b/tests/src/com.google.protobuf/protobuf-java-util/3.21.12/build.gradle
@@ -7,7 +7,7 @@
 
 plugins {
     id "org.graalvm.internal.tck"
-    id "com.google.protobuf" version "0.9.1"
+    id "com.google.protobuf" version "0.9.5"
 }
 
 String libraryVersion = tck.testedLibraryVersion.get()
--- a/tests/src/io.grpc/grpc-core/1.69.0/build.gradle
+++ b/tests/src/io.grpc/grpc-core/1.69.0/build.gradle
@@ -7,7 +7,7 @@
 
 plugins {
     id "org.graalvm.internal.tck"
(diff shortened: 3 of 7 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.10-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.1.0-bin.zip
```
vjovanov, line 3: Do we need to bump the `gradlew` script since it is a major version?

kimeta: Yes, sorry. We need to change `gradlew.bat`, `gradle-wrapper.jar` and `gradlew` as well.

tests/src/io.netty/netty-common/4.1.115.Final/src/test/java/netty/NettyTests.java
```diff
@@ -257,7 +257,7 @@ public void initChannel(SocketChannel ch) {
         }
     }
 
-    private static class HttpServerHandler extends SimpleChannelInboundHandler<Object> {
+    private static final class HttpServerHandler extends SimpleChannelInboundHandler<Object> {
```
vjovanov: Why do we have to change the tests for `gradle`?

kimeta: When we upgraded `gradle` version to 9.1.0, default version of `checkstyle` upgraded,  which caused some of our test code to fail the `checkstyleTest` task:
```
> Task :checkstyleTest FAILED

[ant:checkstyle] [ERROR] ...tests/src/io.netty/netty-common/4.1.80.Final/src/test/java/netty/NettyTests.java:246:5: Class HttpServerHandler should be declared as final. [FinalClass]
```
All changes in the tests are just to fix these failures.

melix: lol: https://mastodon.xyz/@melix/115490434484079017

melix (approved):
Note that this will require to be more careful when making changes since we're going to compile against newer Gradle APIs, so we need to be extra careful with compatibility with older Gradle releases.

vjovanov (approved):

gradle/checkstyle.xml
```diff
@@ -89,7 +89,6 @@
             <property name="ignoreConstructorParameter" value="true"/>
             <metadata name="net.sf.eclipsecs.core.lastEnabledSeverity" value="inherit"/>
         </module>
-        <module name="FinalClass"/>
```
vjovanov, line 92: We can't have such rules as it will cause issues with AI-generated tests in the future.

Issue #770, opened 2025-11-13
Bug: Vulnerability scanner fails when the updated docker image has the same number of vulnerabilities as the old version


jormundur00:
### Describe the bug

In the current implementation of the Vulnerability scanner, `org.graalvm.internal.tck.GrypeTask#scanChangedImages` throws a `IllegalStateException` when the number of vulnerabilities for the newer docker image version is equal to the number of vulnerabilities on the current version. In this case, we should allow the version upgrade and not fail the task.

Also, `org.graalvm.internal.tck.GrypeTask#getAllowedImagesFromMaster` always fails when its called in the GitHub actions CI, as the action-fetched repository only contains the `origin/master` branch, while the method looks for a non-existent `master` branch.

### Steps to reproduce

```markdown
Both issues are discovered in the CI by the attempt to bump up the version of the `postgres` docker image from `16-alpine` to `18-alpine`.

The non-existent `master` branch issue hides the vulnerability equality issue, with the workflow producing this build:

https://github.com/oracle/graalvm-reachability-metadata/actions/runs/19335156151/job/55307954645#step:5:66

While when replacing the `master` check with `origin/master` in the scanner, we can get the original issue:

https://github.com/oracle/graalvm-reachability-metadata/actions/runs/19333907567/job/55303630330#step:5:78
```

### Expected behavior

The Vulnerability scanner should only fail if the updated version of the docker image contains strictly more vulnerabilities than the current version. We should accept version bumps that have an equal amount of vulnerabilities as the current version.

### Logs and error messages

```text
Issue 1:

Run ./gradlew checkAllowedDockerImages --baseCommit=5a4e6a97fc87b2bedde4aeab07abe1d4ffe06ce9 --newCommit=74456c386b5eec7dec5de2a41ac054846e506744
Downloading https://services.gradle.org/distributions/gradle-9.1.0-bin.zip
............10%.............20%.............30%.............40%.............50%............60%.............70%.............80%.............90%.............100%

Welcome to Gradle 9.1.0!

Here are the highlights of this release:
 - Full Java 25 support
 - Native task graph visualization
 - Enhanced console output

For more details see https://docs.gradle.org/9.1.0/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)
> Task :tck-build-logic:extractPluginRequests
> Task :tck-build-logic:generatePluginAdapters
> Task :tck-build-logic:compileJava
> Task :tck-build-logic:compileGroovy
[...]

PR #772, opened 2025-11-14
Fix vulnerability scanner not working in the GitHub actions CI and make it less strict

jormundur00:
## What does this PR do?

In this PR, we make the vulnerability scanner less strict by accepting newer versions of docker images if they have less or equal vulnerabilities to the current version (while currently we only accept if they have strictly less vulnerabilities).

Also in this PR, we fix `org.graalvm.internal.tck.GrypeTask#getAllowedImagesFromMaster` not working in the GitHub actions CI. Previously, whenever this method is reached in the CI, the task would fail with: `fatal: invalid object name 'master'.`. This happens because the `master` branch does not exist in the CI, while `origin/master` does. The change in this PR just made it so we check the `origin/master` when we fetch allowed images.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/770

### DIFF
--- a/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
+++ b/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
@@ -57,7 +57,7 @@ public abstract class GrypeTask extends DefaultTask {
             return vulnerabilities.critical() > 0 || vulnerabilities.high() > 0;
         }
 
-        public boolean isNoMoreVulnerable(DockerImage other) {
+        public boolean isNotMoreVulnerable(DockerImage other) {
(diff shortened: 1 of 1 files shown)


tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/GrypeTask.java
```diff
@@ -61,8 +57,8 @@ public boolean isVulnerableImage() {
             return vulnerabilities.critical() > 0 || vulnerabilities.high() > 0;
         }
 
-        public boolean isLessVulnerable(DockerImage other) {
-            return this.vulnerabilities.critical() < other.vulnerabilities().critical() && this.vulnerabilities.high() < other.vulnerabilities().high();
+        public boolean isNoMoreVulnerable(DockerImage other) {
```
vjovanov: LessEqual, or NotMore

Issue #754, opened 2025-11-10
Add filtering for failing library versions when creating issues


kimeta:
We want to avoid creating tickets when a library upgrade fails during automation if it’s a pre-release or early-access version we’re intentionally skipping for now. Since Maven version labels aren’t standardized, we need to use regex matching to reliably filter out these temporary or experimental releases.
Regex pattern that catch these cases (case-insensitive):
{coordinates}`(-|\.)((?i)rc|cr|m|ea|beta\d+|alpha\d+)`

Issue #4, opened 2022-07-06
Remove index.json from the project root and from the individual test folders


vjovanov:

linghengqian:
- Is this issue ongoing? It seems that https://github.com/oracle/graalvm-reachability-metadata/pull/1 is not set to be associated with this issue.

vjovanov:
Yes, it is still planned and the work will be continued in the near future.

PR #773, opened 2025-11-14
Filter out pre-release versions of libraries when a full release is present

jormundur00:
## What does this PR do?

In this PR we filter out early-access/pre-release library versions from automated ticket creation, when their version upgrade fails. Also in this PR, we introduce the automated removal of pre-release versions from the `tested-versions` of a libraries `index.json` file, when the full version of that specific version is added to the same (e.g.: `6.6.0.rc2` would be removed once `6.6.0` or `6.6.0.Final` is added), as well as automated renaming of the metadata and tests directories for the library once its version is changed.

The implementation follows a 2 step process:

1. When fetching new library versions from maven (by reading its `maven-metadata.xml`) we filter out pre-release versions if their full release version is present in the same file,
2. Later, when adding a new version to a libraries `index.json` `tested-libraries` field, if its not a pre-release version we remove all pre-releases for that version from the same field. If the `metadata-version` field for that library is a pre-release version, and the full release for it is added to the `tested-libraries`, we also replace the `metadata-version` field to refer to the full release and also rename the metadata and tests directories to refer to the same version (also updating the `gradle.properties` file of the test directory). If test directories are renamed, the test root `index.json` is also updated to reflect the new version.

We determine if a library version is a pre-release by regex pattern matching, determining if the version suffix matches one of the pre-release patterns. Currently, we filter out versions that end with a `.` or `-` following one of the following case insensitive suffixes:

- `alpha` with optional numbers after it, or a `.` or `-` separated string,
- `beta` with optional numbers after it, or a `.` or `-` separated string,
- `rc` with optional numbers after it, or a `.` or `-` separated string,
- `cr` with optional numbers after it, or a `.` or `-` separated string,
- `m` with non-optional numbers after it, or a `.` or `-` separated string,
- `ea` with optional numbers after it, or a `.` or `-` separated string,
- `b` with non-optional numbers after it, or a `.` or `-` separated string,
- `preview` with optional numbers after it, or a `.` or `-` separated string,
- specially if separated by `-`: an array of numbers.

Implements: https://github.com/oracle/graalvm-reachability-metadata/issues/754

### DIFF
--- a/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
+++ b/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
@@ -33,9 +33,9 @@ import java.util.regex.Pattern;
 @SuppressWarnings("unused")
 public abstract class TestedVersionUpdaterTask extends DefaultTask {
     /**
-     * Identifies pre-release library versions by pattern matching against common pre-release suffixes.
+     * Identifies library versions, including optional pre-release and ".Final" suffixes.
      * <p>
-     * A version is considered pre-release if its suffix (following the last '.' or '-') matches
+     * A version is considered a pre-release if it has a suffix (following the last '.' or '-') matching
      * one of these case-insensitive patterns:
      * <ul>
      *   <li>{@code alpha} followed by optional numbers (e.g., "alpha", "Alpha1", "alpha123")</li>
@@ -48,11 +48,12 @@ public abstract class TestedVersionUpdaterTask extends DefaultTask {
      *   <li>{@code preview} followed by optional numbers (e.g., "preview", "preview1", "preview42")</li>
      *   <li>Numeric suffixes separated by '-' (e.g., "-1", "-123")</li>
      * </ul>
+     * <p>
+     * Versions ending with ".Final" are treated as full releases of the base version.
      */
-    private static final Pattern PRE_RELEASE_PATTERN = Pattern.compile(
-            "(?i)^(\\d+(?:\\.\\d+)*)(?:[-.](alpha\\d*|beta\\d*|rc\\d*|cr\\d*|m\\d+|ea\\d*|b\\d+|\\d+|preview)(?:[-.].*)?)?$"
+    private static final Pattern VERSION_PATTERN = Pattern.compile(
+            "(?i)^(\\d+(?:\\.\\d+)*)(?:\\.Final)?(?:[-.](alpha\\d*|beta\\d*|rc\\d*|cr\\d*|m\\d+|ea\\d*|b\\d+|\\d+|preview)(?:[-.].*)?)?$"
     );
-    private static final Pattern FINAL_PATTERN = Pattern.compile("(?i)\\.Final$");
 
     @Option(option = "coordinates", description = "GAV coordinates of the library")
     void extractInformationFromCoordinates(String c) {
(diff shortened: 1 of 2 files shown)


.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -217,6 +217,14 @@ jobs:
           git config --local user.name "Github Actions"
 
           FAILED_VERSION="${{ steps.runtests.outputs.failed_version }}"
+          
+          # Skip issue creation if failed version is a pre-release
```
vjovanov: Should we even run the tests for those versions?

.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -217,6 +217,14 @@ jobs:
...
+          shopt -s nocasematch
+          if [[ "$FAILED_VERSION" =~ (-|\.)(rc|cr|m|ea|beta|alpha)(-?[0-9]*) ]]; then
```
vjovanov: Does does capture all of the cases? I have seen all kinds of stuff in the versions: `13.1.1.jre8-preview`, `1.8.0-343`, and others.

Also, there should be an end of string at the end, right?

tests/src/index.json
jvukicev, line 1: This file used to have a weird indentation format, that default jackson/java formatters don't use. When updating this file, the JSON is reformatted to use regular indentation.

Since this file should be removed once https://github.com/oracle/graalvm-reachability-metadata/issues/4 is implemented, I would update the format here, so any PR's in between don't massively change this file like this PR.

.github/workflows/verify-new-library-version-compatibility.yml
jvukicev: Based on offline discussion, I've fully refactored this PR to match the newly-decided functionality, so we no longer do these checks here.

.github/workflows/verify-new-library-version-compatibility.yml
jvukicev: I've updated the pattern to include more different pre-release versions. All the covered pre-release identifiers are listed in the PR description.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
```diff
@@ -19,11 +19,21 @@
...
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public abstract class TestedVersionUpdaterTask extends DefaultTask {
+    private static final Pattern PRE_RELEASE_PATTERN = Pattern.compile(
+            "(?i)^(\\d+(?:\\.\\d+)*)(?:[-.](alpha\\d*|beta\\d*|rc\\d*|cr\\d*|m\\d+|ea\\d*|b\\d+|\\d+|preview)(?:[-.].*)?)?$"
```
vjovanov: All the wisdom of Maven in one regex ;)

vjovanov: Explain it here like in the PR description.

jormundur00:
> We should run this after every library update as the final step. And add the results as the final commit. Please see:
> 
> https://github.com/oracle/graalvm-reachability-metadata/blob/master/.github/workflows/verify-new-library-version-compatibility.yml

We already do this in this PR. The `addTestedVersion` task called inside of `verify-new-library-version-compatibility.yml` (inside of `update_new_versions()`) filters out old pre-release versions when a new version is added (if that new version is the full release for the given pre-release), and renames the metadata and test directories and their references to point to the full release version, if applicable.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
```diff
@@ -21,12 +21,39 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 @SuppressWarnings("unused")
 public abstract class TestedVersionUpdaterTask extends DefaultTask {
```
vjovanov, line 34: Please test this feature on our sample. Add some stuff there and expect that all of these versions will be deleted.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java
```diff
@@ -76,4 +108,169 @@ void run() throws IllegalStateException, IOException {
...
+     *   <li>If the entry's {@code metadataVersion} is a pre-release of the same base version,
+     *       it is updated to the new full release. The corresponding metadata and test directories are renamed accordingly.
+     *       The {@code gradle.properties} file in the tests directory is updated to refer to the new version.</li>
+     *   <li>Version parsing follows {@link #PRE_RELEASE_PATTERN} and treats ".Final" as a base version.</li>
+     * </ul>
+     */
+    private MetadataVersionsIndexEntry handlePreReleases(MetadataVersionsIndexEntry entry, String newVersion, Path baseDir) throws IOException {
+        // strip .Final if present
+        String cleanedNewVersion = FINAL_PATTERN.matcher(newVersion).replaceAll("");
+        Matcher newVersionMatcher = PRE_RELEASE_PATTERN.matcher(cleanedNewVersion);
```
kimeta: Can version contain both Final and pre-release tag?

jvukicev: AFAIK they should be mutually exclusive tags. Previously, we did this ".Final" strip as `PRE_RELEASE_PATTERN` needed to match full releases as well (and we made a distinction if the matched version is a pre release by the presence of the patterns group 2). As this is probably confusing, I unified both of the patterns and named the unified pattern `VERSION_PATTERN`, so we no longer have to do this strip. Good catch, thanks!

Issue #760, opened 2025-11-12
Better labels for [Automation] tickets


vjovanov:
Currently it is not easy to distinguish different categories of failures.

PR #777, opened 2025-11-14
Add proper labeling and java run stage

vjovanov:
## What does this PR do?

Does smarter bisecting when running automated tests. Check if it works first, and only if it fails do the bisecting. Also, adds a new category `java run` and applies labels.

Also, increases the frequency of running the testing job to daily.

Fixes: #760

### DIFF
--- a/.github/workflows/verify-new-library-version-compatibility.yml
+++ b/.github/workflows/verify-new-library-version-compatibility.yml
@@ -255,7 +255,6 @@ jobs:
     permissions: write-all
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-    if: ${{ always() }}
     needs:
       - get-all-libraries
       - test-all-metadata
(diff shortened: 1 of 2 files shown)


.github/workflows/test-all-metadata.yml
```diff
@@ -15,7 +15,6 @@ concurrency:
 
 jobs:
   get-all-metadata:
-    if: github.repository == 'oracle/graalvm-reachability-metadata'
```
jvukicev: Do we want to run this job on master push on forks (but not the `test-all-metadata` and `all-metadata-passed`)?

.github/workflows/test-changed-metadata.yml
```diff
@@ -14,7 +14,6 @@ concurrency:
 
 jobs:
   get-changed-metadata:
-    if: github.repository == 'oracle/graalvm-reachability-metadata'
```
jvukicev: Same question as for `test-all-metadata.yml`.

.github/workflows/test-all-metadata.yml
vjovanov: This one is not automatic, just manual. Although I have a follow up PR:
https://github.com/oracle/graalvm-reachability-metadata/pulls

jvukicev: But it should still trigger on pushes to master (even on a fork), no? The `test-all-metadata` and `all-metadata-passed` jobs are still behind `if: github.repository == 'oracle/graalvm-reachability-metadata'` checks, but the `get-all-metadata` job isn't. Is this by design?

Issue #681, opened 2025-09-25
Split workflow that runs tests into multiple workflows each responisble for specific JDK version


dnestoro:
Currently we have two workflows: `test-all-metadata.yml` and `test-changed-metadata.yml`. There are several things that we can improve:

Reduce the number of different workflows.

- Remove `test-all-metadata` workflow. 
- This is okay because `test-changed-metadata` will test all metadata in case when we changed some logic in tck or when we changed GitHub workflows ([see this](https://github.com/oracle/graalvm-reachability-metadata/blob/master/tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java#L150)). Because of that, `test-changed-metadata` will always run proper amount of tests.

Make workflows JDK specific

- We can make a template for `test-changed-metadata` which accepts JDK version as a parameter.
  -  That way, we can remove `versions` property from job matrix ([here](https://github.com/oracle/graalvm-reachability-metadata/blob/master/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle#L109)), and simplify the logic by simply removing [this](https://github.com/oracle/graalvm-reachability-metadata/blob/master/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle#L156-L170) calculation.
  - We can have workflows dedicated to specific JDK (`17`, `21`, `25`, `latest-ea`) which will just run the template with corresponding version

PR #778, opened 2025-11-14
Make tests run under the 256 tests limit; give this repo the ❤️ it deserves.

vjovanov:
1. Introduce `ci.json` where we store the matrix of JDK and OS and flags. If this file is changed, run all tests in batches.
2. `test-changed-metadata.yml` should just check if metadata changes.
3. `test-changed-infrastructure.yml` should test only a representative set of libraries.
4. All commands now support a `-Pcoordinates=k/n` mode for batched execution.
5. Give this repo the ❤️ it deserves and prepare it for AI coding.
6. Abandon JDK 17. It will not scale for the incoming libraries.
7. One stop shop testing command: `./gradlew testAllParallel -Pparallelism=n`
8. Documentation and structure cleanup. Start review with: 
    - Look at `docs/DEVELOPING.md`
    - Look at `docs/CI.md`
    - Look at `AGENTS.md`

Fixes: #681

### DIFF
--- a/ci.json
+++ b/ci.json
@@ -1,19 +1,19 @@
 {
-  "buildArgs": ["-verbose", "-Ob"],
+  "buildArgs": ["-verbose", "-Ob", "-H:+ReportExceptionStackTraces"],
   "generateMatrixBatchedCoordinates": {
-    "java": ["17", "21", "25", "latest-ea"],
+    "java": ["21", "25", "latest-ea"],
     "os": ["ubuntu-latest"]
   },
   "generateChangedCoordinatesMatrix": {
-    "java": ["17", "21", "25", "latest-ea"],
+    "java": ["21", "25", "latest-ea"],
     "os": ["ubuntu-latest"]
   },
   "generateInfrastructureChangedCoordinatesMatrix": {
-    "java": ["17", "21", "25", "latest-ea"],
+    "java": ["21", "25", "latest-ea"],
     "os": ["ubuntu-latest"]
   },
   "generateMatrixMatchingCoordinates": {
-    "java": ["17", "latest-ea"],
+    "java": ["21", "25","latest-ea"],
     "os": ["ubuntu-latest"]
   }
 }
--- a/tests/src/org.jetbrains.kotlin/kotlin-reflect/1.7.10/build.gradle
+++ b/tests/src/org.jetbrains.kotlin/kotlin-reflect/1.7.10/build.gradle
@@ -17,7 +17,7 @@ dependencies {
 }
 
 compileTestKotlin {
-    kotlinOptions.jvmTarget = "17"
+    kotlinOptions.jvmTarget = "21"
 }
 
 graalvmNative {
--- a/tests/src/org.jetbrains.kotlin/kotlin-stdlib/1.7.10/build.gradle
+++ b/tests/src/org.jetbrains.kotlin/kotlin-stdlib/1.7.10/build.gradle
@@ -20,7 +20,7 @@ dependencies {
 }
 
 compileTestKotlin {
-    kotlinOptions.jvmTarget = "17"
+    kotlinOptions.jvmTarget = "21"
 }
 
 graalvmNative {


.github/workflows/test-changed-infrastructure.yml
```diff
@@ -0,0 +1,102 @@
...
+          java-version: '21'
+
+      - name: "🕸️ Populate matrix"
+        id: set-matrix
+        run: |
+          ./gradlew generateInfrastructureChangedCoordinatesMatrix -PbaseCommit=${{ github.event.pull_request.base.sha }} -PnewCommit=${{ github.event.pull_request.head.sha }}
+
+  test-changed-infrastructure:
+    name: "🧪 ${{ matrix.coordinates }} (GraalVM for JDK ${{ matrix.version }} @ ${{ matrix.os }})"
+    if: needs.get-changed-infrastructure.result == 'success' && needs.get-changed-infrastructure.outputs.none-found != 'true' && github.repository == 'oracle/graalvm-reachability-metadata'
```
jvukicev, line 45: I don't think we need to check `github.repository == 'oracle/graalvm-reachability-metadata'` here, as if `get-changed-infrastructure` is skipped `test-changed-infrastructure` will also be skipped due to `needs.get-changed-infrastructure`.

.github/workflows/test-changed-infrastructure.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: "🧪 Run '${{ matrix.coordinates }}' tests"
+        run: |
+          ./gradlew test -Pcoordinates=${{ matrix.coordinates }}
+
+  all-infrastructure-passed:
+    name: "🧪 All build-logic triggered tests have passed"
+    runs-on: "ubuntu-22.04"
+    timeout-minutes: 1
+    if: ${{ always() }} && github.repository == 'oracle/graalvm-reachability-metadata'
```
jvukicev: Don't think we need the repository check here either, as it's already made if the needed prerequisite job is ran.

.github/workflows/test-changed-metadata.yml
```diff
@@ -4,9 +4,9 @@ on:
   pull_request:
     branches:
       - master
-    paths-ignore:
-      - '**.md'
-      - 'library-and-framework-list*.json'
+    paths:
+      - 'metadata/*'
```
jvukicev: Are we sure we want to check `metadata/*` and not `metadata/**` here? Not an expert on wokflow syntax, but as I understand the wildcard definitions [here](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet), `*` should only include direct `metadata` directory descendants, while `metadata/**` should include nested directories too.

If that's correct, we should probably change that in other places where `/*` is used.

.github/workflows/test-changed-infrastructure.yml
vjovanov: Fixed.

.github/workflows/test-changed-metadata.yml
vjovanov: Fixed.

.github/workflows/scan-docker-images.yml
```diff
@@ -18,18 +18,22 @@ jobs:
...
         with:
-          java-version: '17'
           distribution: 'graalvm'
+          java-version: '21'
           github-token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: "Install required tools"
         run: |
-          curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sudo sh -s -- -b /usr/local/bin
+          curl -sSfL https://get.anchore.io/grype/v0.104.0/install.sh | sudo sh -s -- -b /usr/local/bin
```
melix, line 30: Not different from the previous implementation, but running arbitrary shell commands from remote makes me kind of nervous. Maybe we should at least have a checksum to ensure the remote repo hasn't been compromised.

Issue #784, opened 2025-11-18
Feature: Automatic updating of docker image versions through the dependabot


jormundur00:
### Is your feature request related to a problem?

Currently, the dependabot only changes the version of a docker image in one place, requiring manual updating of the version wherever it is referenced.

### Describe the solution you'd like

The dependabot should open review-ready PR's when suggesting updates for docker image versions, probably grepping through the docker-related directories for the old version and replacing it with the new one.

PR #785, opened 2025-11-19
Add syncing of new docker image versions upon dependabot PR creation

jormundur00:
## What does this PR do?

In this PR we introduce the fully automated updating of docker image versions upon dependabot updates. Previously, in the dependabot opened docker image version update PR's, the version was only updated in `/tests/tck-build-logic/src/main/resources/allowed-docker-images`, while all other references to the docker image had to be manually replaced. With this PR, the newly-added GitHub actions workflow will automatically look through the repository for references to the old image version and replace them with the new version. The replacements it makes are then commited to the dependabot PR (in a separate commit), and the resulting PR should be merge-ready.

These dependabot PR's will still be assigned to me. @jormundur00, so I will still look at their correctness, but this will save time on manual labor.

This workflow has been tested on a fork for the current outdated versions and it does everything that we previously did manually for them: https://github.com/jormundur00/graalvm-reachability-metadata/pulls.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/784.


.github/workflows/sync-docker-images.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - name: "☁️ Checkout repository"
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
+          fetch-depth: 0
+
+      - name: "📝 Sync docker image tags"
+        run: |
+          set -euo pipefail
```
vjovanov, line 25: GPT 5 :)

Issue #808, opened 2025-11-21
Feature: We need a simple copy-paste script that allows to check if a maven artifact is supported


vjovanov:
### Is your feature request related to a problem?

It is hard for people to know if a library is supported or not by the metadata repo. This would allow them to immediately check that.

### Describe the solution you'd like

That script would live in the metadata repo and it would be listed on in the `README.md` for quick checking.

PR #802, opened 2025-11-21
Make the README.md simple and clear

vjovanov:
Fixes: #801

PR #811, opened 2025-11-24
Add script to check if a library is supported by the GraalVM Reachability Metadata repository

jormundur00:
## What does this PR do?

In this PR we introduce a script for easy checking if a library is supported by the GraalVM Reachability Metadata repository. We do this by parsing the `index.json` file for the particular library (if it exists) and looking for the requested version under `tested-versions`. If it exists, the library is supported.

To use the script, position yourself in the root of the repository and call: `./check-library-support.sh <groupId:artifactId:version>` (e.g.: `org.hibernate.orm:hibernate-core:7.0.0.Final`).

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/808.

P.S: We should wait for the merge of https://github.com/oracle/graalvm-reachability-metadata/pull/802 before merging this PR, as we should add the usage of the script to the `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,15 @@ To get out-of-the-box support, use the [GraalVM Gradle Plugin](https://graalvm.g
 
 ### 🔎 Check if Your Library or Framework Is Supported
 
-To see whether your library or framework is supported, visit [this page](https://www.graalvm.org/native-image/libraries-and-frameworks/). It lists libraries and frameworks that are tested and ready for GraalVM Native Image.  
+To quickly check whether reachability metadata exists for a specific library, you can run the following command directly from your terminal (works on **Linux** and **macOS**, or on **Windows** with Git Bash / WSL):
+```bash
+curl -sSL https://raw.githubusercontent.com/oracle/graalvm-reachability-metadata/master/check-library-support.sh | bash -s "<groupId>:<artifactId>:<version>"
+```
+
+For a broader overview of supported libraries and frameworks, you can visit [this page](https://www.graalvm.org/native-image/libraries-and-frameworks/). It lists libraries and frameworks that are tested and ready for GraalVM Native Image.  
 
 If you’d like yours to appear there as well, open a pull request updating [this JSON file](https://github.com/oracle/graalvm-reachability-metadata/blob/master/library-and-framework-list.json).  
 Before submitting a pull request, please read [this guide](docs/CONTRIBUTING.md#tested-libraries-and-frameworks).
-
-You can also check if metadata for a library exists on the repository directly from your terminal without cloning it using:
-```
-curl -sSL https://raw.githubusercontent.com/oracle/graalvm-reachability-metadata/master/check-library-support.sh | bash -s "<groupId>:<artifactId>:<version>"
-```
 
 ### 📚 Request Support for a New Library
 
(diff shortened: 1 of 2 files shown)


check-library-support.sh
```diff
@@ -0,0 +1,26 @@
...
+
+INDEX_FILE="metadata/$GROUP/$ARTIFACT/index.json"
+
+if [ ! -f "$INDEX_FILE" ]; then
+    echo "Library $GAV is NOT supported by the GraalVM Reachability Metadata repository."
+    exit 1
+fi
+
+# Check if the version exists in any tested-versions (exact match)
+MATCH=$(jq --arg ver "$VERSION" 'any(.[]["tested-versions"][]?; . == $ver)' "$INDEX_FILE")
```
vjovanov: I would go without `jq`. We need to add a section in the README.md that allows a copy-paste script similar to install scripts such as `curl -s "https://get.sdkman.io" | bash`.

jvukicev: I replaced the usage of `jq` with `awk` and migrated the script to be usable with `curl`, so it can now be used without cloning the repository. The `copy-paste` command can now be found in the `README.md` too.

README.md
```diff
@@ -13,6 +13,11 @@ To see whether your library or framework is supported, visit [this page](https:/
 If you’d like yours to appear there as well, open a pull request updating [this JSON file](https://github.com/oracle/graalvm-reachability-metadata/blob/master/library-and-framework-list.json).  
 Before submitting a pull request, please read [this guide](docs/CONTRIBUTING.md#tested-libraries-and-frameworks).
 
+You can also check if metadata for a library exists on the repository directly from your terminal without cloning it using:
```
vjovanov: Put this first. Do we have tickets for running with preserve and detecting libraries that don't require metadata?

check-library-support.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+# Copyright and related rights waived via CC0
+#
+# You should have received a copy of the CC0 legalcode along with this
+# work. If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
+
+set -euo pipefail
+
+if [ "$#" -ne 1 ]; then
```
vjovanov, line 10: On which OS-es does this script work?

vjovanov: Maybe mention that in the readme.

check-library-support.sh
```diff
@@ -0,0 +1,40 @@
...
+    echo "Usage: $0 <groupId>:<artifactId>:<version>"
```
vjovanov, line 11: If the library is not parsing properly we need to display nice errors.

README.md
jvukicev: Done, the command is referenced first in the README now. I've also opened the [ticket](https://github.com/oracle/graalvm-reachability-metadata/issues/844).

check-library-support.sh
jvukicev: The script should have native support on Linux and macOS, while Windows can run it through channels such as `WSL` or `Windows Git Bash`. I've added a line in the README explaining this.

check-library-support.sh
jvukicev: I've added an additional check if the argument provided to the command is in the GAV format (displaying an error if wrong format is used).

Issue #807, opened 2025-11-21
Feature: Automatically close RC (and friends) tickets for a library when the new version exists


vjovanov:
### Is your feature request related to a problem?

At the moment we leave those tickets for manual closing and they require manual work.

### Describe the solution you'd like

When we check for the existing tickets, we should also get a list of RC versions that were removed. For those, we should search open tickets on Github and close them all.

PR #815, opened 2025-11-25
Add verify-new-library-version-compatibility.yml step for closing pre-release issues when full release is present

jormundur00:
## What does this PR do?

This PR introduces automatic closing of open GitHub pre-release version issues generated by automation, when the corresponding full release version becomes available.

The workflow step closes issues in two situations:

1. If a test fails for the full version (and we open an issue for that version), we close the pre-release issue for that version.
2. If a test passes for a full release version and a pre-release issue exists for that version, we close the pre-release issue.

The changes to the workflow have been tested on a fork. Examples of both situations can be found here:

1. Situation 1: [pre-release issue closed when full version fails](https://github.com/jormundur00/graalvm-reachability-metadata/actions/runs/19671968469/job/56343005646)
2. Situation 2: [pre-release issue closed when full version passes](https://github.com/jormundur00/graalvm-reachability-metadata/actions/runs/19667578256/job/56328320251)

The new workflow step uses the same regex logic when looking for pre-release versions as it does in the `tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/TestedVersionUpdaterTask.java`, just converted to a `bash` compatible regex.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/807

### DIFF
--- a/.github/workflows/verify-new-library-version-compatibility.yml
+++ b/.github/workflows/verify-new-library-version-compatibility.yml
@@ -281,6 +281,7 @@ jobs:
         if: steps.runtests.outputs.successful_versions != '' || steps.runtests.outputs.failed_version != ''
         env:
           REPO: ${{ github.repository }}
+          PRE_RELEASE_SUFFIX: "(alpha[0-9]*|beta[0-9]*|rc[0-9]*|cr[0-9]*|m[0-9]+|ea[0-9]*|b[0-9]+|[0-9]+|preview)"
         run: |
           set -euo pipefail
           git config --local user.email "actions@github.com"
@@ -299,7 +300,7 @@ jobs:
             if [[ -z "$VERSION" ]]; then continue; fi
           
             # Parse version
-            VERSION_REGEX="^([0-9]+(\.[0-9]+)*)(\.Final)?([-.](alpha[0-9]*|beta[0-9]*|rc[0-9]*|cr[0-9]*|m[0-9]+|ea[0-9]*|b[0-9]+|[0-9]+|preview)([-.].*)?)?$"
+            VERSION_REGEX="^([0-9]+(\.[0-9]+)*)(\.Final)?([-.]$PRE_RELEASE_SUFFIX([-.].*)?)?$"
             if [[ "$VERSION" =~ $VERSION_REGEX ]]; then
               BASE_VERSION="${BASH_REMATCH[1]}"
               PRE_RELEASE_TAG="${BASH_REMATCH[4]-}"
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -277,6 +277,74 @@ jobs:
...
+          fi
+          if [[ -n "${{ steps.runtests.outputs.failed_version }}" ]]; then
+            echo "${{ steps.runtests.outputs.failed_version }}" >> "$VERSIONS_FILE"
+          fi
+          
+          while read -r VERSION; do
+            if [[ -z "$VERSION" ]]; then continue; fi
+          
+            # Parse version
+            VERSION_REGEX="^([0-9]+(\.[0-9]+)*)(\.Final)?([-.](alpha[0-9]*|beta[0-9]*|rc[0-9]*|cr[0-9]*|m[0-9]+|ea[0-9]*|b[0-9]+|[0-9]+|preview)([-.].*)?)?$"
```
vjovanov: Can we reuse this regex?

.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -277,6 +277,74 @@ jobs:
...
+              continue
+            fi
+          
+            # Only full releases trigger closing pre-release issues
+            if [[ -n "$PRE_RELEASE_TAG" ]]; then
+              echo "$VERSION is a pre-release, skipping issue closing."
+              continue
+            fi
+          
+            PRE_RELEASE_SUFFIX="(alpha[0-9]*|beta[0-9]*|rc[0-9]*|cr[0-9]*|m[0-9]+|ea[0-9]*|b[0-9]+|[0-9]+|preview)"
```
vjovanov: These regexes should be in one place.

.github/workflows/verify-new-library-version-compatibility.yml
jvukicev: I moved the `PRE_RELEASE_SUFFIX` to be an env var, and then we reuse it when creating `VERSION_REGEX`. 

If you meant reuse it from the gradle tasks where we use it, I don't believe we can, as we don't have a way to extract it to bash (and bash uses slightly different regex syntax).

.github/workflows/verify-new-library-version-compatibility.yml
jvukicev: `PRE_RELEASE_SUFFIX` is now moved to the env vars. `PRE_RELEASE_REGEX` is constructed only after `VERSION_REGEX` successfully matches, because we need the parsed base version before we can build the `PRE_RELEASE_REGEX` used to match issues.

PR #828, opened 2025-11-29
Complete infrastructure testing

vjovanov:


tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck-harness.gradle
```diff
@@ -63,83 +63,66 @@ if (CoordinateUtils.isFractionalBatch(coordinateFilter)) {
...
 }
 
-for (String coordinates in matchingCoordinates) {
-    registerPerCoordinateTask(coordinates, checkstyle, "checkstyle", CheckstyleInvocationTask, true)
-    registerPerCoordinateTask(coordinates, "clean", "clean", CleanInvocationTask, true)
-    registerPerCoordinateTask(coordinates, compileTestJava, "compileTestJava", CompileTestJavaInvocationTask, true)
-    registerPerCoordinateTask(coordinates, javaTest, "javaTest", JavaTestInvocationTask, true)
-    registerPerCoordinateTask(coordinates, nativeTestCompile, "nativeTestCompile", NativeTestCompileInvocationTask, true)
-    // No aggregate wiring for per-coordinate docker pulls; they are invoked by dedicated workflows.
-    registerPerCoordinateTask(coordinates, null, "pullAllowedDockerImages", DockerTask.class, false) { t ->
```
jvukicev, line 136: With the addition of `ComputeAndPullAllowedDockerImagesTask`, I don't think `DockerTask` is used anywhere (you removed it here, but it's still an unused import). I think we can consider deleting it from the repository, but we can do that in a follow-up PR to avoid rerunning all tests for this change.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/SingleCoordinateTask.groovy
```diff
@@ -0,0 +1,31 @@
...
+ */
+abstract class SingleCoordinateTask extends CoordinatesAwareTask {
+
+    /**
+     * Resolves to exactly one coordinate or fails with a helpful error.
+     */
+    protected String resolveSingleCoordinate() {
+        List<String> coords = resolveCoordinates()
+        if (coords.isEmpty()) {
+            throw new GradleException("No matching coordinates found. Provide a concrete coordinate via -Pcoordinates=group:artifact:version")
```
jvukicev, line 23: Small nit: in the root `build.gradle`, when throwing the `GradleException`, the message states to use `-Pcoordinate` (or `-Pcoordinates` in brackets), but here we ask for `-Pcoordinates` explicitly. If we intend to have `-Pcoordinate` as a valid property, maybe we should add and use it here as well (as a fallback at least). We can also do this in a follow-up, or just ignore it.

jormundur00 (approved):
LGTM! Didn't test every possible combination of commands locally, but what I did worked (and I see that all the `changed-build-logic` tests passed in the CI, so we should be good).

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/SingleCoordinateTask.groovy
vjovanov: I will fix this up later, as this PR runs for a long long time.

Issue #823, opened 2025-11-27
Bug: Inconsistent issue matching in verify-new-library-version-compatibility.yml


jormundur00:
### Describe the bug

The verify-new-library-version-compatibility.yml workflow fails due to the "safety assert" issue matching triggering ("should not reach" part of the workflow is reachable). If a issue for a library fail exists, we should match it early, before running tests, and skip running tests.

### Steps to reproduce

```markdown
The issue happens in the master branch workflow runs:
https://github.com/oracle/graalvm-reachability-metadata/actions/runs/19721107395/job/56503740223.

Just rerun the workflow in the GitHub actions to see it happen.
```

### Expected behavior

We should match the issue early, so we don't perform the tests.

### Logs and error messages

```text
Searching for an issue that ends with:  fails for org.hibernate.orm:hibernate-core:7.2.0.CR1
Must not reach here as the issue already exists: 749
The previous stage should of failed before.
Error: Process completed with exit code 1.
```

### GraalVM version

GraalVM 25.0.2

### Operating system

Linux

### CPU architecture

AMD64 (x86-64)

PR #829, opened 2025-12-01
Skip GitHub issue creation if one already exists instead of failing verify-new-version-compatibility.yml workflow

jormundur00:
## What does this PR do?

This PR updates the workflow in `verify-new-library-version-compatibility.yml` so that the creation of a GitHub issue for a failing library version is **skipped if an issue already exists**, instead of failing the workflow.

Previously, when the workflow detected an existing issue for a failing version, it would trigger a “should-not-reach-here” failure, causing the job to fail. With this change:

- If an issue for the failing version already exists, no new issue is created.
- The workflow continues without failing.
- New issues are only created for versions that do not yet have an existing issue.

This prevents duplicate issues while keeping the workflow stable.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/823

### DIFF
--- a/.github/workflows/verify-new-library-version-compatibility.yml
+++ b/.github/workflows/verify-new-library-version-compatibility.yml
@@ -92,22 +92,24 @@ jobs:
           GROUP_ID="$(echo "${{ matrix.item.name }}" | cut -d: -f1)"
           ARTIFACT_ID="$(echo "${{ matrix.item.name }}" | cut -d: -f2)"
           
-          GAV_COORDINATES_VERSIONLESS="$GROUP_ID:$ARTIFACT_ID"
-          TITLE_END_VERSIONLESS="${{ env.ISSUE_TITLE_MIDDLE }}$GAV_COORDINATES_VERSIONLESS"
+          readarray -t VERSIONS < <(echo '${{ toJson(matrix.item.versions) }}' | jq -r '.[]')
+          FIRST_TESTING_VERSION="${VERSIONS[0]}"
+          GAV_COORDINATES="$GROUP_ID:$ARTIFACT_ID:$FIRST_TESTING_VERSION"
+          TITLE_END="${{ env.ISSUE_TITLE_MIDDLE }}$GAV_COORDINATES"
           
           ISSUE_NUMBER=$(
             gh issue list \
               --repo "${{ github.repository }}" \
               --state open \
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -92,24 +92,22 @@ jobs:
           GROUP_ID="$(echo "${{ matrix.item.name }}" | cut -d: -f1)"
           ARTIFACT_ID="$(echo "${{ matrix.item.name }}" | cut -d: -f2)"
           
-          readarray -t VERSIONS < <(echo '${{ toJson(matrix.item.versions) }}' | jq -r '.[]')
-          FIRST_TESTING_VERSION="${VERSIONS[0]}"
-          GAV_COORDINATES="$GROUP_ID:$ARTIFACT_ID:$FIRST_TESTING_VERSION"
-          TITLE_END="${{ env.ISSUE_TITLE_MIDDLE }}$GAV_COORDINATES"
+          GAV_COORDINATES_VERSIONLESS="$GROUP_ID:$ARTIFACT_ID"
```
vjovanov: We should comment why this is done in the following way.

.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -92,24 +92,22 @@ jobs:
...
+          TITLE_END_VERSIONLESS="${{ env.ISSUE_TITLE_MIDDLE }}$GAV_COORDINATES_VERSIONLESS"
```
vjovanov: This is not a good solution, because if the previous versions pass we should still test them. We model this wrong: the issues are not one big array, but multiple arrays under each version.

The simple solution is to simply update the ticket with the new failure information, or even ignore it completely, that is, simply print that this ticket was already handled and ignore.

The more complicated solution is to run model the repo differently, but this is not for this ticket.

jvukicev: This approach seems cleaner. I’ve reverted the “versionless matching” logic and instead replaced the “should-not-reach-here” failure with simply skipping issue creation if an issue already exists.
I’ve kept the initial check that looks for an issue for the first version after the latest tested version, since in most cases this is the version with an existing issue, allowing us to avoid unnecessary computation.

PR #834, opened 2025-12-02
[Automation] Generated metadata for org.jooq:jooq:3.20.0

kimeta:
## What does this PR do?

Fixes: #752

This PR provides new metadata needed for the org.jooq:jooq:3.20.0, addressing Native Image run failures caused by changes in the updated library version.

### DIFF
--- a/metadata/org.jooq/jooq/index.json
+++ b/metadata/org.jooq/jooq/index.json
@@ -2,6 +2,13 @@
   {
     "latest" : true,
     "module" : "org.jooq:jooq",
+    "metadata-version" : "3.20.0",
+    "tested-versions" : [
+      "3.20.0"
+    ]
+  },
+  {
+    "module" : "org.jooq:jooq",
     "metadata-version" : "3.18.2",
     "tested-versions" : [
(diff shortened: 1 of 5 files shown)


metadata/org.jooq/jooq/index.json
```diff
@@ -1,6 +1,14 @@
 [
   {
     "latest" : true,
+    "module" : "org.jooq:jooq",
+    "metadata-version" : "3.20.0",
+    "tested-versions" : [
+      "3.20.0"
+    ],
+    "skipped-versions" : [ ]
```
jvukicev: Small nit (unrelated to the correctness of the metadata): `skipped-versions` is an optional field (like `default-for`, etc.), so we don't really need to generate it if we don't intend to put data in it?

vjovanov (approved):

metadata/org.jooq/jooq/index.json
kimeta: Yes, thank you. I will fix the script in the follow up PR.

Issue #673, opened 2025-09-09
Run all tests from `tested-versions` when metadata for certain library changes


dnestoro:
Currently, we only test all `tested-versions` for a certain library once—when the scheduled job discovers new versions. After the list is updated, we only test the version defined in `metadata-version` field. ([Example of the index file](https://github.com/oracle/graalvm-reachability-metadata/blob/master/metadata/org.hibernate.orm/hibernate-core/index.json#L1)) 

Idea: When metadata or tests for a certain library change, we should run tests with all `tested-versions` as well (not just the version the tests are currently assigned to) to see if the change broke support for some `tested-versions`. 

For discussion: Once contributed to the repository, the metadata can only be extended. Extending metadata should not break support for already tested versions, since those library versions are already released and never change. The metadata that was once appropriate for that version should always remain valid for that version. We should therefore consider testing with all `tested-versions` only if the tests change.

PR #835, opened 2025-12-02
Make test-changed-metadata test the new metadata with all previously tested versions

jormundur00:
## What does this PR do?

This PR updates the metadata testing workflow so that **any newly added or modified metadata entry is tested against *all previously validated versions* of that library**.

Specifically, the workflow now:

- Reads the list of `tested-versions` from `index.json` for the affected metadata-version.
- Runs the full metadata test suite against **every** version in that list.
- Fails immediately if any version fails, and reports which version broke.
- Ensures that new or updated metadata remains backward compatible with all earlier versions it claimed to support.

This prevents regressions where updated metadata might break support for older library versions.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/673


.github/workflows/test-changed-infrastructure.yml
```diff
@@ -90,6 +90,7 @@ jobs:
     runs-on: "ubuntu-22.04"
     timeout-minutes: 1
     needs: test-changed-infrastructure
+    if: always()
```
jvukicev, line 101: This is a bit unrelated to the whole PR, but I've added it here since we also made this change in `test-changed-metadata.yml`. Without the `if: always()`, the "Some tests failed" step was unreachable, as if `test-changed-infrastrucutre` (or `test-changed-metadata`) failed, then the `all-infrastructure-passed` (`all-tests-passed`) job would always be skipped.

Issue #870, opened 2025-12-09
Bug: Netty tests should use random port assignment


kimeta:
### Describe the bug

When running our automation actions, updates to the io.netty library can fail because some Netty tests use hard-coded ports. This can result in `java.net.BindException: Address already in use` errors, as multiple test instances may attempt to bind to the same port.

### Steps to reproduce

This issue occurs during CI runs when multiple tests are executed in parallel. If tests attempt to use the same hard-coded port, or if the server socket is not given enough time to shut down, subsequent tests can fail to bind to the port.

### Expected behavior

Tests in CI should not fail due to unpredictable issues caused by port conflicts. Each test should use a randomly assigned, available port to ensure reliable and deterministic test execution.

### Logs and error messages

```text
> Task :generateTestResourcesConfigFile
[native-image-plugin] Resources configuration written into /home/runner/work/graalvm-reachability-metadata/graalvm-reachability-metadata/tests/src/io.netty/netty-transport/4.1.115.Final/build/native/generated/generateTestResourcesConfigFile/resource-config.json

> Task :nativeTestCompile
[native-image-plugin] GraalVM Toolchain detection is disabled
[native-image-plugin] GraalVM location read from environment variable: GRAALVM_HOME
[native-image-plugin] Native Image executable path: /opt/hostedtoolcache/graalvm-jdk-25_linux-x64_bin/25.0.0/x64/graalvm-jdk-25.0.1+8.1/lib/svm/bin/native-image
Picked up NATIVE_IMAGE_OPTIONS: -H:BuildOutputJSONFile=/tmp/native-image-build-output.json
Warning: Using a deprecated option --allow-incomplete-classpath from command line. Allowing an incomplete classpath is now the default. Use --link-at-build-time to report linking errors at image build time for a class or package.
Warning: The option '-H:BuildOutputJSONFile=/tmp/native-image-build-output.json' is experimental and must be enabled via '-H:+UnlockExperimentalVMOptions' in the future.
Warning: Please re-evaluate whether any experimental option is required, and either remove or unlock it. The build output lists all active experimental options, including where they come from and possible alternatives. If you think an experimental option should be considered as stable, please file an issue.
========================================================================================================================
GraalVM Native Image: Generating 'netty-tests-tests' (executable)...
========================================================================================================================
For detailed information and explanations on the build output, visit:
https://github.com/oracle/graal/blob/master/docs/reference-manual/native-image/BuildOutput.md
[...]

PR #871, opened 2025-12-09
Assign random ports in Netty tests to prevent BindException 

kimeta:
## What does this PR do?

Fixes: #870 

This PR resolves an issue where Netty tests were using hard-coded ports, causing `java.net.BindException: Address already in use` errors during our automated CI workflows. 

To address this, the affected Netty tests have been refactored to use randomly assigned available ports at runtime, preventing port conflicts and improving the reliability of test execution in the CI. This change should be applied to all test versions to ensure future reliability and enable safe automation when introducing new library versions.


vjovanov:
@jormundur00 This should run all the netty versions, but it ended up running most of the repo. Any idea why?

jormundur00 (approved):

Issue #855, opened 2025-12-08
Bug: test-changed-metadata.yml uses -O2 instead of -Ob, as specified by ci.json


jormundur00:
### Describe the bug

The workflow specified by `test-changed-metadata.yml` doesn't use build arguments specified by `ci.json`, when it should.

### Steps to reproduce

```markdown
The issue can be seen on workflow runs of `test-changed-metadata.yml`, for example:
https://github.com/oracle/graalvm-reachability-metadata/actions/runs/20024175588/job/57417773264?pr=853
```

### Expected behavior

The workflow should use the arguments specified by `ci.json`.

### Logs and error messages

```text
https://github.com/oracle/graalvm-reachability-metadata/actions/runs/20024175588/job/57417773264?pr=853
```

### GraalVM version

GraalVM 25.0.2

### Operating system

Linux

### CPU architecture

AMD64 (x86-64)

PR #875, opened 2025-12-10
Fix tck not properly parsing ci.json

jormundur00:
## What does this PR do?

This PR resolves an issue where internal TCK builds failed to retrieve the necessary native-image build arguments (such as optimization flags like `-Ob`) from the central `ci.json` configuration file, causing tests to run with incorrect or default configurations.

Also in this PR, we add a hard-fail when `ci.json` is missing.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/855


jormundur00:
> We need to test all metadata manually.

Due to the change in `ci.json` (because `--verbose` should be used, not `-verbose`), the `test-all-metadata` workflow ran automatically, and successfully completed.

Issue #850, opened 2025-12-06
Update existing library: org.liquibase:liquibase-core:4.32.0


Haarolean:
### Full Maven coordinates

org.liquibase:liquibase-core:4.32.0

### Additional context and logs

Hello, 

encountered #711 with spring-boot 3.5.6 and liquibase 4.32.0. After bumping metadata repo version to the latest (`0.3.32`), I started to encounter the following:

```
Caused by: liquibase.exception.UnexpectedLiquibaseException: java.lang.NoSuchMethodException: liquibase.configuration.ConfiguredValueModifierFactory.<init>()
	at liquibase.Scope.getSingleton(Scope.java:373)
	at liquibase.configuration.LiquibaseConfiguration.getCurrentConfiguredValue(LiquibaseConfiguration.java:151)
	at liquibase.configuration.ConfigurationDefinition.getCurrentConfiguredValue(ConfigurationDefinition.java:106)
	at liquibase.configuration.ConfigurationDefinition.getCurrentConfiguredValue(ConfigurationDefinition.java:90)
	at liquibase.integration.spring.SpringLiquibase.afterPropertiesSet(SpringLiquibase.java:246)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.invokeInitMethods(AbstractAutowireCapableBeanFactory.java:1873)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1822)
	... 18 more
Caused by: java.lang.NoSuchMethodException: liquibase.configuration.ConfiguredValueModifierFactory.<init>()
	at java.base@24.0.2/java.lang.Class.checkConstructor(DynamicHub.java:1189)
	at java.base@24.0.2/java.lang.Class.getConstructor0(DynamicHub.java:1386)
	at java.base@24.0.2/java.lang.Class.getDeclaredConstructor(DynamicHub.java:2493)
	at liquibase.Scope.getSingleton(Scope.java:368)
	... 24 more
```

### GraalVM version

oracle 24.0.2

### Operating system

macOS

### CPU architecture

AARCH64 (ARM64)

Haarolean:
@kimeta thank you. Any chance this could be released?

PR #876, opened 2025-12-10
Update metadata for liquibase-core 4.25.0-4.33.0

kimeta:
## What does this PR do?
Fixes: #850

This PR addresses metadata issues introduced with merging #666, which resulted in missing metadata for liquibase.core versions 4.25.0-4.33.0. These metadata gaps caused failures in several Spring smoke tests. 

To fix these issues, this PR adds the necessary metadata for the affected liquibase.core versions.


tests/src/index.json
```diff
@@ -1247,7 +1247,12 @@
           "4.26.0",
           "4.27.0",
           "4.28.0",
-          "4.29.0"
+          "4.29.0",
```
vjovanov: These are not tested, why?

vjovanov: Did you rebase?

kimeta: Yes. These are not tested because this PR was merged after metadata for version 5.0.1 is merged. That all happened before our automation testing workflow. But this list should be reduced only to versions that have metadata directories.

vjovanov: Cool, all of these are then tested, they are just hidden in the current task as a batch test.

Issue #774, opened 2025-11-14
Feature: Use Java instead of Groovy in all the tasks in the repo


vjovanov:
### Is your feature request related to a problem?

We would like to have a language that is more approachable by the general public. This is a nice to have though, and low priority.

### Describe the solution you'd like

Simply rewrite all scripts from Groovy to Kotlin.

PR #882, opened 2025-12-11
Rewrite Groovy into Java

vjovanov:
Fixes: #774


tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -100,16 +100,38 @@ public Path getTestDir(String coordinates) {
...
+                );
+                if (found) {
+                    return testRoot().resolve((String) entry.get("test-project-path"));
+                }
+            }
+        } catch (Exception ignored) {
+            // Fall through to conventional layout resolution below.
+        }
+        // Fallback: conventional layout tests/src/<group>/<artifact>/<version>
+        Path conventional = testRoot().resolve(groupId).resolve(artifactId).resolve(version);
```
jvukicev, line 144: So previously we only matched using `tests/src/index.json`, and failed otherwise? These fallbacks will be useful once we get rid of that `index.json`.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/FetchExistingLibrariesWithNewerVersionsTask.java
```diff
@@ -0,0 +1,249 @@
...
+
+    private static final List<String> INFRASTRUCTURE_TESTS = List.of("samples", "org.example");
+
+    /**
+     * Identifies library versions, including optional pre-release and ".Final" suffixes.
+     *
+     * Pre-release identifiers (case-insensitive): alpha, beta, rc, cr, m<num>, ea, b<num>, preview, and pure numeric suffixes.
+     * Versions ending with ".Final" are treated as full releases of the base version.
+     */
+    private static final Pattern VERSION_PATTERN = Pattern.compile("(?i)^(\\\\d+(?:\\\\.\\\\d+)*)"
```
jvukicev: This pattern is not correct. Please take a look at `org.graalvm.internal.tck.TestedVersionUpdaterTask#VERSION_PATTERN` for the Java equivalent of this Groovy pattern, and either copy it, or even better, reuse it (reference the other from whichever class is best).

vjovanov: Fixed.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
vjovanov: We can add this when we need it. For now I have deleted the fallback.

Issue #813, opened 2025-11-24
Feature: Introduce a version for the metadata repo structure


vjovanov:
### Is your feature request related to a problem?

The problem is that when we update the repo structure, the previous build-tools will not work.

### Describe the solution you'd like

The metadata repo should always have a version for its internal structure. It would be best to have JSON schmeas for all JSON files and if they are modified we have bump the version.

PR #883, opened 2025-12-11
Add versioned schemas and PR validation for index.json files

jormundur00:
## What does this PR do?

This PR introduces **versioned schemas** for all existing `index.json` files in the repository and adds a workflow to **validate any changed `index.json` files** in a pull request against their respective schemas.  

Currently, there are **three types of `index.json` files** governed by these schemas:

1. **Global metadata `index.json`** - Located at `metadata/index.json`
2. **Library metadata `index.json`** - Located at `metadata/<groupId>/<artifactId>/index.json` for each library in the repository
3. **Optional test sub-project `index.json`** - Located at `tests/src/<groupId>/<artifactId>/<version>/index.json`  

---

### New Commands & Tasks

A new Java-based Gradle task `validateIndexFiles` has been added to the harness. This task is **coordinate-aware** (supporting `-Pcoordinates` and `--coordinates`) and handles space-separated lists for batch validation.

**Validate all index files locally**:
```console
./gradlew validateIndexFiles -Pcoordinates=all
```

**Validate specific library and its test indexes**:
```console
./gradlew validateIndexFiles -Pcoordinates=org.postgresql:postgresql:42.7.3
```
Infrastructure Integration: The validateIndexFiles task is now a core part of the testInfra and testAllInfra verification suites. It runs in parallel with other checks (like checkstyle and spotless) to ensure data integrity without increasing CI runtime significantly.

### Additional Changes

* **CI Workflow Update**: The GitHub Action now utilizes a space-separated string of changed coordinates. This allows for highly targeted validation on Pull Requests, significantly reducing CI overhead.
* **Documentation Improvements**: Updated `CONTRIBUTING.md` to include specific validation steps within the local testing recipes and a new entry in the **Quick Reference** section for faster developer onboarding.
* **Release Packaging**: To ensure compatibility and detect version mismatches in **Native Build Tools**, the following schemas are now bundled with all future releases of GraalVM Reachability Metadata:
    * `schemas/metadata-root-index-schema-v1.0.0.json`
    * `schemas/metadata-library-index-schema-v1.0.0.json`
    * `schemas/library-and-framework-list-schema-v1.0.0.json`

---
Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/813

### DIFF
--- a/docs/DEVELOPING.md
+++ b/docs/DEVELOPING.md
@@ -51,7 +51,7 @@ For a single coordinate, CI runs three steps in this order:
     ```console
    ./gradlew pullAllowedDockerImages -Pcoordinates=org.postgresql:postgresql:42.7.3
     ```
-2. Validate index schemas:
+2. Validate index.json schemas:
     ```console
    ./gradlew validateIndexFiles -Pcoordinates=org.postgresql:postgresql:42.7.3
     ```


.github/workflows/index-file-validation.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: |
+          pip install jq check-jsonschema
+
+      - name: "Check that the changed index.json files conform to their schemas"
+        if: steps.filter.outputs.changed == 'true'
+        run: |
+          set -e
+          
+          # Treat changed_files as a proper array
+          FILES=(${{ steps.filter.outputs.changed_files }})
```
vjovanov: We should always strive to do this in gradle. That way I can also do it locally to verify everything is OK.

jvukicev: Makes sense. I've removed this file-producing part of the GitHub action and created a `generateChangedIndexFileMatrix` gradle task that will do this in gradle (and is locally test-able). As these tasks require a `ci.json` entry, I've added it as only tested on `java: latest-ea` and `os: ubuntu-latest` as not to produce multiple jobs (because there shouldn't be any change with different jdks for this functionality as we don't run native-image).

.github/workflows/index-file-validation.yml
```diff
@@ -0,0 +1,114 @@
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'graalvm'
+          java-version: '21'
+
+      - name: "🕸️ Populate matrix"
+        id: set-matrix
+        if: steps.filter.outputs.changed == 'true'
+        run: |
+          ./gradlew generateChangedIndexFileMatrix -PbaseCommit=${{ github.event.pull_request.base.sha }} -PnewCommit=${{ github.event.pull_request.head.sha }}
```
vjovanov: Why do we need a matrix for this? It takes 1 s to do, so we can do it all at once.

schemas/metadata-library-index-schema-v1.0.0.json
```diff
@@ -0,0 +1,100 @@
...
+          "minLength": 1
+        }
+      },
+      "latest": {
+        "type": "boolean",
+        "description": "Marks this entry as the latest/default metadata for currently supported versions."
+      },
+      "default-for": {
+        "type": "string",
+        "description": "Java regular expression describing the version range for which this entry should be used by default (e.g. '7\\\\.1\\\\..*')."
```
vjovanov, line 46: And why do we need these escapes?

schemas/metadata-library-index-schema-v1.0.0.json
```diff
@@ -0,0 +1,100 @@
...
+      "pattern": "^[^:]+:[^:]+$"
+    }
+  },
+  "examples": [
+    [
+      {
+        "metadata-version": "0.0.1",
+        "module": "org.example:library",
+        "tested-versions": [ "0.0.1", "0.0.2" ],
+        "default-for": "0\\.0\\..*"
```
vjovanov: Also here, why do we need these escapes here?

schemas/metadata-library-index-schema-v1.0.0.json
```diff
@@ -0,0 +1,100 @@
...
+      },
+      {
+        "latest": true,
+        "metadata-version": "1.0.0",
+        "module": "org.example:library",
+        "tested-versions": [ "1.0.0", "1.1.0-M1", "1.1.0" ]
```
vjovanov: I would not put the milestone and CRs here because we don't want to encourage that.

schemas/metadata-library-index-schema-v1.0.0.json
```diff
@@ -0,0 +1,100 @@
...
+      },
+      {
+        "metadata-version": "1.19.0",
+        "module": "io.opentelemetry:opentelemetry-exporter-logging",
+        "tested-versions": [ "1.19.0" ],
+        "override": true,
+        "skipped-versions": [
+          { "version": "1.0.5", "reason": "Known incompatible API change." }
```
vjovanov: Usually this is the case because native image does not build due to some parsing issues.

.github/workflows/index-file-validation.yml
jvukicev: Yeah, this was a bit misleading, as we generated a one-dimensional matrix (so an array) which we fully parsed in one job. I renamed this to `generateChangedIndexFileList` (and made it so we generate a list, no longer needing `ci.json` input either), and we still validate all changed `index.json` files in one job run (as we don't need parallelism here).

schemas/metadata-library-index-schema-v1.0.0.json
jvukicev: In the description, we need four backslashes to 'show' two to the user (and the need for two is explained in the other comment).

The IDE documentation viewer parses this JSON string and 'collapses' the four backslashes into two. This ensures the user sees `7\\.1\\..*` in their tooltip, which is exactly what they need to copy-paste into their own `index.json` to be valid.

schemas/metadata-library-index-schema-v1.0.0.json
jvukicev: Because this is a Regex inside a JSON string.

In Regex, the dot is a wildcard, so we need `\.` to match a literal version dot. But in JSON, a backslash is a special character that must be escaped by another backslash.

So, `\\.` in the file tells the JSON parser to pass `\.` to the Regex engine, which then correctly matches the literal version number.

schemas/metadata-library-index-schema-v1.0.0.json
jvukicev: Makes sense, replaced the milestone version with another full release version.

schemas/metadata-library-index-schema-v1.0.0.json
jvukicev: I've replaced this example with a "real world" example where tests fail due to an un-parsable `reflect-config.json`.

docs/DEVELOPING.md
```diff
@@ -51,6 +51,10 @@ For a single coordinate, CI runs three steps in this order:
     ```console
    ./gradlew pullAllowedDockerImages -Pcoordinates=org.postgresql:postgresql:42.7.3
     ```
+2. Validate index schemas:
```
vjovanov: `index.json`

Issue #848, opened 2025-12-05
Feature: Remove global tests index.json from the repository


jormundur00:
### Is your feature request related to a problem?

As a part of the GraalVM Reachability Metadata repository 1.0 release, we should remove the index file located in the `tests/src` directory.

### Describe the solution you'd like

We need to change the logic of the tck so we can infer the information given by the said index file, either through paths, or through the remaining metadata `index.json`.

PR #894, opened 2025-12-16
Remove global tests index.json file

jormundur00:
## What does this PR do?

This PR streamlines the metadata management process by **removing the redundant global test index** (`tests/src/index.json`) and **consolidating all test configuration functionality** into the per-library `metadata/<groupId>/<artifactId>/index.json` files.

#### 1. Removal of Redundant Global Test Index

* **File Removed:** The global test index (`tests/src/index.json`) is deleted.
* **Rationale:** This eliminates a source of redundant information and manual synchronization errors, simplifying the workflow for adding and changing metadata.

#### 2. Consolidation and Introduction of the `test-version` Field

All functionality from the removed global index is moved to the per-library metadata index, facilitated by a new field.

| Feature | New Location/Field | Functionality |
| :--- | :--- | :--- |
| **Test Directory Mapping** | **`test-version`** (New Optional String) | **Overrides** the default test directory derived from `metadata-version`. This allows multiple metadata configurations (e.g., across versions) to share a single, stable set of tests. |
| **Data Integrity** | **JSON Schema Update** | The `metadata-library-index-schema-v1.0.0.json` is updated to include the new **`test-version`** property, maintaining schema validation integrity for the modified files. |

#### 3. Automated Maintenance and Integrity

The process for updating metadata is made more robust through new automation:

* **Pre-Release Migration:** The `TestedVersionUpdaterTask` is updated to automatically handle `test-version` propagation. When a pre-release metadata directory is renamed to its full release counterpart (e.g., `1.0.0-M1` $\rightarrow$ `1.0.0`), any index entries that relied on the old directory via the `test-version` override are **automatically updated** to point to the new full release version.
* **Generation Default:** The `FixTestNativeImageRun` Gradle task is updated to automatically set a default value for the `test-version` field when new library metadata is generated.

---

### Prerequisites

This PR must only be merged after merging the schema update PR:

* **https://github.com/oracle/graalvm-reachability-metadata/pull/883** (This PR introduces the required **`test-version`** property into the JSON schema definition.)

---

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/848

### DIFF
--- a/check-library-support.sh
+++ b/check-library-support.sh
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-
 # Copyright and related rights waived via CC0
 #
 # You should have received a copy of the CC0 legalcode along with this


tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/ComputeAndPullAllowedDockerImagesTask.java
```diff
@@ -101,34 +101,44 @@ public void run() throws IOException {
...
+                        .resolve("required-docker-images.txt");
+
+                File f = getProject().file(dockerImagesPath);
+                if (f.exists()) {
+                    Files.readAllLines(f.toPath()).stream()
+                            .map(String::trim)
+                            .filter(s -> !s.isEmpty())
+                            .filter(s -> !s.startsWith("#"))
+                            .forEach(requiredImages::add);
+                }
```
kimeta, line 141: It should probably throw a `GradleException` if metadata version is not found.

schemas/metadata-library-index-schema-v1.0.1.json
```diff
@@ -0,0 +1,106 @@
...
+  "$id": "https://github.com/oracle/graalvm-reachability-metadata/schemas/metadata-library-index.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "description": "Schema for metadata/<groupId>/<artifactId>/index.json. Each entry describes a metadata bundle for a range of library versions.",
+  "type": "array",
+  "minItems": 1,
+  "items": {
+    "type": "object",
+    "additionalProperties": false,
+    "required": [
+      "module",
```
kimeta: Do we need `module` field in every entry of index.json? Can it differ between two entries in the same index.json?

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/ComputeAndPullAllowedDockerImagesTask.java
jvukicev: The `metadata-version` it defaults to must always be present in the `index.json`. The null check is there mostly to avoid warnings. We currently don't test this when adding new metadata, but will do this once the PR for schema validation is merged: https://github.com/oracle/graalvm-reachability-metadata/pull/883.

schemas/metadata-library-index-schema-v1.0.1.json
jvukicev: As current functionality stands, we do need it in every entry. I don't think it can differ between two entries. I'll look into completely removing this field in a follow up PR for removing the `metadata/index.json` file.

The schema introduction I made in this PR has been removed, as I want to introduce all schemas in one uniform PR.

Issue #904, opened 2025-12-19
Bug: checkAllowedDockerImages task fails due to incomplete GrypeEntry class


jormundur00:
### Describe the bug

The checkAllowedDockerImages task fails when attempting to update a docker image due to missing a new Grype field.

### Steps to reproduce

```markdown
https://github.com/oracle/graalvm-reachability-metadata/actions/runs/20268061977/job/58196259899?pr=890
```

### Expected behavior

The check shouldn't fail the GrypeTask.

### Logs and error messages

```text
> Task :checkAllowedDockerImages FAILED


[Incubating] Problems report is available at: file:///home/runner/work/graalvm-reachability-metadata/graalvm-reachability-metadata/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':checkAllowedDockerImages'.
> Cannot parse grype output for image: mariadb:10.8 .Reason: Unrecognized field "cwes" (class org.graalvm.internal.tck.model.grype.GrypeEntry), not marked as ignorable (12 known properties: "cvss", "urls", "dataSource", "knownExploited", "id", "description", "severity", "risk", "fix", "advisories", "namespace", "epss"])
   at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 15685] (through reference chain: org.graalvm.internal.tck.model.grype.GrypeEntry["cwes"])

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to generate a Build Scan (Powered by Develocity).
> Get more help at https://help.gradle.org.

BUILD FAILED in 1m 20s
```

### Reproducer (optional)

https://github.com/oracle/graalvm-reachability-metadata/actions/runs/20268061977/job/58196259899?pr=890

### GraalVM version

GraalVM 25.0.2

### Operating system

Linux

### CPU architecture

AMD64 (x86-64)

PR #905, opened 2025-12-19
Add the acceptance of cwes field to the GrypeEntry

jormundur00:
## What does this PR do?

Currently, GrypeTask fails for some docker images due to a missing `cwes` field. In this PR, we add the missing field to the `GrypeEntry` class, so it is properly ignored.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/904

### DIFF
--- a/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/model/grype/GrypeEntry.java
+++ b/tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/model/grype/GrypeEntry.java
@@ -6,12 +6,12 @@
  */
 package org.graalvm.internal.tck.model.grype;
 
-/*
- * JSON model for metadata/index.json.
- */
 import java.util.List;
 import java.util.Map;
 
+/*
+ * JSON model for metadata/index.json.
+ */
 public class GrypeEntry {
     public String id;
     public String dataSource;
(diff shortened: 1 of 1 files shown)


tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/model/grype/GrypeEntry.java
```diff
@@ -22,11 +22,19 @@ public class GrypeEntry {
...
     public List<KnownExploited> knownExploited;
     public List<Epss> epss;
+    public List<Cwe> cwes;
     public Fix fix;
     public List<Object> advisories; // unknown structure
     public Double risk;
 }
 
+class Cwe {
+    public String cve;
```
matneu, line 32: Does that map the CWE (Common Weakness Enumerator, ~= vulnerability type) to an individual CVE (actual vulnaerability) in the json?

jvukicev: Yes, exactly. An example of an Cwe object looks like the following: `CWE(cve=CVE-2022-2879, cwe=CWE-770, type=Primary)`.

We currently don't use this specific field in our analysis, but must register (and ignore) it so our parsing of the Grype output doesn't fail.

matneu (approved):

Issue #914, opened 2026-01-06
Bug: verify-new-library-version-compatibility.yml doesn't take into consideration metadata to test mapping overrides


jormundur00:
### Describe the bug

When resolving the latest version of an existing metadata entry, the workflow resolves it based on sorting the test directories for that library and picking out the latest there. This completely ignores the `latest` field of specific metadata entries and creates problems when we match multiple metadata files to the same tests.

For example:
```
[
  {
    "latest" : true,
    "module" : "org.flywaydb:flyway-sqlserver",
    "metadata-version" : "10.13.0",
    "test-version": "10.10.0",
    "tested-versions" : [
      "10.13.0"
    ]
  },
  {
    "module" : "org.flywaydb:flyway-sqlserver",
    "metadata-version" : "10.10.0",
    "tested-versions" : [
      "10.10.0",
      "10.11.0",
      "10.11.1",
      "10.12.0",
      // new version will be added here
    ]
  }
]
```
new versions will be added to the older metadata entry here, as the newer `10.13.0` metadata is matched to the `10.10.0` tests.

### Steps to reproduce

```markdown
Run the `verify-new-library-version-compatibility.yml` workflow.
```

### Expected behavior

We should parse the `index.json` file to resolve the latest metadata version and ONLY fallback to the old parsing ways if there is no `latest` tag in the index file.

### Logs and error messages

```text
No logs.
```

### GraalVM version

GraalVM 25.0.2

### Operating system

Linux

### CPU architecture

AMD64 (x86-64)

PR #917, opened 2026-01-08
Fix "latest" version retrieval in verify-new-library-version-compatibility.yml and improve index.json validation

jormundur00:
## What does this PR do?

This PR transitions the extraction of library test parameters from a GitHub Actions Bash step into a dedicated Gradle task and introduces stricter semantic validation for library indexes.

### 1. New `extractLibraryTestParams` Gradle Task
Moved the logic for resolving test paths and versions from unmaintained shell scripts into a formal Gradle task.
* **Authority-Based Resolution:** Fixed a bug where the workflow previously ignored the `"latest": true` flag and the `"test-version"` overrides in `index.json`. Previously, the system relied on a simple filesystem directory scan, which could lead to incorrect version selection.
* **Source of Truth:** The task now treats `index.json` as the primary authority. It resolves versions in the following priority:
    1. The entry explicitly marked `"latest": true`.
    2. The entry with the highest `"metadata-version"` defined in the JSON.
    3. (Fallback) The highest version directory found on the filesystem.
* **CI Integration:** Centralizes the resolution of `LATEST_VERSION`, `TEST_PATH`, and `TEST_COORDINATES`, automatically populating `GITHUB_ENV` when run in CI.

### 2. Enhanced Semantic Index Validation
Updated the `validateIndexFiles` task with a new "Forward-Referencing" integrity check.
* **The Rule:** In any `index.json`, a version in the `tested-versions` array cannot be higher than the next available `metadata-version` in the same file.
* **The Problem:** This prevents logical inconsistencies where an older metadata entry mistakenly "claims" library versions that actually belong to a newer metadata entry.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/914

### DIFF
--- a/.github/workflows/verify-new-library-version-compatibility.yml
+++ b/.github/workflows/verify-new-library-version-compatibility.yml
@@ -17,7 +17,7 @@ jobs:
   get-all-libraries:
     name: "📋 Get list of all supported libraries with newer versions"
     # Opens PRs, with labels and assignees. Works only on the main repo.
-    if: github.repository == 'jormundur00/graalvm-reachability-metadata'
+    if: github.repository == 'oracle/graalvm-reachability-metadata'
     runs-on: ubuntu-22.04
     timeout-minutes: 5
     permissions: write-all


.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   get-all-libraries:
     name: "📋 Get list of all supported libraries with newer versions"
     # Opens PRs, with labels and assignees. Works only on the main repo.
-    if: github.repository == 'oracle/graalvm-reachability-metadata'
+    if: github.repository == 'jormundur00/graalvm-reachability-metadata'
```
vjovanov: Change this one.

PR #964, opened 2026-01-22
Bump buildtools version and allow mavenLocal buildtools versions to run by default [1/2]

jormundur00:
## What does this PR do?

In this PR, we bump the `nativeBuildtools` version to the currently latest version (`0.11.4`). As this version uses a newer gradle, we have to explicitly define the junit plugin `testRuntimeOnly "org.junit.platform:junit-platform-launcher"`.

Also in this PR, we add the ability to use local versions of the native build tools located in the local maven repository. As we have to change the tests in all `settings.gradle` files (which will spawn more jobs than the GitHub Actions CI limit), this part of the PR will be split into two separate PRs. On the second PR, we will also implement the usage of SNAPSHOT NBT versions in the CI.


jormundur00:
> There is a merge commit we need to get rid off.

I accidentally merged when rebasing to the master. I'll squash it when merging.

Issue #968, opened 2026-01-26
Bug: Metadata repo doesn't work with the latest buildtools snapshot


jormundur00:
### Describe the bug

The schemas directory location on the reachability metadata repository doesn't have the same structure as the metadata we package with new releases. This leads to issues with schema checking.

### Steps to reproduce

```markdown
Change the buildtools version in `libs.versions.toml` to the latest snapshot (`0.11.5-SNAPSHOT`, the snapshot needs to be built locally), then run any command that uses metadata, such as: `./gradlew test -Pcoordinates=ch.qos.logback:logback-classic:1.4.9 --info`.
```

### Expected behavior

The command should run.

### Logs and error messages

```text
[Incubating] Problems report is available at: file:///home/jovan/Work/graalvm-reachability-metadata/tests/src/ch.qos.logback/logback-classic/1.4.1/build/reports/problems/problems-report.html
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':nativeTestCompile'.
> Failed to create service 'nativeConfigurationService'.
   > Could not create an instance of type org.graalvm.buildtools.gradle.internal.GraalVMReachabilityMetadataService.
      > The configured GraalVM reachability metadata repository at /home/jovan/Work/graalvm-reachability-metadata/metadata does not contain a 'schemas' directory. Requires GraalVM Reachability Metadata 0.3.33 or newer which packages the required schemas.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to generate a Build Scan (Powered by Develocity).
> Get more help at 
https://help.gradle.org.
Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.


You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
BUILD FAILED in 5s

For more on this, please refer to https://docs.gradle.org/9.1.0/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
8 actionable tasks: 8 up-to-date

> Task :tckTest FAILED

[Incubating] Problems report is available at: file:///home/jovan/Work/graalvm-reachability-metadata/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':tckTest'.
> Test for ch.qos.logback:logback-classic:1.4.9 failed with exit code 1.

* Try:
[...]

PR #969, opened 2026-01-26
Make the metadata directory match the directory packaged with releases

jormundur00:
## What does this PR do?

Currently, the result of `./gradlew package` doesn't 1:1 match the structure of the `metadata` directory in the repository. This leads to the inability to run newer releases of buildtools on the reachability metadata repository (starting with `0.11.5-SNAPSHOT`), as there we begin verifying the existence of schemas, that currently get packaged in `metadata/schemas`.

To mitigate this, in this PR we move the metadata directory to have the same structure as the metadata that is packaged. For this goal, we move the `schemas` directory, as well as the `library-and-framework-list.json` file to the `metadata` subdirectory. As all this data is used by buildtools (either to verify correct file formats through schemas, or in the dynamic access section of the build report in the case of `library-and-frameworks-list.json`), it is reasonable to consider these files metadata (just not "reachability" metadata). Reachability metadata releases already package `schemas` and the library list inside of the `metadata` directory, we just match that structure here.

Also in this PR, we remove an unused test-project-index schema and file, as it is unused code (other than the example we define it in).

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/968

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,7 +2,7 @@ metadata/*       @oracle/graalvm-reachability-maintainer
 tests/src/*      @oracle/graalvm-reachability-maintainer
 tests/tck-build-logic/* @vjovanov
 tests/tck-build-logic/src/main/resources/allowed-docker-images/* @matneu
-library-and-framework-list.json @fniephaus
+metadata/library-and-framework-list.json @fniephaus
 .github/* @vjovanov
 docs/*    @vjovanov
 README.md @vjovanov


tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -206,8 +206,9 @@ List<String> diffCoordinates(String baseCommit, String newCommit) {
         if (changed.get("metadata") != null) {
             changedEntries = changed.get("metadata")
                     .stream()
-                    .map(m -> m.toString().split("reachability-metadata/metadata/")[1])
-                    .filter(m -> !m.equalsIgnoreCase("index.json"))
+                    .map(p -> metadataRoot().relativize(p).toString().replace('\\', '/'))
+                    .filter(rel -> !rel.equalsIgnoreCase("library-and-framework-list.json"))
+                    .filter(rel -> !rel.startsWith("schemas/"))
```
jvukicev, line 211: Changing any file in the `metadata` directory previously failed CI if `metadata/index.json` hasn't been changed. I've added explicit ignores of this rule for the new files, as this is a simple solution, and this rule/check will be removed in a soon-to-be-merged PR anyways (https://github.com/oracle/graalvm-reachability-metadata/pull/960).

metadata/schemas/library-and-framework-list-schema-v1.0.0.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://raw.githubusercontent.com/oracle/graalvm-reachability-metadata/master/schemas/library-and-framework-support-schema-v1.0.0.json",
+  "$id": "https://raw.githubusercontent.com/oracle/graalvm-reachability-metadata/master/metadata/schemas/library-and-framework-support-schema-v1.0.0.json",
```
vjovanov, line 2: Do we need  new version of this schema?

metadata/schemas/metadata-root-index-schema-v1.0.0.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://raw.githubusercontent.com/oracle/graalvm-reachability-metadata/master/schemas/metadata-root-index-schema-v1.0.0.json",
+  "$id": "https://raw.githubusercontent.com/oracle/graalvm-reachability-metadata/master/metadata/schemas/metadata-root-index-schema-v1.0.0.json",
```
vjovanov, line 2: I think they can stay in the same spot if we remove the unnecessary one?

metadata/schemas/library-and-framework-list-schema-v1.0.0.json
jvukicev: Buildtools will see no changes to the packaging (all changes are internal to the metadata repo), so there is no need to bump the schema version (as nothing changed inside of it content-wise, only its location).

metadata/schemas/metadata-root-index-schema-v1.0.0.json
jvukicev: Unfortunately, we still have to move these schemas inside of the `metadata` directory, as thats where buildtools expects them (currently they are packaged there, so metadata releases already see them as if they were in `metadata/schemas`, but CI and local runs here will fail as schemas are located in the root and not in `metadata`).

Issue #966, opened 2026-01-22
Feature: Allow CI to run with SNAPSHOT versions of native build tools


jormundur00:
### Is your feature request related to a problem?

Currently, it's impossible to use a non-maven-central-released version of native build tools in the CI. This in turn makes it impossible to test "breaking" changes in the buildtools repository with our current CI.

### Describe the solution you'd like

We should allow the CI to pull the latest SNAPSHOT version from the native build tools repository and upload it to the local maven repository with `./gradlew publishToMavenLocal --no-parallel`. After this, we should be able to use that version in the CI.

PR #975, opened 2026-01-28
Add github action to use latest buildtools snapshot and add it to CI

jormundur00:
## What does this PR do?

In this PR we introduce the `setup-native-build-tools` GitHub action that builds the latest Native Build Tools snapshot and sets the buildtools version in `libs.versions.toml` to that snapshot. This allows us to test the Reachability Metadata repositories CI with the latest buildtools changes. This is a requirement for merging PR whose changes create circular dependencies between the two repositories, while also allowing the CI to run without breaking in the time it takes to release new buildtools releases.

The newly added action is fully-customizable through its inputs, where `nbt-ref` can be used to make specific PRs run their CI on different Native Build Tools branches, if that is needed (and it will be needed for https://github.com/oracle/graalvm-reachability-metadata/pull/960).

We can most likely stop using this action in our workflows once the new release of NBT is released (with the big refactoring changes), to stay consistent with "stable" NBT releases, but we should still keep the action in the repository for future changes like this. The `test-all-metadata` workflow was ran on this PR with the latest NBT snapshot and it passes ([workflow run](https://github.com/oracle/graalvm-reachability-metadata/actions/runs/21439051592)).

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/966

### DIFF
--- a/.github/actions/setup-native-build-tools/action.yml
+++ b/.github/actions/setup-native-build-tools/action.yml
@@ -3,8 +3,8 @@ description: "Sets up JDK 17, checks out native-build-tools (preferring a same-n
 author: "GraalVM Reachability Metadata repository maintainers"
 
 inputs:
-  enabled:
-    description: "Whether to enable this setup step (set to 'false' to skip)."
+  enabled-by-default:
+    description: "Whether to enable this setup step even if no same-named branch is found."
     required: false
     default: "false"
   java-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-changed-infrastructure.yml
```diff
@@ -63,6 +63,9 @@ jobs:
       - name: "☁️ Checkout repository"
         uses: actions/checkout@v4
 
+      - name: "🛠️ Setup latest native-build-tools snapshot"
```
vjovanov, line 66: Well, this should be optional, right? We want to test with the latest release always, but on demand we can enable this.

jvukicev: I believe we will need to test with the snapshot only (so no latest-release) in the time between us merging the `metadata/index.json` removal PR (both here and on buildtools) and the buildtools release including that PR releasing on the central Maven repository (otherwise, no PRs would be able to be merged in the meantime).

After that, I think we can remove this step of the workflows and on demand return it if a scenario like this arises again.

kimeta:
Changes LGTM.

.github/workflows/test-changed-infrastructure.yml
vjovanov: Just add a variable to be able disable this.

jvukicev: Per our offline discussion I added the following:
1. The action is disabled by default (has a variable named `enabled-by-default` which is currently set to false),
2. The action WILL run (even if disabled by default) if there is a branch of the same name on the NBT repository,
3. Short section in `CI.md` explaining this.

`enabled-by-default` will then temporarily be set to true once `index.json` removal is merged (until the new NBT release is released and added here).

vjovanov (approved):

Issue #976, opened 2026-01-28
Feature: Add JaCoCo code coverage task


kimeta:
### Is your feature request related to a problem?

Currently, we don't have a way to track test code coverage of libraries. This metric would be especially useful in benchmarking AI test generation.

### Describe the solution you'd like

We need to add support for the `jacocoTestReport` task to measure how much of the library’s code is covered by tests. The coverage report should focus exclusively on the JARs of the provided library.

PR #977, opened 2026-01-28
Add jacocoTestReport task

kimeta:
## What does this PR do?

Fixes: #976

This PR introduces support for the `jacocoTestReport` task, enabling us to measure how much of the library’s code is covered by the test suite. Report configuration is set to generate coverage that focus exclusively on the provided library JARs, excluding unrelated external dependencies. This enhancement will help us better monitor and improve our test coverage.
This will be particularly beneficial for benchmarking and monitoring AI test generation.

### DIFF
--- a/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
+++ b/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
@@ -131,7 +131,7 @@ tasks.withType(org.gradle.testing.jacoco.tasks.JacocoReport).configureEach { r -
     }
 
     if (matched.isEmpty()) {
-        throw new GradleException("JaCoCo report generation failed. No eligible artifacts found on the classpath for ${libraryGAV}.")
+        throw new GradleException("JaCoCo report generation failed. No eligible artifacts found on the classpath for ${libraryGAV}.\nPlease verify that the provided artifact exists.")
     }
 
     // Produce one deterministic directory of .class files that represents exactly what the JVM would load at runtime for a given library jar, without duplicates


tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
```diff
@@ -92,6 +93,41 @@ tasks.named('test') {
...
+        csv.required = false
+    }
+
+    // Collect resolved dependency artifacts from the test runtime classpath
+    def artifacts = configurations.testRuntimeClasspath.resolvedConfiguration.resolvedArtifacts
+
+    def matched = artifacts.findAll { a ->
+        (a.moduleVersion.id.group == targetGroup ) && (a.moduleVersion.id.name == targetArtifact)
+    }.collect { it.file }
+    if (matched.isEmpty()) {
```
jvukicev, line 133: Are we sure this fallback is safe? Isn't this going to include all dependencies of the test (except the ones in the 3 hardcoded groups).

Is there a situation other than the user providing wrong `-Pcoordinates` where this fallback would trigger? If not, wouldn't it be better to just fail the task altogether if there are no matched artifacts (as opposed to checking the coverage of all dependencies, which I assume would be low for library-specific tests)?

jormundur00 (commented):
I think we should document this task briefly in `DEVELOPING.md`. Being unfamiliar with JaCoCo, it took a second to figure out what the task did (and that it was a wrapper for a task of the same name).

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
```diff
@@ -92,6 +93,41 @@ tasks.named('test') {
...
+        matched = artifacts.findAll { a ->
+            !(a.moduleVersion.id.group.startsWith('org.junit')
```
vjovanov: What are these?

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/JacocoTestReportInvocationTask.java
```diff
@@ -0,0 +1,29 @@
+/*
```
vjovanov, line 1: Add a test for this.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
kimeta: I think you are completely right. There should always be an artifact that completely matches the previous filter, if provided coordinates are correct.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
kimeta: This will be removed. If matched is empty we should fail the task.

docs/DEVELOPING.md
```diff
@@ -98,6 +98,27 @@ Each stage of the testing can be run with `-Pcoordinates=[group:artifact:version
 ./gradlew test -Pcoordinates=[group:artifact:version|k/n|all]
 ```
 
+### Coverage (JaCoCo)
+
+Purpose
```
jvukicev: Small nit: I'd go without these `Purpose`, `How it works`, `Usage` sections here, just to stay consistent with other sections in this file (e.g., you can do `Example/s:` for the `Usage` subsection, for the other subsections just follow the format of other sections). Not required, just a nit.

build.gradle
```diff
@@ -126,7 +138,7 @@ static List<Map> testInfraCommands(String gradleCmd, String target) {
 
 tasks.register('testInfra') { t ->
     t.group = "verification"
-    t.setDescription("""Runs clean and pullAllowedDockerImages, then runs (${testInfraCommands('', '').collect { it.name }.join(', ')}) concurrently with isolated logs for a single coordinate.
+    t.setDescription("""Runs clean, pullAllowedDockerImages and jacocoTestReport, then runs (${testInfraCommands('', '').collect { it.name }.join(', ')}) concurrently with isolated logs for a single coordinate.
```
jvukicev: Is there a reason `jacocoTestReport` is ran separately from `testInfraCommands` (like `pullAllowedDockerImages`)? Why not just add it to the list?

kimeta: It should be ran with `testInfraCommands`. Fixed that.

tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
```diff
@@ -92,6 +94,101 @@ tasks.named('test') {
...
+        // Fallback for relocated or substituted modules: the requested G:A coordinates may resolve
+        // to a different group in published metadata, so no resolved artifact carries the original group.
+
+        matched = artifacts.findAll { a ->
+            (a.moduleVersion.id.name == targetArtifact) && (a.moduleVersion.id.version == targetVersion)
+        }.collect { it.file }
+    }
+
+    if (matched.isEmpty()) {
+        throw new GradleException("JaCoCo report generation failed. No eligible artifacts found on the classpath for ${libraryGAV}.")
```
vjovanov: Good, but what should I do to fix this? Can this happen?

kimeta: This can happen if there is an error in fetching the library artifact, or library's substitution has different artifact name than the tested one. Both are rare cases. Added more verbose error.

Issue #847, opened 2025-12-05
Feature: Remove global metadata index.json from the repository


jormundur00:
### Is your feature request related to a problem?

As a part of the GraalVM Reachability Metadata 1.0 release, we should remove the index file located in the `metadata` directory.

### Describe the solution you'd like

This index file IS parsed by buildtools, so we need to be sure that removing it won't break builds of buildtools by changing how data is parsed NBT-side, probably by moving some of the important fields to the remanining `index.json` file.

PR #983, opened 2026-02-02
Remove global metadata/index.json file and its usage

jormundur00:
## What does this PR do?

In this PR we remove the global `index.json` file located at `metadata/index.json` to simplify new metadata addition (so we only need to create/maintain one `index.json` file located at `metadata/groupId/artifactId/index.json` per library artifact).

To do this, we move the `allowed-packages` and `requires` fields to the inner `index.json`, as this is data that only the global one had previously. As the `module` field was used to map between the global and inner index files, we remove this field as it was redundant (the information it keeps can always be resolved from the inner-index's path). All library metadata resolution is now based on file paths (so every library is strictly in `metadata/<groupId>/<artifactId>`). The `scaffold` and other tasks which generate these metadata directories have also been updated accordingly to match the new format.

As these changes include relatively major changes to the index structure, the remaining schema version was bumped from `1.0.0` to `2.0.0`.

As running changes in this PR depend on the [buildtools PR](https://github.com/graalvm/native-build-tools/tree/jormundur00/gh-802) which adds support for the removal of the `metadata/index.json` file, and this PR depends on buildtools working with the new format to successfully pass CI, the merge/release process of these PRs should have the following order:

1. This PR should be merged first (running CI on the branch of the buildtools PR to complete successfully),
2. The next reachability metadata release should be dispatched so that buildtools can use it in their corresponding PR,
3. The buildtools PR should then include the bump to this new reachability metadata version and get merged ASAP,
4. This repository's CI will work on the latest SNAPSHOT of buildtools until the new NBT release is released (which should be scheduled as soon as the buildtools PR is merged),
5. We open a follow-up PR here to use the newest release of buildtools and no longer use snapshots in the CI.

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/847


tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -287,33 +268,32 @@ private boolean metadataIndexContainsChangedEntries(Set<String> changedCoordinat
...
-            }
-        }
+        Set<String> dirs = new LinkedHashSet<>();
+        Path root = metadataRoot();
+
+        try (Stream<Path> groupDirs = Files.list(root)) {
+            groupDirs.filter(Files::isDirectory).forEach(g -> {
+                String gName = g.getFileName().toString();
+                // Filter by group if provided
+                if (groupId != null && !groupId.equals(gName)) return;
```
kimeta, line 262: When can `groupId`  be null?

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -287,33 +268,32 @@ private boolean metadataIndexContainsChangedEntries(Set<String> changedCoordinat
...
+
+                try (Stream<Path> artifactDirs = Files.list(g)) {
+                    artifactDirs.filter(Files::isDirectory).forEach(a -> {
+                        String aName = a.getFileName().toString();
+                        // Filter by artifact if provided
+                        if (artifactId != null && !artifactId.equals(aName)) return;
```
kimeta, line 268: Same question for `artifactId`.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -287,33 +268,32 @@ private boolean metadataIndexContainsChangedEntries(Set<String> changedCoordinat
...
-    @SuppressWarnings("unchecked")
     Set<String> getMatchingMetadataDirs(String groupId, String artifactId) {
-        Set<String> dirs = new HashSet<>();
-        List<Map<String, ?>> index = (List<Map<String, ?>>) readIndexFile(metadataRoot());
-        for (Map<String, ?> library : index) {
-            if (coordinatesMatch((String) library.get("module"), groupId, artifactId)) {
-                if (library.containsKey("directory")) {
-                    dirs.add((String) library.get("directory"));
-                }
-                if (library.containsKey("requires")) {
```
kimeta, line 299: In the changed code you removed adding directories, which are present in the requires field. Is this intentional? We should probably still parse requires?

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
```diff
@@ -361,22 +343,25 @@ public List<String> getMatchingCoordinates(String coordinateFilter) {
...
-                                .forEach(t -> matchingCoordinates.add(entry.get("module") + ":" + t));
-                    } else { // We have a specific version pinned.
-                        if (metadataRoot().resolve(coordinates.get(0)).resolve(coordinates.get(1)).resolve(version).toFile().exists()) {
-                            matchingCoordinates.add(entry.get("module") + ":" + version);
-                        }
-                    }
+                if (testedVersions == null) continue;
+
+                // Derive group/artifact from directory path (index is per artifact)
+                String g = fullDir.getParent().getFileName().toString();
```
kimeta, line 345: Is there a more elegant approach?

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/utils/MetadataGenerationUtils.java
```diff
@@ -159,31 +159,33 @@ public static void makeVersionLatestInIndexJson(ProjectLayout layout, Coordinate
...
-                testVersion,
+                Boolean.TRUE, // latest
+                null, // override
+                null, // default-for
+                newCoords.version(), // metadata-version
+                testVersion, // test-version
                 testedVersions,
-                null
+                null, // skipped-versions
+                null, // allowed-packages
```
kimeta: Now that `allowed-packages` and `requires` are fields in each version entry, it's a bit tricky deciding what should be their values when entry is automatically generated. Probably best solution would be to copy that information from the previous `latest` version.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/utils/MetadataGenerationUtils.java
```diff
@@ -159,31 +159,33 @@ public static void makeVersionLatestInIndexJson(ProjectLayout layout, Coordinate
...
+                null // requires
         );
-        entries.addFirst(newEntry);
+        entries.add(0, newEntry);
```
kimeta: You can revert this change.

tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/MetadataFilesCheckerTask.java
```diff
@@ -399,18 +399,30 @@ private List<String> getAllowedPackages() {
...
+        List<Map<String, Object>> entries = ((List<Object>) parsed).stream()
+                .map(e -> (Map<String, Object>) e)
+                .toList();
+
+        // With "module" removed, index.json is per artifact path, so simply use the first entry that defines allowed-packages.
+        for (Map<String, Object> entry : entries) {
+            Object ap = entry.get("allowed-packages");
+            if (ap != null) {
+                if (ap instanceof List) {
+                    return (List<String>) ap;
```
kimeta: This could be a potential problem if `allowed-packages` can differ between versions. Can this happen?

jvukicev: Good catch. Since we have the `metadata-version` here (which uniquely distinguishes an `index.json` entry), we can retrieve a specific versions `allowed-packages` here. Fixed.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
jvukicev: I left it as it is for now, as I don't see a much more elegant way to do this that wouldn't break semantics.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
jvukicev: We discussed this offline and decided that parsing `requires` here is redundant. We already parse it on native build tools when building the classpath (deciding what metadata to include), and adding it here does functionally nothing.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
jvukicev: Explained in the `groupId` comment above.

tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
jvukicev: `groupId` and `artifactId` can be null when using coordinates such as `-Pcoordinates=all`. In those cases, nothing gets filtered, so all metadata directories are added to `dirs`.

kimeta (approved):

Issue #1006, opened 2026-02-12
Bug: New ch.qos.logback:logback-classic metadata is incompatible with tests on native build tools


jormundur00:
### Describe the bug

The [LayeredApplicationFunctionalTest](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/LayeredApplicationFunctionalTest.groovy) on native build tools currently fails when using the new `1.5.7` metadata for `ch.qos.logback:logback-classic`. It passes normally when using the previous latest `1.4.1` version of the metadata. This is not a layers issue, as refactoring this test to not use layers and rerunning it ends up with the same result.

### Steps to reproduce

```markdown
To reproduce, bump the version of `logback-classic` in the `build.gradle` on the test to anything `1.5.7+`, as that will make the buildtools use the newly added metadata.
```

### Expected behavior

The test should complete regularly, like it does with `1.4.1` metadata.

### Logs and error messages

```text
https://github.com/graalvm/native-build-tools/actions/runs/21896353516/job/63220968214


    ========================================================================================================================

    Finished generating 'layered-java-application' in 35.0s
    .

    [native-image-plugin] Native Image written to: /tmp/spock_can_build_a_native__0_testDirectory4058962804547442956/build/native/nativeCompile


    > Task :nativeRun


    BUILD SUCCESSFUL
     in 36s

    7 actionable tasks: 3 executed, 4 up-to-date


LayeredApplicationFunctionalTest > can build a native image using layers FAILED
    org.spockframework.runtime.ConditionNotSatisfiedError at LayeredApplicationFunctionalTest.groovy:87

LayeredApplicationFunctionalTest > can build a layered Micronaut application SKIPPED

2 tests completed, 1 failed, 1 skipped

> Task :native-gradle-plugin:functionalTest FAILED

FAILURE: Build failed with an exception.
```

### GraalVM version

GraalVM 25.0.2

### Operating system

Linux

### CPU architecture

AMD64 (x86-64)

PR #1009, opened 2026-02-13
Add missing logback-classic metadata.

kimeta:
## What does this PR do?

Fixes: #1006 

After adding metadata for the `ch.qos.logback:logback-classic:1.5.7`, [LayeredApplicationFunctionalTest](https://github.com/graalvm/native-build-tools/blob/master/native-gradle-plugin/src/functionalTest/groovy/org/graalvm/buildtools/gradle/LayeredApplicationFunctionalTest.groovy) on native build tools fails when using logback-classic version 1.5.7 or later.  This PR corrects the recently added metadata so the test passes.

### DIFF
--- a/metadata/ch.qos.logback/logback-classic/1.5.7/reflect-config.json
+++ b/metadata/ch.qos.logback/logback-classic/1.5.7/reflect-config.json
@@ -575,21 +575,7 @@
   },
   {
     "condition": {
-      "typeReachable": "ch.qos.logback.core.model.processor.DefaultProcessor$$Lambda/0x00007d00bc1874c0"
-    },
-    "name": "ch.qos.logback.core.encoder.Encoder",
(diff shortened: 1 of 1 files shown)


jormundur00 (approved):

metadata/ch.qos.logback/logback-classic/1.5.7/reflect-config.json
```diff
@@ -631,7 +617,7 @@
   },
   {
     "condition": {
-      "typeReachable": "ch.qos.logback.core.model.processor.DefaultProcessor$$Lambda/0x00007dff1019a570"
+      "typeReachable": "ch.qos.logback.core.util.OptionHelper"
```
vjovanov, line 620: Agent bug, luckily that will not happen in the newest version.

Issue #790, opened 2025-11-20
Feature: Badges for number of artifacts, number of version, percentage of metadata entries covered


vjovanov:
### Is your feature request related to a problem?

It is currently hard to understand how many libraries are covered in the ecosystem.

### Describe the solution you'd like

These badges should be on the first line of the README.md.

PR #1026, opened 2026-02-26
Add tested libraries table to README

thomaswue:
## Summary
- add a new "Tested Libraries" section to README with a markdown table of currently tested library coordinates
- sort all library coordinates alphabetically in the first column for easier lookup
- include a short one-sentence description for each listed library in the second column

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,118 @@ Open an [update ticket](https://github.com/oracle/graalvm-reachability-metadata/
 
 ---
 
+### Tested Libraries
+
+| Library | Description |
+| --- | --- |
+| `berkeleydb:je` | Berkeley DB Java Edition is an embedded key-value database library for Java applications. |
+| `ch.qos.logback.contrib:logback-jackson` | Logback Jackson adds Jackson-based JSON support for Logback appenders and layouts. |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,6 +28,118 @@ Open an [update ticket](https://github.com/oracle/graalvm-reachability-metadata/
 
 ---
 
+### Tested Libraries
```
vjovanov, line 31: We have this planned via badges: https://github.com/oracle/graalvm-reachability-metadata/issues/790

I find that having this list in the README.md is not maintainable: 1) It will soon be too long (1000 libraries), and 2) it needs to be generated by the CI.

Challenge for the coding agent:
 1. We should move this table into a separate file called `supported-libraries.md`. Each line in this table should also support a range of library versions that we test in this repo (as the thrid column).
 2. In the CI, we should automatically update `supported-libraries.md` file for every merged PR as well as the JSON file that contains the information for `shields.io`.
 3. The `README.md` page should then have a badge (at the top) that leads to `supported-libraries.md`. The badge should contain the number of libraries and the number of different versions that are supported.
 4. The badge should be implemented via the `shields.io` service. Here is how this achieved: host a JSON file with a stable URL that contains the count of libraries, and versions. Humans will then point a Shields dynamic badge to that URL and path.

vjovanov:
We have the stats now, and the description for each library. @jormundur00 please add this table to our stats file.

Issue #1061, opened 2026-03-12
Bug: version resolution fails for versions that reuse metadata-version/test-version


oleksii-tumanov:
### Describe the bug

Exact resolution fails for versions that are supported through shared metadata and shared tests.

One metadata entry can cover multiple library versions via tested-versions. 

For example, ch.qos.logback:logback-classic:1.5.12 is supported via:
  1. metadata/ch.qos.logback/logback-classic/1.5.7
  2.  tests/src/ch.qos.logback/logback-classic/1.4.1

  However, the TCK harness assumed the requested version had to match metadata-version, so selection doesn't not work correctly.

### Steps to reproduce

```markdown
1. Open metadata/ch.qos.logback/logback-classic/index.json.
2. Note that 1.5.12 is listed in tested-versions, while the same entry uses 'metadata-version' 1.5.7 and test-version 1.4.1.
3. Run ./gradlew listCoordinates -Pcoordinates=ch.qos.logback:logback-classic:1.5.12 --console=plain
4. Before the fix, the command reported no matching coordinates.
5. The same resolution issue also affected exact-version test execution for shared metadata/test entries.
```

### Expected behavior

If a version is listed in `tested-versions`, the harness should treat it as a valid exact coordinate.

When resolving tests, it should:
- locate the matching index entry through `tested-versions`
- use `test-version` when present
- otherwise fall back to `metadata-version`.

### Logs and error messages

```text
Before the fix:


> Task :listCoordinates
No matching coordinates found!
coordinates=


A related path in the harness also assumed exact equality between requested version and `metadata-version`, which breaks shared test layouts.
```

### Additional context

This is a repository infrastructure bug, not a single-library metadata bug.

It affects contributor workflow because the documented exact-version commands in `README.md` / `docs/CONTRIBUTING.md` do not work for libraries that intentionally share metadata/test directories across several versions.

I have a local fix and regression tests for this.

### GraalVM version

Oracle GraalVM 21.0.10+8.1

### Operating system

macOS

### CPU architecture

AARCH64 (ARM64)

PR #1062, opened 2026-03-12
Fix shared metadata coordinate resolution

oleksii-tumanov:
Fix for #1061 

## What does this PR do?
This PR fixes resolution for versions that are supported through shared metadata and tests .

  For example `ch.qos.logback:logback-classic:1.5.12` is supported via:
  - `metadata/ch.qos.logback/logback-classic/1.5.7`
  - `tests/src/ch.qos.logback/logback-classic/1.4.1`

 The TCK harness assumed the requested version had to match `metadata-version`. This change resolves exact versions through `tested-versions` and adds regression tests for that behavior. 

 1. tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/TckExtension.java
    Reads metadata index files and resolves metadata/test directories
    
 2. tests/tck-build-logic/src/test/java/org/graalvm/internal/tck/harness/TckExtensionTests.java
    Creates temporary repository layouts on disk for harness regression tests.

## Checklist before merging
- [x] I have considered including reachability metadata directly in the library or the framework (see [our contributing document](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md))
- [x] For all tests where I am not the sole author, I have added a comment that proves I may publish them under the specified license (see [the tests section from our contributing document](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#tests))
- [x] I have properly formatted metadata files (see [our formatting guide](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#format-metadata-files))
- [x] I have added thorough tests (see [the tests section from our contributing document](https://github.com/oracle/graalvm-reachability-metadata/blob/master/CONTRIBUTING.md#Tests))
- [x] I have filled all places where my pull request accesses files, network, docker, or any other external service (see the section above)

### DIFF
--- a/.github/workflows/checkstyle.yml
+++ b/.github/workflows/checkstyle.yml
@@ -3,11 +3,17 @@ name: "Check code style"
 on:
   pull_request:
 
+concurrency:
+  group: "workflow = ${{ github.workflow }}, ref = ${{ github.event.ref }}, pr = ${{ github.event.pull_request.id }}"
+  cancel-in-progress: true
+
 jobs:
-  checkstyle:
-    name: "📋 Check style according to checkstyle.xml"
+  detect-relevant-changes:
+    name: "📋 Detect relevant file changes"
     runs-on: "ubuntu-22.04"
-    timeout-minutes: 15
+    timeout-minutes: 5
+    outputs:
+      relevant-files-changed: ${{ steps.filter.outputs.changed }}
     steps:
       - name: "☁️ Checkout repository"
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/checkstyle.yml
```diff
@@ -21,16 +27,88 @@ jobs:
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: 'graalvm'
+          java-version: '21'
+
+      - name: "Run Spotless"
+        run: ./gradlew spotlessCheck --console=plain
+
+  checkstyle-shards:
+    name: "📋 Checkstyle shard ${{ matrix.coordinates }}"
```
vjovanov, line 50: @jormundur00 Why don't we run checkstyle only on changed tests? We have the ability to generate changed coordinates, and we could feed it to checkstyle. The style of the harness should always be checked, no matter which coordinates we use.

Soon we will have 1k libraries with multiple versions, so testing them all every time is expensive and unnecessary.

jvukicev: I think we can reuse `generateChangedCoordinatesMatrix` (see `test-changed-metadata.yml`) to get the list of changed metadata/test coordinates, so we can only run checkstyle on them (ignoring `ci.json`s os/java-version specifications).

.github/workflows/checkstyle.yml
```diff
@@ -21,16 +27,88 @@ jobs:
...
+      matrix:
+        coordinates:
+          - "1/8"
+          - "2/8"
+          - "3/8"
+          - "4/8"
+          - "5/8"
+          - "6/8"
+          - "7/8"
+          - "8/8"
```
jvukicev, line 66: Good work, generally looks good to me! We should have this not work with batches, but actually use the existing `generateChangedCoordinatesMatrix` task to only check style for libraries whose files actually changed (and not all libraries every time in batches).

I think we can merge this PR, as it already fixes a lot (and a different issue), and I can add this feature to the checkstyle in a follow up.

Issue #1272, opened 2026-03-24
Testing new library version


vjovanov:
If a tested library version is added, either by bot or manually, when merging with PR, the same library version has to be tested with all GraalVM versions mentioned in the `ci.json`.

When running with verify library versions, we need to test with both versions of GraalVM, GraalVM 25 and GraalVM Latest, and if any of them fails we need to make this as a failing test and then we also need to include that into the errors.

PR #1286, opened 2026-03-25
Run CI tests on index.json changes and run verify-new-library-version-compatibility.yml with all supported OS/JDK combinations

jormundur00:
## What does this PR do?

In this PR we introduce two major CI changes:

1. `index.json` changes in opened PRs now trigger `test-changed-metadata` and `test-affected-spring-aot` tests. The new implementation of change detection uses a "smarter" detection which will match the changed `index.json` entries, so we only test the metadata/test whose entries have changed (and not all entries in an `index.json` file),
2. `verify-new-library-version-compatibility.yml` now runs tests on all environments specified by `ci.json` (currently being JDK25/`latest-ea` and `ubuntu-latest`). The workflow test results across all different env combinations for a specific test and adds tested versions in the generated PR only if the test passed for all env combinations. For every failed combination, the workflow will aggregate their failure outputs into a GitHub issue. The issue will include a subsection for each detected failure. When the PR is generated for the passing tested versions, the description of this PR and the body of the commit that is generated by it will include a list of all newly-added tested versions, for easier tracking.

New workflow run: https://github.com/oracle/graalvm-reachability-metadata/actions/runs/23597221028
New failure issue format: https://github.com/oracle/graalvm-reachability-metadata/issues/1292
New tested-version-add PR format: https://github.com/oracle/graalvm-reachability-metadata/pull/1295 (ignore non-bot commits, as they are from this PR)

Fixes: https://github.com/oracle/graalvm-reachability-metadata/issues/1272

### DIFF
--- a/.github/workflows/verify-new-library-version-compatibility.yml
+++ b/.github/workflows/verify-new-library-version-compatibility.yml
@@ -106,6 +106,7 @@ jobs:
           ARTIFACT_ID="$(echo "${{ matrix.name }}" | cut -d: -f2)"
 
           readarray -t VERSIONS < <(echo '${{ toJson(matrix.versions) }}' | jq -r '.[]')
+          ISSUE_TITLE_PREFIX="[Automation] "
           FIRST_TESTING_VERSION="${VERSIONS[0]}"
           GAV_COORDINATES="$GROUP_ID:$ARTIFACT_ID:$FIRST_TESTING_VERSION"
           TITLE_END=" fails for $GAV_COORDINATES"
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -305,63 +500,77 @@ jobs:
...
+                  n = asorti(updates, libraries)
+                  for (i = 1; i <= n; i++) {
+                    library = libraries[i]
+                    printf "#### `%s`%s\n", library, updates[library]
+                  }
+                }
+              ' successful_updates.txt
+            } > "$PR_BODY_FILE"
+            gh pr create \
+              --title "[Automation] Update supported library versions $(date '+%Y-%m-%dT%H:%M')" \
```
vjovanov: Why drop a variable for "[Automation]"?

.github/workflows/verify-new-library-version-compatibility.yml
```diff
@@ -305,63 +500,77 @@ jobs:
...
+              --body-file "$PR_BODY_FILE" \
+              --reviewer "kimeta" \
```
vjovanov: Add me as the reviewer as well. I anyhow merge them all the time.

.github/workflows/verify-new-library-version-compatibility.yml
jvukicev: The variable wasn't used here in the old implementation. As we now have access to it here, I reused it now.

Issue #1610, opened 2026-04-02
Set up Macaron checks for GitHub Actions workflows


vjovanov:
Context
- oracle/graal PR #13217 was merged (https://github.com/oracle/graal/pull/13217), enabling Macaron checks for GHA workflows there.

Request
- Set up equivalent Macaron workflow-check jobs for:
  - oracle/graalvm-reachability-metadata
  - oracle/graalvm-native-build-tools (NBT)

Acceptance Criteria
- Macaron workflow-check jobs are configured and running for metadata and NBT repositories.
- CI shows the new checks as expected in PRs for both repos.
- Any repo-specific setup/docs updates are added where needed.

PR #1611, opened 2026-04-02
Set up Macaron check for GitHub Actions workflows

vjovanov:
## Summary
- add a dedicated GitHub Actions workflow to run Macaron workflow checks
- trigger the check on PR changes under `.github/workflows/**` and via manual dispatch
- configure `policy_purl` for `oracle/graalvm-reachability-metadata`

## Testing
- added workflow file only; validation is via workflow execution on PR

Ref: #1610


.github/workflows/macaron-check-github-actions.yml
```diff
@@ -0,0 +1,25 @@
+name: "Check GitHub Actions with Macaron"
```
vjovanov, line 1: Leave the comment on why this is needed.

vjovanov:
Updated in this branch: the Macaron workflow now also triggers on `.github/actions/**` (in addition to `.github/workflows/**`), so local composite actions are covered too.

PR #1724, opened 2026-04-14
Avoid consulting library-and-framework-list during triage

### LINKED ISSUE #1723: Feature: Avoid consulting library-and-frameworks-list.json when triaging issues
### Is your feature request related to a problem?

With the release of the `1.0-M1` of the metadata repo, we started moving towards supporting as much artifacts as possible from the Maven ecosystem. Currently, an opened issue will get automatically closed by our triaging workflow if the artifact in the issue exists in the `libraries-and-frameworks-list.json`.

### Describe the solution you'd like

We should avoid checking the `libraries-and-frameworks-list.json` when triaging. We should only consult the `check-library-support.sh` script.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

jormundur00:
## Summary

This PR removes `metadata/library-and-framework-list.json` from the automation paths involved in new library issue triage.

## What changed

- In `.github/workflows/triage-new-issues.yml`, removed the `group:artifact` derivation, the lookup against `metadata/library-and-framework-list.json`, and the close path that treated list membership as repository support.
- The triage workflow now uses `check-library-support.sh` as the only automated check for whether the requested coordinates are already supported, and otherwise continues to dependency graph expansion.
- In `.github/workflows/scripts/open-dependency-issues-and-link-blockers.js`, removed the `loadListedArtifacts()` path and the early filtering that added listed artifacts to `supportedGA`.
- That means listed artifacts are no longer suppressed from `toCreate`, so dependency issue creation is decided by `check-library-support.sh` and the existing exclusion rules rather than by presence in the library list.
- Updated the diagnostic comment emitted when no dependency issues are created so it no longer claims that list membership is a reason for suppression.

## Verification

- Parsed `.github/workflows/triage-new-issues.yml` successfully with Ruby YAML loading.
- Ran `node --check .github/workflows/scripts/open-dependency-issues-and-link-blockers.js` successfully.

Fixes: #1723

### DIFF
--- a/.github/workflows/scripts/open-dependency-issues-and-link-blockers.js
+++ b/.github/workflows/scripts/open-dependency-issues-and-link-blockers.js
@@ -423,23 +423,6 @@ function parseCreateTSV(createTSV) {
 }
 
 /**
- * Loads the curated list of artifacts already tracked by the repository.
- */
-function loadListedArtifacts(workspaceDir) {
-  const listedArtifactsPath = path.join(
-    workspaceDir,
-    'metadata',
-    'library-and-framework-list.json'
-  );
-  const listedArtifacts = JSON.parse(fs.readFileSync(listedArtifactsPath, 'utf8'));
-  return new Set(
-    listedArtifacts
-      .map((entry) => entry?.artifact)
-      .filter((artifact) => typeof artifact === 'string' && artifact.length > 0)
-  );
-}
-
-/**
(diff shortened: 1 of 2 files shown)


vjovanov (approved):

kimeta (approved):
Minor note: the removal of the library-list short-circuit means every dependency node now shells out to check-library-support.sh, which could slow down large dependency graphs.

PR #1730, opened 2026-04-14
Cache generateDynamicAccessReport results

### LINKED ISSUE #1729: Cache generateDynamicAccessReport when tracked classpath is unchanged
## Summary
`generateDynamicAccessReport` always reruns `nativeTestCompile` even when the tracked library classpath and relevant task inputs have not changed.

## Problem
The task is effectively uncached today, so repeated invocations rebuild native tests and regenerate the same dynamic-access JSON output. That makes the report workflow slower than it needs to be, especially when iterating on unrelated changes.

## Expected behavior
If the command inputs, metadata/test inputs, and resolved tracked library JAR contents are unchanged, `generateDynamicAccessReport` should reuse the existing dynamic-access output instead of invoking `nativeTestCompile` again.

## Proposed fix
- fingerprint the relevant report inputs, including the resolved tracked library JARs
- persist a small cache manifest alongside the task output
- skip execution when the fingerprint matches and the expected dynamic-access JSON output is still present
- keep printing the cached report paths so the task output remains usable

## Validation
- add focused tests for the new path fingerprinting helper
- run the `tests/tck-build-logic` test suite

vjovanov:
## Summary
- Adds a `generateDynamicAccessReport` task that wraps `nativeTestCompile` with declared inputs (tracked library JARs, test outputs, scoped metadata JSONs, `ci.json`, native-image args) and the dynamic-access output directory, so Gradle's UP-TO-DATE check skips rebuilding when nothing changed.
- Repoints `GenerateDynamicAccessReportInvocationTask` (per-coordinate dispatcher) at the new task.
- Auto-enables `-Ptck.generateDynamicAccessReport=true` when `generateDynamicAccessReport` is requested by name on the command line.

Closes #1729.

## Notes
- The output dir (`build/native/nativeTestCompile/dynamic-access`) lives inside `nativeTestCompile`'s output tree. Gradle should still cache correctly because the wrapper only declares this one subdirectory, but watch for "overlapping outputs" warnings under `--info` on first run.
- Task-name detection uses `gradle.startParameter.taskNames` and will not pick up abbreviated names (e.g. `gDAR`) or task-selector forms; pass `-Ptck.generateDynamicAccessReport=true` explicitly in those cases.

## Testing
- `../../gradlew test --stacktrace` (from `tests/tck-build-logic`)

### DIFF
--- a/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
+++ b/tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle
@@ -7,6 +7,8 @@
 import groovy.json.JsonSlurper
 import org.graalvm.internal.tck.utils.DynamicAccessUtils
 
+import org.gradle.api.tasks.PathSensitivity
+
 import java.util.jar.JarFile
 
 plugins {
@@ -57,7 +59,10 @@ overrideVal = overrideVal ?: "false"
 boolean override = overrideVal.toBoolean()
 
 String generateDynamicAccessReportVal = providers.gradleProperty('tck.generateDynamicAccessReport').getOrElse("false")
-boolean generateDynamicAccessReport = generateDynamicAccessReportVal.toBoolean()
+boolean generateDynamicAccessReportTaskRequested = gradle.startParameter.taskNames.any { taskName ->
+    taskName == "generateDynamicAccessReport" || taskName.endsWith(":generateDynamicAccessReport")
+}
+boolean generateDynamicAccessReport = generateDynamicAccessReportVal.toBoolean() || generateDynamicAccessReportTaskRequested
 
 // Determine native-image build arguments from ci.json.
 def ciJsonFile = tck.repoRoot.file("ci.json").get().asFile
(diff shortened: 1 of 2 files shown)


jormundur00 (approved):
Local review notes from `/tmp/graalvm-rm-pr1730-review`:

1. Non-blocking follow-up: the new `generateDynamicAccessReport` cache fingerprint in `GenerateDynamicAccessReportInvocationTask.computeInputFingerprint()` hashes the command text, repo inputs, and tracked library JAR contents, but it does not currently include execution-environment inputs like `GVM_TCK_NATIVE_IMAGE_MODE`, `JAVA_HOME`, or a GraalVM/native-image version signal. That means a cached report can theoretically be reused across mode or toolchain changes if the on-disk inputs stay the same.

2. Non-blocking follow-up: the added tests cover `DynamicAccessUtils.fingerprintPaths()` well, but there is still no task-level coverage for cache hit/miss behavior in `GenerateDynamicAccessReportInvocationTask` itself, especially for manifest invalidation and skip-path reuse.

Targeted local check run:
- `cd tests/tck-build-logic && ../../gradlew test --tests org.graalvm.internal.tck.utils.DynamicAccessUtilsTests --no-daemon`

Approving as requested.

vjovanov:
@jormundur00 I rewrote the PR. Please have a look again.

jormundur00 (changes requested):
No blocking findings on cache invalidation after re-reviewing the rewritten PR locally.

I verified the earlier invalidation concern does not apply here: `nativeTestCompile` already tracks the dynamic-access build args derived from the tracked library jars, and metadata changes flow into the native-image configuration directories consumed by the native build.

One blocking semantic issue remains: when `TrackDynamicAccess` finds no dynamic access calls, `build/native/nativeTestCompile/dynamic-access` is not created. The previous behavior treated that as a valid zero-results case, but the new `generateDynamicAccessReport` task turns it into a failure. In the current `AbstractLibraryStatsTask` flow, that failure causes stats generation to fall back to dynamic access `N/A`, so results that were previously reported as `0` calls will now be reported as `N/A`. That is incorrect for stats.

Local check run:
- `cd tests/tck-build-logic && ../../gradlew test --stacktrace --no-daemon`

vjovanov:
No blocking findings after re-reviewing the rewritten PR locally.

I re-checked the earlier cache invalidation concern and it does not apply here: invoking `generateDynamicAccessReport` now enables the dynamic-access native-image args on `nativeTestCompile`, and the wrapper task tracks the resolved library jars, test outputs, metadata JSON, `ci.json`, and `nativeImageArgs` as inputs.

I also re-checked the zero-results case. I do not see a blocking stats regression from a missing `build/native/nativeTestCompile/dynamic-access` directory on this branch: the harness `generateDynamicAccessReport` flow treats that case as non-fatal, and the stats parser already interprets a missing dynamic-access directory as 0 calls rather than dynamic access N/A.

Local check run:

```text
cd tests/tck-build-logic && ../../gradlew test --stacktrace --no-daemon
```

jormundur00 (approved):
No blocking findings after re-reviewing the current PR head locally.

I rechecked the current implementation of `generateDynamicAccessReport`, and the earlier concern about converting valid zero-call results into `N/A` no longer applies. The previous revision had a `doLast` failure when `build/native/nativeTestCompile/dynamic-access` was missing, but that failure path is no longer present in the current version.

I also re-verified the invalidation side: `nativeTestCompile` already tracks the dynamic-access build args derived from the tracked library jars, and metadata changes flow into the native-image configuration directories consumed by the native build, so the wrapper task is not the only place those inputs are modeled.

Local check run:
- `cd tests/tck-build-logic && ../../gradlew test --stacktrace --no-daemon`

PR #2028, opened 2026-04-21
Fix stale stats directory when addTestedVersion replaces a pre-release metadata version

### LINKED ISSUE #2027: addTestedVersion leaves stale stats directory when replacing a pre-release metadata version
## Bug

When `addTestedVersion` replaces a pre-release metadata version with the final release version, it rewrites the stats into the new `stats/<group>/<artifact>/<newVersion>/stats.json` location, but it does not remove the old `stats/<group>/<artifact>/<oldVersion>/` directory.

A concrete example is bumping `org.jboss.spec.javax.interceptor:jboss-interceptors-api_1.1_spec` from `1.0.0.Beta1` to `1.0.0`: the new `1.0.0` stats directory is created and updated correctly, but the old `1.0.0.Beta1` stats directory remains behind.

## Expected behavior

When a pre-release metadata version is replaced by the matching full release, `addTestedVersion` should move or update the stats and remove the old stats version directory if it is now empty.

## Impact

This leaves stale exploded stats content in the repository and makes the metadata-version move incomplete.

jormundur00:
## Summary
- remove the old exploded stats version directory after moving stats from a pre-release metadata version to the full release
- add regression coverage for the stale stats directory case
- keep the compatibility automation PR staging the stats updates

Fixes #2027

### DIFF
--- a/.github/workflows/verify-new-library-version-compatibility.yml
+++ b/.github/workflows/verify-new-library-version-compatibility.yml
@@ -378,7 +378,7 @@ jobs:
             git restore --source=HEAD --staged --worktree gradle/libs.versions.toml || true
 
             if git status --porcelain | grep -q .; then
-              git add -A -- metadata tests/src
+              git add -A -- metadata stats tests/src
               COMMIT_BODY_FILE="$(mktemp)"
               {
                 printf 'Update tested library versions\n\n'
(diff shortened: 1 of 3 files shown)


kimeta (approved):
Changes LGTM.

Issue #2037, opened 2026-04-21
Bug: checkMetadataFiles ignores shared-only supported versions from index.json


oleksii-tumanov:
### Describe the bug

`checkMetadataFiles` resolves coordinates strictly from physical metadata directories, so supported versions that reuse a shared metadata-version are skipped even though the repository already treats them as valid supported coordinates via `index.json`.

For example:
1. `org.slf4j:slf4j-api:2.0.17` is listed as a supported tested version in `metadata/org.slf4j/slf4j-api/index.json`, 
2. the filesystem only contains `1.5.3` and `1.7.36`.
3. Because there is no physical `metadata/org.slf4j/slf4j-api/2.0.17/` directory, strict resolution drops the coordinate before validation starts and `checkMetadataFiles` reports no matches.

### Steps to reproduce

1. Run `./gradlew listCoordinates -Pcoordinates=org.slf4j:slf4j-api:2.0.17 --quiet`.
2. Observe it returns `org.slf4j:slf4j-api:2.0.17`.
3. Run `./gradlew checkMetadataFiles -Pcoordinates=org.slf4j:slf4j-api:2.0.17 --console=plain`.
4. Observe: `No matching coordinates found for metadata checks. Nothing to do.`

### Expected behavior

If a version is supported by artifact `index.json`, `checkMetadataFiles` should resolve the mapped metadata directory and validate that metadata instead of skipping the coordinate.

### Logs and error messages

```text
> Task :checkMetadataFiles
No matching coordinates found for metadata checks. Nothing to do.
```

### GraalVM version

OpenJDK 25.0.2+10-jvmci-25.1-b16

### Operating system

macOS

### CPU architecture

AARCH64 (ARM64)

oleksii-tumanov:
Based on the feedback on #2038 and #2041, it sounds like the current behavior is matching the design, I’m fine with closing this.

I suggest to following up separately with a docs change (developer/contributor doc). Distinction between shared-aware coordinate resolution and strict batched execution is not obvious.

PR #2038, opened 2026-04-21
fix: shared-version resolution during batching execution

oleksii-tumanov:
Relates to #2037

## Problem

The repository already supports library versions that reuse a shared metadata directory through `metadata/<group>/<artifact>/index.json`.

Execution batching was not fully aligned with that model. Some `k/n` batching paths still expanded work from physical metadata directories only, so shared-only supported versions were skipped from batched execution.

Example:
- `org.slf4j:slf4j-api:2.0.17` is a supported version in `metadata/org.slf4j/slf4j-api/index.json`
- it reuses metadata directory `1.7.36`
- strict directory-based batching can skip it because there is no physical `metadata/org.slf4j/slf4j-api/2.0.17/` directory

## What this PR changes

- makes `CoordinatesAwareTask` resolve fractional batches from `tck.getMatchingCoordinates("all")`
- makes `ComputeAndPullAllowedDockerImagesTask` use the same shared-version-aware coordinate set
- aligns root `testInfra` / `testAllInfra` batch expansion with the same coordinate model

## Why this matters

- `k/n` execution batching now matches the repository's supported-coordinate model
- shared-only supported versions are included in batched execution
- batched execution is consistent with single-coordinate execution paths that already honor `index.json`

## Out of scope

- changed-coordinate / `diffCoordinates` behavior remains strict

## Validation

- `./gradlew -p tests/tck-build-logic test --tests org.graalvm.internal.tck.harness.TckExtensionTests --tests org.graalvm.internal.tck.harness.tasks.CoordinatesAwareTaskTests --tests org.graalvm.internal.tck.harness.tasks.ComputeAndPullAllowedDockerImagesTaskTests --console=plain`
- `./gradlew help --quiet`
- sampled `./gradlew testInfra -Pcoordinate=1/64 -Pparallelism=1` and verified that the expanded reproducer list includes shared-only supported versions before interrupting the full run

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -145,7 +145,7 @@ List<String> resolveConcreteCoordinates(String coordinateFilter) {
     List<String> matchingCoordinates
     if (CoordinateUtils.isFractionalBatch(coordinateFilter)) {
         int[] fraction = CoordinateUtils.parseFraction(coordinateFilter)
-        List<String> allCoordinates = tck.getMatchingCoordinatesStrict("all")
+        List<String> allCoordinates = tck.getMatchingCoordinates("all")
         matchingCoordinates = CoordinateUtils.computeBatchedCoordinates(allCoordinates, fraction[0], fraction[1])
     } else {
         matchingCoordinates = tck.getMatchingCoordinates(coordinateFilter)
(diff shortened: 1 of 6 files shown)


oleksii-tumanov:
I opened 2 separate PRs related to the same issue because they have different flows and they are independent:
 
 - one flow is about building work lists and quietly skipping unsupported version  (#2038)
 - the other is about running metadata validation for a selected coordinate (#2041)
 
 
Let me know if you have other ideas.

jormundur00:
This is a non-issue. We explicitly don't want this functionality with batched/all task runs, as that would make any gate running `-Pcoordinates=all` or `-Pcoordinates=[k/n]` last indefinitely and timeout (such as `test-all-metadata.yml`), as there are already over 2400 `tested-versions` supported on the repository.

oleksii-tumanov:
Thanks for the clarification.

  I opened this to align batched resolution with the repository's shared-version model.

  Since that's not the intended behavior, I'll close this PR.

PR #2138, opened 2026-04-27
[GenAI] Add support for io.netty:netty-common:4.1.79.Final using gpt-5.4

### LINKED ISSUE #1412: Support for io.netty:netty-common:4.1.79.Final
### Full Maven coordinates

io.netty:netty-common:4.1.79.Final

### Library source code URL

N/A

### Link to the existing test suite

N/A

_This issue was created by automation._

kimeta:
## What does this PR do?

Fixes: #1412

This PR introduces tests and metadata for io.netty:netty-common:4.1.79.Final, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.4
- Agent: pi
- Model: gpt-5.4
- Input tokens: 1392896
- Cached input tokens: 22318976
- Output tokens: 206567
- Entries: 64
- Iterations: 34
- Library coverage percentage: 7.9
- Generated lines of code: 551
- Tested library lines of code: 877

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 31/46 covered calls (67.39%)
- Reflection: 26/40 covered calls (65.00%)
- Resources: 5/6 covered calls (83.33%)

Library coverage:
- Instruction: 3579/47041 (7.61%)
- Line: 877/11108 (7.90%)
- Method: 216/3008 (7.18%)

### Post-Generation Intervention

- Stage: `metadata_fix_failed`

- Intervention file: `post-gen-interventions/io.netty_netty-common_4.1.79.Final.md`

# Post-generation intervention report

Library: io.netty:netty-common:4.1.79.Final
Stage: `metadata_fix_failed`

## Summary

The Codex metadata-fix log and the Gradle failure output both show that the remaining failures were **not missing-metadata failures**. The final failing run had **no `MissingReflectionRegistrationError`** entries. Instead, the generated suite still contained several tests that depend on runtime class definition, jar/file code-source discovery, or classpath URL behavior that does not hold inside a GraalVM native test image.

Because these are native-image/test-environment limitations rather than missing reachability metadata, I removed the failing generated tests and kept the rest of the generated support intact.

### Removed whole test files

- `CleanerJava6Test`
  - Failure: `Unsupported netty-common location: resource:/io/netty/util/internal/PlatformDependent.class`
  - Root cause: the test assumes the library is exposed as a `jar:` or `file:` URL so it can create an isolated `URLClassLoader`. In the native image run it is exposed through a native-image resource URL instead. That is a packaging/runtime-environment mismatch, not missing metadata.

- `NativeLibraryLoader$1Test`
  - Failure: assertion `Expected helper class lookup to be forced through defineClass`
  - Root cause: the test tries to force Netty through a helper-definition path using a custom class loader and then asserts a very specific fallback path. In the native image runtime that helper lookup path is not exercised the same way, so the assertion fails even though no metadata registration error occurs.

- `NativeLibraryLoader$2Test`
  - Failure: assertion `Expected helper lookup to fall back to defineClass`
  - Root cause: same class-loader-path assumption as above. The test is validating a runtime class definition/fallback behavior, not metadata reachability.

- `PlatformDependent0$6Test`
  - Failure: assertion `Expected netty-common to have a code source`
  - Root cause: `ProtectionDomain.getCodeSource()` is `null` in the native test image, so the test cannot build the isolated loader it expects. This is a native-image packaging/runtime characteristic, not missing metadata.

- `PlatformDependent0$7Test`
  - Failure: assertion `Expected netty-common to have a code source`
  - Root cause: same as `PlatformDependent0$6Test`; the test depends on a jar/file code source that is unavailable in the native image.

### Removed only the failing methods from mixed test files

- `ResourcesUtilTest#getFileFallsBackToTheOriginalEncodedPathWhenDecodingFails`
  - Failure: `UnsupportedFeatureError: Classes cannot be defined at runtime by default ... Tried to define class 'io.netty.util.internal.ResourcesUtil'`
  - Root cause: the test patches bytecode and calls `defineClass` from a custom loader at runtime. GraalVM native image forbids runtime class definition by default. This is an unsupported native-image feature for the test technique, not a metadata problem.

- `UnsafeAccessTest#fallsBackToThePrivateUnsafeConstructorWhenTheSingletonFieldCannotBeCast`
  - Failure: assertion `Expected netty-common on the test classpath`
  - Root cause: the test expects to locate `UnsafeAccess.class` as a normal classpath resource and reopen it through a fresh `URLClassLoader`. That resource lookup assumption does not hold in the native-image test layout. Again, this is a test/classpath modeling issue, not missing metadata.

## Why the remaining generated support should be preserved

The remaining generated support is still useful and should stay because:

- the final failure set no longer pointed to missing reachability metadata;
- after removing the non-metadata tests, the targeted native test run for `io.netty:netty-common:4.1.79.Final` passed successfully;
- the preserved tests still exercise meaningful Netty behaviors in native image, including cleaner access, native library loading fallback, reference counting, leak detector construction, resource lookup, generic type handling, unsafe offset access, and version metadata loading.

In short, Codex appears to have fixed the metadata-related part well enough for the library, and the remaining breakage came from generated tests that rely on unsupported or unavailable native-image runtime mechanics rather than from missing metadata entries.


vjovanov (approved):
Scoped to io.netty:netty-common:4.1.79.Final with library-specific tests in a separate package, no native-image skip or exact-version test issues, and CI is green for metadata validation plus current-defaults and future-defaults-all. I found no blocking issues.

vjovanov (approved):
Reviewed the diff, test shape, and CI for io.netty:netty-common:4.1.79.Final. The PR stays scoped to one library, the tests are library-specific and run under native image, and I found no blocking issues.

vjovanov (changes requested):
Requesting changes:  registers , but the actual nested test class is . That binary-name mismatch targets a non-existent type, so the custom leak-detector reflection config needs to be corrected before this can be accepted.

vjovanov (changes requested):
Requesting changes: tests/src/io.netty/netty-common/4.1.79.Final/src/test/resources/META-INF/native-image/reachability-metadata.json registers io_netty.netty_common.ResourceLeakDetectorFactory$DefaultResourceLeakDetectorFactoryTest$CustomResourceLeakDetector, but the actual nested test class is io_netty.netty_common.ResourceLeakDetectorFactory_DefaultResourceLeakDetectorFactoryTest$CustomResourceLeakDetector. That binary-name mismatch targets a non-existent type, so the custom leak-detector reflection config needs to be corrected before this can be accepted.

PR #4326, opened 2026-05-02
Fix empty native trace metadata progress detection

### LINKED ISSUE #4325: Forge native verification loops on empty trace metadata
### Problem

The native-test verification gate treats any trace metadata file as progress, even when GraalVM writes an empty `reachability-metadata.json` containing only `{}`.

In a failing run for `org.eclipse.jetty:jetty-security:9.3.19.v20170502`, the binary exited with `ExitStatus.MISSING_METADATA` (`172`) while every cycle produced only an empty trace metadata file. The gate appended each empty cycle directory to `metadataConfigDirs` and repeated until the iteration budget was exhausted.

### Expected behavior

A 172 run with no usable metadata entries should fail fast and surface the missing-registration stacktrace instead of feeding an empty trace directory into the next cycle.

### Notes

The observed GraalVM wiring was correct: the image was built with metadata tracing support and exact missing-registration reporting. The issue was gate-side progress detection.

vjovanov:
### Summary

- Treat trace metadata files with no entries, such as `{}`, as no usable progress in the native-test verification gate.
- Avoid appending empty trace directories to `metadataConfigDirs` after a `172` missing-metadata exit.
- Print the last 300 lines of the `runNativeTraceImage` Gradle log when the gate fails fast so the native failure is visible in stage output.
- Add regression tests for the empty `reachability-metadata.json` case and the 300-line failure log tail.

### Testing

- `python3 -m unittest tests.test_native_test_verification`
- `python3 -m py_compile utility_scripts/native_test_verification.py tests/test_native_test_verification.py`
- `git diff --check`

### Open question

Should a follow-up widen `traceMetadataConditionPackages` for harness packages, or parse missing-registration diagnostics directly? This PR only fixes the gate-side progress detection so empty trace output fails fast instead of looping.

Closes #4325

### DIFF
--- a/forge/tests/test_native_test_verification.py
+++ b/forge/tests/test_native_test_verification.py
@@ -345,8 +345,52 @@ class GateRoutingTests(unittest.TestCase):
         self.assertEqual(result.status, ntv.STATUS_FAILED)
         self.assertEqual(result.iterations_used, 1)
         printed = output.getvalue()
-        self.assertIn("produced no trace metadata; failing fast", printed)
+        self.assertIn("produced no usable trace metadata; failing fast", printed)
         self.assertIn("com.example.MissingThingException: boom", printed)
+
+    def test_fails_fast_when_172_produces_empty_metadata_json(self) -> None:
+        def _fake_run(cmd, **kwargs):  # type: ignore[no-untyped-def]
+            stdout = kwargs.get("stdout")
+            if hasattr(stdout, "write"):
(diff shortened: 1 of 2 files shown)


kimeta (approved):

vjovanov:
Superseded by #4327. GitHub kept this PR pinned to the old head commit after the branch update, so I opened a fresh PR with the current fix.

PR #5170, opened 2026-05-03
Derive native trace packages from user-code-filter

### LINKED ISSUE #5169: Derive metadata package lists from library JARs
## Problem

Native metadata tracing can fail when Forge passes Maven group IDs as trace condition packages. For example, `commons-beanutils` is a Maven group ID, not a Java package prefix, and GraalVM rejects it for `TraceMetadataConditionPackages`.

## Required behavior

- Derive package lists from the resolved, non-transitive library JAR.
- Use the complete sorted list of Java packages from the JAR for `allowed-packages`.
- Use the same JAR-derived package list for `TraceMetadataConditionPackages`.
- Do not fall back to Maven group IDs, collapsed package roots, manually supplied package lists, or packages inferred from checker errors.
- Fail if the resolved JAR contains no Java packages.

## Validation

- `tests/tck-build-logic` tests cover full package derivation and no group-id fallback.
- Native trace callers no longer pass `-PtraceMetadataConditionPackages`.

vjovanov:
## What does this PR do?

Fixes #5169.

This makes the resolved library JAR the single source of truth for metadata package lists:

- derives the complete sorted Java package list from the non-transitive target JAR
- writes that list to `allowed-packages` during scaffold, generateMetadata, contribution, version update, and finalization flows
- derives `TraceMetadataConditionPackages` inside Gradle instead of accepting `-PtraceMetadataConditionPackages` from Forge callers
- removes the `checkMetadataFiles` allowed-packages auto-repair path so packages are not inferred from checker output
- documents the invariant in both functional specs and workflow docs

Validation run:

- `./gradlew -p tests/tck-build-logic test --stacktrace`
- `python3 -m unittest forge.tests.test_native_test_verification`
- `python3 -m py_compile forge/utility_scripts/native_test_verification.py forge/utility_scripts/native_metadata_exploration.py forge/utility_scripts/library_finalization.py forge/ai_workflows/workflow_strategies/workflow_strategy.py`
- `git diff --check`
- `./gradlew runNativeTraceImage -Pcoordinates=commons-cli:commons-cli:1.5.0 -PtraceMetadataPath=/tmp/forge-trace-commons-cli-1.5.0 --stacktrace` completed without the previous trace package parser failure

- `tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/utils/MetadataGenerationUtils.java`: resolves the target coordinate JAR through Gradle dependency resolution and scans `.class` entries.
- `tests/tck-build-logic/src/main/groovy/org.graalvm.internal.tck.gradle`: derives runtime trace condition packages from the resolved JAR before launching the trace-enabled native image.
- `forge/utility_scripts/library_finalization.py`: invokes `./gradlew updateAllowedPackages` during finalization before `checkMetadataFiles`.

### DIFF
--- a/forge/utility_scripts/native_metadata_exploration.py
+++ b/forge/utility_scripts/native_metadata_exploration.py
@@ -78,7 +78,6 @@ def run_native_metadata_exploration(
         reachability_repo_path: str,
         coordinate: str,
         output_dir: str,
-        condition_packages: list[str] | None = None,
         max_iterations: int = 5,
         step_timeout_seconds: int = DEFAULT_TRACE_STEP_TIMEOUT_SECONDS,
 ) -> NativeExplorationResult:
@@ -90,10 +89,6 @@ def run_native_metadata_exploration(
         raise ValueError("max_iterations must be >= 1")
     if not os.path.isabs(output_dir):
         raise ValueError("output_dir must be an absolute path")
-
-    group = coordinate.split(":", 1)[0]
-    packages = list(condition_packages) if condition_packages else [group]
-    condition_packages_arg = ",".join(packages)
 
     runs_dir = os.path.normpath(os.path.join(output_dir, "..", "runs"))
     _reset_directory(output_dir)
@@ -149,7 +144,6 @@ def run_native_metadata_exploration(
             "runNativeTraceImage",
             f"-Pcoordinates={coordinate}",
             f"-PtraceMetadataPath={run_i}",
-            f"-PtraceMetadataConditionPackages={condition_packages_arg}",
         ]
         run_log = _new_log_path(coordinate, f"run-iter-{i}")
         run_log_paths.append(run_log)
--- a/forge/utility_scripts/native_test_verification.py
+++ b/forge/utility_scripts/native_test_verification.py
@@ -94,7 +94,6 @@ def verify_native_test_passes(
         reachability_repo_path: str,
         coordinate: str,
         output_dir: str,
-        condition_packages: list[str] | None = None,
         max_iterations: int = 100,
         cycle_timeout_seconds: int = DEFAULT_CYCLE_TIMEOUT_SECONDS,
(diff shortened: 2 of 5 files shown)


jormundur00 (approved):

tests/tck-build-logic/src/test/java/org/graalvm/internal/tck/GenerateMetadataTaskTests.java
```diff
@@ -63,7 +64,7 @@ void runWithFromJarCreatesFilterFromJarRootsAndAddsAgentBlock() throws IOExcepti
     }
 
     @Test
-    void runWithFromJarFallsBackToGroupIdWhenJarHasNoPackages() throws IOException {
+    void runWithFromJarFailsWhenJarHasNoPackages() throws IOException {
```
kimeta: Changes related to resolving `allowed-packages` and determining which packages are included in `user-
code-filter.json` should not be part of this PR. Those areas are probably not the source of the package-related failure, and changing them makes the fix broader than necessary.

The failure is likely caused only by the native trace filter, which was hardcoded to use the group ID
instead of the actual packages found in the JAR. This PR should focus only on that issue: deriving the
native trace filter packages from the existing `user-code-filter.json` and leaving the existing `allowed-packages` and `user-code-filter.json` behavior unchanged.

If the JAR has no packages, it is probably only a linking library, so dynamic-access workflow will not be invoked.

vjovanov: Addressed. I narrowed the PR to the native trace path only: callers no longer pass `-PtraceMetadataConditionPackages`, `runNativeTraceImage` now derives `TraceMetadataConditionPackages` from the existing `user-code-filter.json`, and the `allowed-packages` / `user-code-filter.json` generation behavior is back to `master`.

kimeta (approved):

PR #5925, opened 2026-05-05
Fix per-class native metadata aggregation

### LINKED ISSUE #5924: Fix per-class native metadata aggregation
The dynamic-access native verification flow should keep metadata cumulative after each class. The metadata agent runs once, native tests run with the durable metadata, native tracing only runs when native tests still fail, and any traced metadata must be folded back into metadata/<group>/<artifact>/<version>/ before the next class starts.

Changes needed:
- aggregate successful native-trace output into durable library metadata
- commit durable metadata after each successful per-class gate
- allow metadataConfigDirs to reach nativeTest/nativeTestCompile, not only native trace builds
- cover the routing and aggregation behavior with focused tests

Validation:
- PYTHONPATH=forge python3 -m unittest forge.tests.test_native_test_verification
- ./gradlew :tck-build-logic:test --stacktrace

vjovanov:
Closes #5924

## Summary
- aggregate successful native-trace output into durable library metadata before the next class starts
- commit durable metadata after each successful per-class native-test gate
- pass metadataConfigDirs through normal nativeTest/nativeTestCompile invocations
- document the cumulative metadata behavior and add focused unit coverage

## Validation
- PYTHONPATH=forge python3 -m unittest forge.tests.test_native_test_verification
- ./gradlew :tck-build-logic:test --stacktrace

### DIFF
--- a/forge/ai_workflows/workflow_strategies/dynamic_access_iterative_strategy.py
+++ b/forge/ai_workflows/workflow_strategies/dynamic_access_iterative_strategy.py
@@ -19,7 +19,7 @@ from utility_scripts.dynamic_access_report import (
     load_dynamic_access_coverage_report,
 )
 from utility_scripts.large_library_progress import LargeLibraryProgressState
-from utility_scripts.metadata_index import resolve_test_version
+from utility_scripts.metadata_index import resolve_metadata_version, resolve_test_version
 from utility_scripts.native_test_verification import (
     STATUS_FAILED as NATIVE_TEST_GATE_FAILED,
     per_class_output_dir,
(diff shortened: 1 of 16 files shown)


jormundur00 (approved):

jormundur00 (commented):
Deep review notes from the original revision:

1. The trace aggregation path used the raw coordinate version instead of the index-resolved `metadata-version`. For tested-version aliases, Gradle consumes `metadata/<group>/<artifact>/<metadata-version>`, while the original aggregation and per-class metadata commit would write/stage `metadata/<group>/<artifact>/<tested-version>`, so the next class and CI would not see the accumulated metadata.

2. `metadataConfigDirs` was forwarded to `nativeTest`, and the subproject build learned to honor it for `nativeTestCompile`, but the top-level `nativeTestCompile` invocation task still dropped the property before invoking the per-coordinate subproject build. That left the issue's stated `nativeTestCompile` scope incomplete.

I pushed follow-up commit `b41691894` to this PR branch to address both: aggregation/staging now resolves the metadata version through `index.json`, and `NativeTestCompileInvocationTask` forwards `metadataConfigDirs` with focused regression coverage. I only ran `git diff --check`; no tests were run locally.

Issue #3603, opened 2026-04-30
Support for io.ktor:ktor-client-core-jvm:3.2.0


jvukicev:
### Full Maven coordinates

io.ktor:ktor-client-core-jvm:3.2.0

### Reason for requesting the library

Popular Maven library rank 225 from `/home/jovan/mvnrepository-popular-scraper/popular-maven-libraries.json`.
Mirrored from fork issue https://github.com/jormundur00/graalvm-reachability-metadata/issues/3483 after strict `groupId:artifactId` metadata and open-issue filtering.

_This issue was created by automation._

vjovanov:
Human intervention needed

The automated job failed for `io.ktor:ktor-client-core-jvm:3.2.0`.

Preserved work branch: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fvjovanov%2Fhuman-intervention%2Fissue-3603-library-new-request-io.ktor-ktor-client-core-jvm-3.2.0-10220ae8

The failure appears to have happened in the `discover-artifact-metadata` stage before any metadata, tests, stats, or run metrics were generated for this coordinate. In `human-intervention-logs/io.ktor:ktor-client-core-jvm:3.2.0/discover-artifact-metadata/discover_artifact_metadata.log`, Gradle stops during project configuration with `Plugin [id: 'com.diffplug.spotless', version: '6.3.0'] was not found`, pointing at `build.gradle` line 19. The same log shows Gradle only searched `MavenLocal` and the Gradle Central Plugin Repository, then exited with `BUILD FAILED in 13s`. There is no generated `metadata/io.ktor/ktor-client-core-jvm`, `tests/src/io.ktor/ktor-client-core-jvm`, or `stats/io.ktor/ktor-client-core-jvm`, and no run-specific files were present under `forge/script_run_metrics` or `forge/benchmark_run_metrics`, which matches the unavailable dynamic-access coverage/metrics.

Please inspect:
- `human-intervention-logs/io.ktor:ktor-client-core-jvm:3.2.0/discover-artifact-metadata/discover_artifact_metadata.log`

Next steps:
1. Reproduce `discover-artifact-metadata` in an environment with working Gradle plugin resolution and confirm whether `com.diffplug.spotless:6.3.0` can be resolved there.
2. If plugin resolution is an environment issue, fix the repository/network/plugin-portal access and rerun the workflow for `io.ktor:ktor-client-core-jvm:3.2.0`.
[...]

Issue #2631, opened 2026-04-28
Support for org.specs2:specs2-core_3:4.20.5


jvukicev:
### Full Maven coordinates

org.specs2:specs2-core_3:4.20.5

_This issue was created by automation._

vjovanov:
Human intervention needed

The automated job failed for `org.specs2:specs2-core_3:4.20.5`.

Preserved work branch: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fvjovanov%2Fhuman-intervention%2Fissue-2631-library-new-request-org.specs2-specs2-core_3-4.20.5-d98b67c8

Although the workflow was marked `test_generation_failed`, the preserved evidence points to an earlier failure in `discover-artifact-metadata`, before any tests or metrics were generated. In `human-intervention-logs/org.specs2:specs2-core_3:4.20.5/discover-artifact-metadata/discover_artifact_metadata.log`, Gradle aborts during root project configuration with `Plugin [id: 'com.diffplug.spotless', version: '6.3.0'] was not found` and points to `build.gradle:19`. There is no preserved `tests/src/org.specs2/specs2-core_3/4.20.5/` or `stats/org.specs2/specs2-core_3/4.20.5/`, which matches the unavailable dynamic-access coverage.

Please inspect `human-intervention-logs/org.specs2:specs2-core_3:4.20.5/discover-artifact-metadata/discover_artifact_metadata.log` and `build.gradle:19`.

Next steps:
1. Reproduce the `discover-artifact-metadata` step in an environment where the Spotless plugin can resolve, or fix the plugin repository/version issue first.
2. Once Gradle configuration succeeds, rerun generation for `org.specs2:specs2-core_3:4.20.5` and confirm that `tests/src/org.specs2/specs2-core_3/4.20.5/` and `stats/org.specs2/specs2-core_3/4.20.5/` are created.
3. If generation still fails after the bootstrap issue is fixed, inspect the newly produced generation/test logs for the library-specific failure.

Issue #2166, opened 2026-04-27
Support for org.scalatest:scalatest_3:3.2.19


jvukicev:
### Full Maven coordinates

org.scalatest:scalatest_3:3.2.19

### Library source code URL

N/A

### Link to the existing test suite

N/A

_This issue was created by automation._

vjovanov:
Human intervention needed

The automated job failed. Preserved work is available here: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fvjovanov%2Fhuman-intervention%2Fissue-2166-library-new-request-org.scalatest-scalatest_3-3.2.19-982cd4af

The available artifacts suggest the failure is most likely not in Java test synthesis itself, but in artifact selection and/or Gradle setup during the preserved rerun. The requested coordinate is `org.scalatest:scalatest_3:3.2.19`, but the preserved worktree only contains generated outputs for `org.scalatest:scalatest-compatible:3.2.19` under `metadata/org.scalatest/scalatest-compatible/3.2.19/`, `tests/src/org.scalatest/scalatest-compatible/3.2.19/`, and `stats/org.scalatest/scalatest-compatible/3.2.19/`. `metadata/org.scalatest/scalatest-compatible/index.json` already lists `3.2.19` as a tested version, so the run appears to have selected the wrong module or a module that was already present.

The concrete preserved log at `human-intervention-logs/org.scalatest:scalatest_3:3.2.19/discover-artifact-metadata/discover_artifact_metadata.log` fails before any library-specific task runs: Gradle stops while evaluating the root `build.gradle` at line 19 because plugin `com.diffplug.spotless` version `6.3.0` could not be resolved from `mavenLocal` or the Gradle Plugin Portal. The generated metrics for the preserved artifacts are in `stats/org.scalatest/scalatest-compatible/3.2.19/execution-metrics.json` and show `status: success`, `generated_loc: 140`, `metadata_entries: 0`, and dynamic access `0/0`, which reinforces that generation completed for `scalatest-compatible`, not for `scalatest_3`.

Please inspect:
- `human-intervention-logs/org.scalatest:scalatest_3:3.2.19/discover-artifact-metadata/discover_artifact_metadata.log`
- `stats/org.scalatest/scalatest-compatible/3.2.19/execution-metrics.json`
- `tests/src/org.scalatest/scalatest-compatible/3.2.19/src/test/java/org_scalatest/scalatest_compatible/Scalatest_compatibleTest.java`
[...]

PR #6139, opened 2026-05-06
[Forge] Fix and harden Gradle bootstrap discovery

vjovanov:
## What does this PR do?

Forge issue runs stopped in `discover-artifact-metadata` while configuring the root Gradle build because `com.diffplug.spotless` could not be resolved from Maven Local or the Gradle plugin repositories. Each stop was charged to whichever library happened to be in flight, so one host-wide condition produced many unrelated `human-intervention` items.

**Root cause (found while validating this PR, and now fixed):** Forge isolates the Gradle *user home* but did not carry over host Gradle *configuration*. On a host that reaches Maven Central and the Gradle plugin repository only through an HTTP proxy, `~/.gradle/gradle.properties` carries the proxy settings — so an isolated home without it fails every plugin and distribution download **deterministically**, after a full connect timeout. Reproduced on a clean checkout: a cold isolated Forge Gradle home failed the Spotless plugin resolution after `BUILD FAILED in 6m 7s`; with the host `gradle.properties` present, the same discovery is `BUILD SUCCESSFUL in 35s`.

The PR therefore fixes the cause and keeps a safety net for the transient case:

- links the host `gradle.properties` into the isolated Gradle user home — linked rather than copied so host credentials are not duplicated into a shared temporary directory; an explicit `FORGE_GRADLE_USER_HOME` is taken as given and left alone. Inheritance is deliberately wholesale rather than filtered to proxy keys, so a Forge run behaves the way `./gradlew` behaves for a developer on that host; filtering would require writing a copy, which is precisely what would duplicate a host proxy password into a shared temp dir. The accepted cost is precedence: a Gradle user home `gradle.properties` outranks the repository's own, so a host that sets keys like `org.gradle.java.home` overrides repository intent for Forge runs. Hosts are expected to keep that file to host concerns such as proxy configuration.
- keys the Gradle user home on the checkout's common git directory (resolved from the filesystem the way `git rev-parse --git-common-dir` does, so building an environment never shells out), so all linked issue worktrees of a checkout share one plugin/dependency cache
- runs discovery through `./gradlew --no-daemon` and records structured per-attempt logs (attempt name, exit code, rendered command, output)
- detects Spotless plugin and Gradle wrapper distribution bootstrap failures and retries once with `--stacktrace --info --refresh-dependencies`
- raises `GradleBootstrapFailure` after the retry is exhausted. It is a **typed external failure**, so the existing external-failure boundary releases the claim with no `human-intervention` label, no comment, and no failed-work preservation; because the cause is host-wide, the run also stops after reverting still-active claims and exits with a dedicated status that the do-work loop retries after its normal sleep — the same shape as a rate-limit stop.

This is a rebase and redesign of the original May version of this PR onto current `master`, which has since grown a typed external-failure boundary and an explicit human-intervention classification policy. The stop now flows through that boundary and reports its own reason and exit code instead of masquerading as a Ctrl+C.

Spec: new `FS-shared-infrastructure-bootstrap-failure`, plus the third typed external boundary added to `FS-human-intervention-policy`.

## Validation

- `PYTHONPATH=forge python -m pytest forge/tests -q` → **363 passed**, 3 failures that are pre-existing on `master` (`test_library_update_passes_issue_requested_metadata_context_to_workflow`, `test_fixture_issue_listing_can_exclude_non_user_authors`, `test_validate_no_scaffold_placeholders_rejects_placeholder_text`). `master` itself reports 5 failures / 340 passed; two of those were `InterruptHandlingTests` cases that assumed a pre-existing `/tmp/reachability` directory, which this PR makes environment-independent.
- `grund check` → clean (only the two pre-existing `outdated grund init block` warnings on `AGENTS.md` and `forge/AGENTS.md`).
- Real Gradle runs, full failure path: initial `--no-daemon` attempt failed with the exact `Plugin [id: 'com.diffplug.spotless', version: '6.3.0'] was not found` error, was classified as a bootstrap failure, retried with `--stacktrace --info --refresh-dependencies`, and raised `GradleBootstrapFailure` with both attempts in one structured log.
- Real Gradle run, fixed path: after the `gradle.properties` link, `discover_artifact_metadata` for `org.postgresql:postgresql:42.7.3` succeeded through a cold isolated Forge Gradle home in one attempt.
- Filesystem common-dir resolution is asserted equal to `git rev-parse --git-common-dir` for a checkout, a linked worktree, and a nested subdirectory; also verified against this repo's real Forge worktrees.
- The `Connect timed out` wrapper-download message observed from a real failing bootstrap is kept as a test fixture — the original pattern only matched `Connection timed out` and missed it.

- `forge/utility_scripts/gradle_environment.py` reads `.git` / `commondir` pointer files to derive the Gradle cache identity, and symlinks the host `gradle.properties` into the isolated Gradle user home.
- `forge/utility_scripts/source_context.py` runs `./gradlew discoverArtifactMetadata` and may access Gradle plugin/dependency repositories during bootstrap and the diagnostic retry.

## Open question

- **Fixture E2E not completed.** Per `CLAUDE.md` I ran the hermetic fixture E2E (`--fixture-testing --issue-number 9101`). It got through fixture routing, isolated worktree setup, library preflight, and dispatch, then failed for two environmental reasons unrelated to this change: a leftover worktree from an earlier run still holds `ai/vjovanov/add-lib-support-com.google.http-client-google-http-client-1.42.2` (`git switch -C` → exit 128), and the agent workspace has hit its spend cap, so both the preflight and the failure analysis returned errors. A clean fixture E2E needs the stale worktree pruned and agent budget available. Worth noting the run did show the generic failure path still behaving correctly around the new `GradleBootstrapFailure` clause: the `CalledProcessError` was classified as logical and went through failed-work preservation.
- **Follow-up on the affected issues.** This PR intentionally does not remove `human-intervention` from the originally affected issues (#2166, #2631, #2740, #3603, #3689, #5400, #5434, #5467, #5471, #5474, #5476, #5478, #5480, #5483, #5490, #5493, #5505, #5507, #5510, #5511). The follow-up is to rerun representative ones after this lands and clear the label only where discovery now passes.

### DIFF
--- a/forge/do_up_to_date_work.sh
+++ b/forge/do_up_to_date_work.sh
@@ -29,6 +29,7 @@ REVIEW_MODEL="${FORGE_REVIEW_MODEL:-gpt-5.4}"
 USER_REQUESTED_ONLY="${FORGE_USER_REQUESTED_ISSUES_ONLY:-0}"
 WORK_STRATEGY_NAME="${FORGE_STRATEGY_NAME:-dynamic_access_main_sources_pi_gpt-5.6-sol}"
 GITHUB_RATE_LIMIT_EXIT_CODE=75
+GRADLE_BOOTSTRAP_EXIT_CODE=76
 MAX_PARALLELISM=4
 RUN_ONCE=0
 REQUEST_STOP=0
@@ -368,6 +369,13 @@ run_step() {
 
     if [[ "$status" -eq "$GITHUB_RATE_LIMIT_EXIT_CODE" ]]; then
         log "Skipping remaining work because the GitHub API limit is exhausted."
+        return 0
+    fi
+
+    # §FS-shared-infrastructure-bootstrap-failure: a shared Gradle bootstrap outage
+    # is host-wide, so skip the rest of this cycle and retry after the normal sleep.
+    if [[ "$status" -eq "$GRADLE_BOOTSTRAP_EXIT_CODE" ]]; then
+        log "Skipping remaining work because the shared Gradle bootstrap failed; retrying after sleep."
         return 0
     fi
 
--- a/forge/docs/functional-spec.md
+++ b/forge/docs/functional-spec.md
@@ -508,7 +508,14 @@ isolated Gradle user home keep bootstrap from failing in the first place:
   download deterministically, after a full connect timeout. The file is linked
   rather than copied so host credentials are not duplicated into a shared
   temporary directory. An explicit Gradle user home chosen by an operator is
-  taken as given and left alone.
+  taken as given and left alone. Inheritance is deliberately wholesale rather
+  than filtered to proxy keys: a Forge run should behave the way `./gradlew`
+  behaves for a developer on that host, and filtering would mean writing a copy,
+  which is exactly what would duplicate a host proxy password into a shared
+  temporary directory. The accepted cost is precedence — a Gradle user home
+  `gradle.properties` outranks the repository's own, so a host that sets keys
+  like `org.gradle.java.home` overrides repository intent for Forge runs. Hosts
+  are expected to keep that file to host concerns such as proxy configuration.
 - **One cache per checkout.** Forge keys the Gradle user home on the checkout's
   common git directory, so every linked issue worktree of one checkout shares a
   single plugin and dependency cache and resolves the root build's plugins once
--- a/forge/docs/functional-spec.md
+++ b/forge/docs/functional-spec.md
@@ -425,8 +425,10 @@ CI-equivalent verification (compile, test, native-test, and finalization gates),
 which includes the rare case of an agent timeout. A workflow failure is
 **external** — and must not get the label — only when it surfaces as a typed
 exception from the dependency boundary Forge itself crosses: GitHub (`gh`: rate
-limits and 5xx/network) as `GitHubError` / `GitHubRateLimitExceeded`, and remote
-git operations (push/pull/fetch/clone/ls-remote) as `GitTransportError`. Maven
+limits and 5xx/network) as `GitHubError` / `GitHubRateLimitExceeded`, remote
+git operations (push/pull/fetch/clone/ls-remote) as `GitTransportError`, and a
+Gradle build that never left configuration as `GradleBootstrapFailure`
+(§FS-shared-infrastructure-bootstrap-failure). Maven
 Central and Docker registry failures have no such boundary — Gradle owns them
 inside `./gradlew test` and they reach Forge only as an opaque CI-check `rc != 0`,
 indistinguishable from a real test failure once the in-workflow retries have run —
@@ -434,7 +436,8 @@ so they are not special-cased and fall through to the safe logical default. When
 failure is external, Forge takes no issue action: it applies no
 `human-intervention` label and posts no comment, and silently releases the issue
 claim (status back to `Todo`, assignees cleared) so the issue is retried later.
-Rate limits additionally stop the current run for retry after reset.
+Rate limits and shared bootstrap failures additionally stop the current run for
+a later retry.
 
 The label can appear on issues or pull requests. On an issue, it means Forge
 could not safely produce a PR-ready result and posted enough diagnostics for a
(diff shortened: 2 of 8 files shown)


jormundur00 (commented):
Findings:

1. **P2: Gradle bootstrap failures are reclassified as Ctrl+C at the top level.**  
   `forge/forge_metadata.py:5777` catches the `KeyboardInterrupt` raised from `GradleBootstrapFailure`, then unconditionally calls `mark_user_interrupt_requested()` unless a stop marker exists. That overwrites `INTERRUPT_REASON_GRADLE_BOOTSTRAP`, so the final output says `Run interrupted by Ctrl+C` and exits `130`. This loses the dedicated infrastructure-failure signal the PR is trying to introduce. Preserve an existing interrupt reason before defaulting to Ctrl+C, and handle `INTERRUPT_REASON_GRADLE_BOOTSTRAP` explicitly in this block.

2. **P2: Parallel runs can also overwrite the bootstrap reason before the main loop sees it.**  
   `forge/forge_metadata.py:4415` still calls `mark_user_interrupt_requested()` unconditionally when `handle_completed_run` catches `KeyboardInterrupt`. With `FORGE_PARALLELISM > 1`, one issue can set `INTERRUPT_REASON_GRADLE_BOOTSTRAP` while another worker observes the interrupt and resets the reason to Ctrl+C. The main loop then reverts remaining active issues with the wrong reason. This should use the same guard pattern added elsewhere: only mark Ctrl+C if no interrupt reason is already set.

I did not run tests manually.

jvukicev (dismissed):
Requesting changes for the two P2 interrupt-reason handling issues noted in my review comment. The PR should preserve `INTERRUPT_REASON_GRADLE_BOOTSTRAP` through top-level and parallel-worker unwind paths instead of reclassifying it as Ctrl+C.

kimeta (approved):
Non-blocking: in the rarely used parallel mode, two workers could theoretically race and overwrite the recorded stop reason. Since this only affects diagnostic reporting and not the core Gradle bootstrap recovery, I don’t think it needs to block this PR.

vjovanov:
Thanks — both P2 findings are addressed. The branch has been rebased and reworked onto current `master` since your review, so the line numbers have moved; head is now `c2ff38ff26`.

**Finding 1 — bootstrap reason reclassified as Ctrl+C at the top level.** Fixed. `main()` now calls `preserve_user_interrupt_reason()` (`forge_metadata.py:8786`) instead of unconditionally resetting the reason, and handles the bootstrap case explicitly at `forge_metadata.py:8793`, reporting the infrastructure failure and exiting with a dedicated `GRADLE_BOOTSTRAP_EXIT_CODE` (76) rather than `130`. `do_up_to_date_work.sh` recognizes 76 and retries after its normal sleep, mirroring the rate-limit stop.

**Finding 2 — parallel runs overwrite the reason before the main loop sees it.** Fixed. The `handle_completed_run` interrupt path now uses the same guard (`forge_metadata.py:7235`).

**Third instance of the same race, from applying your principle uniformly.** `invoke_pipeline` had the same unguarded reset when a workflow driver returned an interrupt exit code, so with parallelism > 1 one worker could still relabel another issue's bootstrap stop as Ctrl+C. Also guarded now (`forge_metadata.py:5489`). The shared helper is `preserve_user_interrupt_reason` (`forge_metadata.py:552`); a genuine SIGINT still wins, because the signal handler records Ctrl+C unconditionally.

Regression tests added in `InterruptHandlingTests`: the parallel-race invariant (a recorded bootstrap stop survives a driver returning exit 130), that a plain Ctrl+C is still recorded when no reason was set, that the bootstrap failure is classified as external, and that it reverts the claim without human-intervention follow-up and stops the queue.

**Scope also grew since your review — worth a fresh look.** Validating the retry path surfaced the actual root cause, which is not transient: Forge isolates the Gradle user home but was discarding host Gradle configuration. On a host that reaches Maven Central and the Gradle plugin repository only through an HTTP proxy, `~/.gradle/gradle.properties` carries the proxy settings, so an isolated home without it fails plugin resolution deterministically after a full connect timeout — a cold Forge Gradle home failed the Spotless resolution in `BUILD FAILED in 6m 7s`, and the same discovery is `BUILD SUCCESSFUL in 35s` once the file is linked in. The PR now fixes that and keeps the retry as a safety net. There is an open question in the PR description about `gradle.properties` precedence that I would value your opinion on.

Two caveats I have not resolved: the hermetic fixture E2E did not complete (a leftover worktree holds the target branch, and the agent workspace hit its spend cap), and `human-intervention` is intentionally still on the originally affected issues pending reruns after this lands.

CI is green on `c2ff38ff26` (14 pass, 4 matrix jobs correctly skipped). Re-review would be appreciated.

kimeta (approved):
Re-reviewed the latest changes; they look good to me.

Issue #5322, opened 2026-05-03
Support for org.jetbrains.exposed:exposed-dao:1.2.0


vjovanov:
### Full Maven coordinates

org.jetbrains.exposed:exposed-dao:1.2.0

### Reason for requesting the library

Expand Kotlin library coverage toward the repository target of about 75 Kotlin and 75 Scala libraries.

_This issue was created from the Kotlin/Scala library coverage planning pass._

vjovanov:
Human intervention needed

The automated job failed for `org.jetbrains.exposed:exposed-dao:1.2.0`. Preserved work branch: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fvjovanov%2Fhuman-intervention%2Fissue-5322-library-new-request-org.jetbrains.exposed-exposed-dao-1.2.0-7dda76b1

The failure appears to be in the `discover-artifact-metadata` stage, before test generation started. In `human-intervention-logs/org.jetbrains.exposed:exposed-dao:1.2.0/discover-artifact-metadata/discover_artifact_metadata.log`, Gradle reports `:discoverArtifactMetadata FAILED` with `UnrecognizedPropertyException: Unrecognized field "metadata-version"` while reading `build/discovered-artifact-metadata/org.jetbrains.exposed-exposed-dao-1.2.0.json`. The agent trace in `build/agent-artifact-discovery-logs/org.jetbrains.exposed-exposed-dao-1.2.0.log` shows that the discovery step added that unsupported field, so the pipeline stopped before producing PR-ready metadata, generated tests, or dynamic-access stats. Dynamic-access coverage is therefore unavailable, and there are no generated files under `metadata/org.jetbrains.exposed`, `tests/src/org.jetbrains.exposed`, or `stats/org.jetbrains.exposed` in the preserved worktree.

Please inspect:
- `human-intervention-logs/org.jetbrains.exposed:exposed-dao:1.2.0/discover-artifact-metadata/discover_artifact_metadata.log`
- `build/agent-artifact-discovery-logs/org.jetbrains.exposed-exposed-dao-1.2.0.log`

Recommended manual follow-up:
- Fix the discovered artifact metadata output for this coordinate so it matches `DiscoveredArtifactMetadata` schema, then rerun the discovery/generation flow.
- Review the preserved branch contents, especially `build/discovered-artifact-metadata/org.jetbrains.exposed-exposed-dao-1.2.0.json`, and salvage any correct URL/description findings.
[...]

Issue #4865, opened 2026-05-03
Support for com.softwaremill.sttp.tapir:tapir-json-circe_3:1.11.25


vjovanov:
### Full Maven coordinates

com.softwaremill.sttp.tapir:tapir-json-circe_3:1.11.25

_This issue was created by automation._

vjovanov:
Human intervention needed

The automated job failed for `com.softwaremill.sttp.tapir:tapir-json-circe_3:1.11.25`. Preserved work branch: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fvjovanov%2Fhuman-intervention%2Fissue-4865-library-new-request-com.softwaremill.sttp.tapir-tapir-json-circe_3-1.11.25-0019687f

The most likely failing stage was artifact metadata discovery, before test generation or stats collection. In `human-intervention-logs/com.softwaremill.sttp.tapir:tapir-json-circe_3:1.11.25/discover-artifact-metadata/discover_artifact_metadata.log`, Gradle reports `> Task :discoverArtifactMetadata FAILED`, followed by `Execution failed for task ':discoverArtifactMetadata'` and `UnrecognizedPropertyException: Unrecognized field "metadata-version" (class org.graalvm.internal.tck.model.DiscoveredArtifactMetadata)` at the end of the log. The preserved branch only contains this discovery log, and there are no generated `metadata/`, `tests/src/`, or `stats/` files for this coordinate in the worktree, which matches the unavailable metrics and dynamic-access coverage.

Please inspect:
- `human-intervention-logs/com.softwaremill.sttp.tapir:tapir-json-circe_3:1.11.25/discover-artifact-metadata/discover_artifact_metadata.log`
- `build/agent-artifact-discovery-logs/com.softwaremill.sttp.tapir-tapir-json-circe_3-1.11.25.log`

Manual next steps:
1. Fix the discovery output/schema mismatch that introduced `metadata-version` into `DiscoveredArtifactMetadata`, then rerun `discoverArtifactMetadata` for `com.softwaremill.sttp.tapir:tapir-json-circe_3:1.11.25`.
2. After discovery succeeds, regenerate the coordinate’s library-new-request artifacts and confirm files are created under `metadata/`, `tests/src/`, and `stats/`.
3. Rerun the full library workflow to produce metrics and dynamic-access coverage for review.

vjovanov:
Forge fix published in #6141.

[...]

PR #6141, opened 2026-05-06
[Forge] Handle discovery metadata-version drift

vjovanov:
## What does this PR do?

Root cause: the external artifact-discovery agent could write build-local discovery JSON with a top-level `"metadata-version"` field. That was plausible because the prompt said `entry metadata-version` and repository index examples contain `"metadata-version"`, but the build-local discovery file is read back through strict Jackson binding to `DiscoveredArtifactMetadata`, which does not model that field. The result was `UnrecognizedPropertyException` in `:discoverArtifactMetadata` before scaffold generation, metadata generation, tests, stats, or Native Image execution.

This PR fixes the discovery contract narrowly:

- clarifies that URL checks target the derived metadata index version, not a discovery-file field
- lists the allowed top-level discovery JSON keys and explicitly tells the agent not to add `"metadata-version"`
- accepts only the redundant `"metadata-version"` field when rereading `DiscoveredArtifactMetadata`, while keeping other unknown discovery-file properties strict
- adds a regression test that runs the post-agent `DiscoverArtifactMetadata.run()` path against an agent script that writes a discovery file containing `"metadata-version"`

GitHub human-intervention items addressed:

- Addresses #4865 for `com.softwaremill.sttp.tapir:tapir-json-circe_3:1.11.25`
- Addresses #5322 for `org.jetbrains.exposed:exposed-dao:1.2.0`

Validation:

- `git diff --check origin/master...HEAD` passed on the clean PR branch
- `./gradlew -p tests/tck-build-logic --no-daemon test --tests org.graalvm.internal.tck.harness.tasks.DiscoverArtifactMetadataTests --stacktrace` passed on the clean PR branch with `BUILD SUCCESSFUL in 9s`
- reviewed fix-pass validation also passed the full `tests/tck-build-logic` suite after the binding change
- reviewed Forge reruns passed `discoverArtifactMetadata` for both affected coordinates using `codex -a never exec -s danger-full-access`, and both reruns completed without the `Unrecognized field "metadata-version"` failure

The original issue labels are intentionally left for post-merge Forge rerun cleanup: after this PR lands, rerun the affected issues on the fixed Forge code and remove `human-intervention` only from items whose discovery reruns pass; later library-specific failures should be handled as narrower follow-up classes.

- `tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/DiscoverArtifactMetadata.java` writes and reads build-local discovery JSON and runs the configured external artifact-discovery agent command. That agent command can access the network depending on the Forge invocation.
- `tests/tck-build-logic/src/main/java/org/graalvm/internal/tck/model/DiscoveredArtifactMetadata.java` controls Jackson binding for the build-local discovery JSON file.

### DIFF
--- a/tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/DiscoverArtifactMetadata.java
+++ b/tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/DiscoverArtifactMetadata.java
@@ -161,7 +161,7 @@ public class DiscoverArtifactMetadata extends DefaultTask {
                 Mark "not-for-native-image": true only for certain cases such as Android-only AAR/tooling artifacts, Scala.js artifacts, Kotlin Native/JS/Wasm artifacts, or artifacts with no JVM class files.
                 If you set "not-for-native-image": true, also set "reason" to a short evidence-based explanation and set "replacement" when there is an obvious JVM replacement artifact.
                 If the artifact might still be usable from JVM code, leave "not-for-native-image", "reason", and "replacement" absent.
-                The sources URL, the test suite URL, and the documentation URL must render to the entry metadata-version "%s" when "$version$" is replaced with "%s".
+                The sources URL, the test suite URL, and the documentation URL must render to the metadata index version "%s" when "$version$" is replaced with "%s".
                 The source, test suite, and documentation URLs should point to the right tag of the library.
                 If we have these source of these artifacts on maven, that should be prefered.
                 If there are tests on maven that should also be prefered.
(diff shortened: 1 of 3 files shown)


jormundur00 (approved):
No blocking findings.

I reviewed the local diff and surrounding consumers of `DiscoveredArtifactMetadata`: `DiscoverArtifactMetadata.run()`, `ArtifactMetadataDiscoveryUtils.readDiscoveryFile()`, `ScaffoldTask.loadDiscoveredArtifactMetadata()`, and the Forge path that reads `not-for-native-image` from the discovery file. The change is narrow: Jackson still fails on unrelated unknown fields, while only `"metadata-version"` is ignored.

One small non-blocking test gap: the PR claims other unknown properties remain strict, but the new regression only proves `"metadata-version"` is accepted. Adding a negative case with another unknown key would lock that behavior down.

I did not run any tests manually.

PR #6237, opened 2026-05-06
[GenAI] Test fix for org.jetbrains.exposed:exposed-core:1.0.0 using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #6220

This PR provides test fixes and new metadata for org.jetbrains.exposed:exposed-core:1.0.0, addressing compile java failures caused by changes in the updated library version.

Summary:
- Strategy: javac_iterative_with_coverage_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 32989
- Cached input tokens: 211968
- Output tokens: 3013
- Metadata entries: 5
- Iterations: 1
- Library coverage percentage: 8.73
- Previous library version metadata entries: 5
- Previous library version coverage percentage: 9.69

### Metadata Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `f242aa2b0f61de3a6f0f8706e1da0826d5cf33ec`

#### Dynamic access coverage

- org.jetbrains.exposed:exposed-core:1.0.0-beta-4: 0/0 covered calls (100.00%)
- org.jetbrains.exposed:exposed-core:1.0.0: 0/0 covered calls (100.00%)

#### Library coverage

**Instruction:**
- org.jetbrains.exposed:exposed-core:1.0.0-beta-4: 4466/54706 (8.16%)
- org.jetbrains.exposed:exposed-core:1.0.0: 4336/58766 (7.38%)

**Line:**
- org.jetbrains.exposed:exposed-core:1.0.0-beta-4: 710/7325 (9.69%)
- org.jetbrains.exposed:exposed-core:1.0.0: 699/8004 (8.73%)

**Method:**
- org.jetbrains.exposed:exposed-core:1.0.0-beta-4: 339/3066 (11.06%)
- org.jetbrains.exposed:exposed-core:1.0.0: 322/3421 (9.41%)

**Comparison between existing test version and AI-Generated update**

```diff

```

### Local CI Verification

- Status: `success`
- Commands run: 15
- Fixup attempts: 0

Issue #6220, opened 2026-05-06
[Automation] javac compile fails for org.jetbrains.exposed:exposed-core:1.0.0


graalvmbot:
## javac compile on ubuntu-latest / JDK 25 / mode current-defaults

Failure kind: javac compile
OS: ubuntu-latest
JDK: 25
Native-image mode: current-defaults
Reproducer: `GVM_TCK_NATIVE_IMAGE_MODE="current-defaults" GVM_TCK_LV="1.0.0" ./gradlew clean compileTestJava -Pcoordinates="org.jetbrains.exposed:exposed-core:1.0.0-beta-4"`
Runner log: https://github.com/oracle/graalvm-reachability-metadata/actions/runs/25454136977/job/74679302936
Last 300 lines of the log (excerpt truncated to fit the GitHub issue size limit):
```console
t.kt:165:13 'fun <T, S : T?> coalesce(expr: ExpressionWithColumnType<S>, alternate: Expression<out T>, vararg others: Expression<out T>): Coalesce<T, S>' is deprecated. This interface method will be removed following release 1.0.0 and should be replaced with the equivalent top-level function.
e: file:///home/runner/work/graalvm-reachability-metadata/graalvm-reachability-metadata/tests/src/org.jetbrains.exposed/exposed-core/1.0.0-beta-4/src/test/kotlin/org_jetbrains_exposed/exposed_core/Exposed_coreTest.kt:173:27 'object SqlExpressionBuilder : ISqlExpressionBuilder' is deprecated. This builder object will continue to be phased out following release 1.0.0. All expression builder methods previously restricted to this object have also been deprecated in favor of equivalent top-level functions, making this object useless as a receiver in any scope. It will no longer be necessary to import each individual method when used outside a scoped block, and on demand imports will now be possible via 'import org.jetbrains.exposed.v1.core.*', if required.
e: file:///home/runner/work/graalvm-reachability-metadata/graalvm-reachability-metadata/tests/src/org.jetbrains.exposed/exposed-core/1.0.0-beta-4/src/test/kotlin/org_jetbrains_exposed/exposed_core/Exposed_coreTest.kt:174:13 'fun rowNumber(): RowNumber' is deprecated. This interface method will be removed following release 1.0.0 and should be replaced with the equivalent top-level function.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileTestKotlin'.
[...]

PR #6246, opened 2026-05-07
[Forge] Promote newer javac metadata indexes

vjovanov:
## What does this PR do?

Promotes javac-fix metadata index entries when the target metadata version is parseably newer than the single current `latest` entry. This addresses the Forge human-intervention item in #6237 and the related javac update issue #6220 without editing the generated `org.jetbrains.exposed:exposed-core:1.0.0` metadata branch.

Root cause: the javac workflow always used `addLibraryMetadataIndexJson`, which intentionally preserves the existing `latest` marker. For #6237 that added `1.0.0` while leaving `1.0.0-beta-4` marked latest. Existing validators did not reject that stale latest marker.

The fix keeps the preserving task for historical backfills and unparseable versions, but switches javac-fix publication to `addLibraryAsLatestMetadataIndexJson` when Forge can confidently compare the target version and determine it is newer. The comparison covers final-over-prerelease cases such as `1.0.0-beta-4` to `1.0.0`, newer patch versions, `.Final` / `.RELEASE`, and common prerelease qualifiers.

Validation:
- `PYTHONPATH=$PWD/forge python -m pytest forge/tests/test_java_fail_workflow.py` -> `9 passed`
- `PYTHONPATH=$PWD/forge python -m pytest forge/tests -k 'java_fail_workflow or metadata_index or latest'` -> `10 passed, 236 deselected`
- `git diff --check origin/master...HEAD` -> passed
- Latest Forge review: `runtime/reviews/fetch-prs-forge-index-latest-forge-javac-index-promotion-forge-review-2.md` -> no blocking findings, `Ready to publish: yes`

GitHub items addressed:
- #6237: generated javac-fix PR with stale latest marker; this system fix is the durable handoff.
- #6220: related javac compile update issue for `org.jetbrains.exposed:exposed-core:1.0.0`.

- `forge/utility_scripts/metadata_index.py` reads local `metadata/<group>/<artifact>/index.json` files.
- `forge/ai_workflows/java_fail_workflow.py` continues to invoke the existing local Gradle metadata-index task.
- No new network, docker, or external-service access is added.

### DIFF
--- a/forge/ai_workflows/improve_library_coverage.py
+++ b/forge/ai_workflows/improve_library_coverage.py
@@ -50,6 +50,7 @@ from utility_scripts.library_stats import stats_artifact_dir
 from utility_scripts.metadata_index import (
     MATCH_NEW_VERSION,
     LibraryUpdateTarget,
+    is_newer_than_latest_metadata_version,
     load_index_entries,
     resolve_library_update_target,
 )
@@ -471,6 +472,7 @@ def _new_index_entry_from_baseline(
         baseline_entry: dict[str, Any],
         requested_version: str,
         tested_versions: list[str] | None = None,
+        mark_latest: bool = True,
 ) -> dict[str, Any]:
     old_versions = {
         str(value)
@@ -483,7 +485,10 @@ def _new_index_entry_from_baseline(
     }
     new_entry = copy.deepcopy(baseline_entry)
     new_entry = _replace_version_in_value(new_entry, old_versions, requested_version)
-    new_entry["latest"] = True
+    if mark_latest:
+        new_entry["latest"] = True
+    else:
+        new_entry.pop("latest", None)
     new_entry["metadata-version"] = requested_version
     new_entry.pop("test-version", None)
     new_entry.pop("default-for", None)
@@ -572,17 +577,22 @@ def clone_library_update_support(
         tested_versions = baseline_entry.get("tested-versions")
         if isinstance(tested_versions, list):
             new_entry_tested_versions = [str(version) for version in tested_versions]
+    mark_new_entry_latest = (
+        baseline_metadata_version == requested_version
+        and baseline_entry.get("latest") is True
+    ) or is_newer_than_latest_metadata_version(repo_path, group, artifact, requested_version)
     updated_entries: list[dict[str, Any]] = []
     new_entry = _new_index_entry_from_baseline(
         baseline_entry,
         requested_version,
         new_entry_tested_versions,
+        mark_latest=mark_new_entry_latest,
     )
     for entry in entries:
         if not isinstance(entry, dict):
             continue
         entry_copy = copy.deepcopy(entry)
-        if entry_copy.get("latest") is True:
+        if mark_new_entry_latest and entry_copy.get("latest") is True:
             entry_copy.pop("latest", None)
         if entry_copy.get("metadata-version") == baseline_metadata_version:
             if baseline_metadata_version == requested_version:
--- a/forge/ai_workflows/java_fail_workflow.py
+++ b/forge/ai_workflows/java_fail_workflow.py
@@ -36,6 +36,7 @@ from utility_scripts.strategy_loader import require_strategy_by_name
 from utility_scripts.workflow_setup import resolve_graalvm_java_home, validate_repo_paths
 
 DEFAULT_MODEL_NAME = "oca/gpt5"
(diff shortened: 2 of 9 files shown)


jormundur00 (changes requested):
The direction looks right. The javac-fix workflow should promote a newly added final/newer metadata version instead of preserving a stale prerelease `latest` marker. The tests cover the important happy path (`1.0.0-beta-4 -> 1.0.0`), patch promotion, historical backfill, unparseable versions, and the java-run path.

**Blocking Finding**

The Forge comparator is stricter than the prerelease language used by `addTestedVersion`.

`TestedVersionUpdaterTask.VERSION_PATTERN` accepts prerelease-like versions with trailing classifier payloads, for example `3.0.0-M5-javax`, because it allows `(?:[-.].*)?` after the recognized prerelease token. The new Forge parser rejects anything after the consumed qualifier/number, so a version like `3.0.0-M5-javax` is treated as unparseable.

That does not reject PR creation, but it means javac-fix will fall back to `addLibraryMetadataIndexJson` and preserve the old `latest` marker. This can reproduce the same stale-latest class of bug for version shapes that our existing `addTestedVersion` task already recognizes as prereleases.

I’d align the Forge parser with `TestedVersionUpdaterTask.VERSION_PATTERN`, or explicitly document and test that this path is intentionally more conservative.

**Non-Blocking Suggestion**

Since this PR is motivated by stale latest markers, it would be useful to repair the currently stale parseable indexes in the same PR. I found these:

- `metadata/com.graphql-java/graphql-java-extended-validation/index.json`: `latest` is `19.1`, newer entry exists: `19.2`
- `metadata/org.codehaus.plexus/plexus-container-default/index.json`: `latest` is `1.0-alpha-7`, newer entry exists: `1.0-alpha-8`
- `metadata/org.jetbrains.exposed/exposed-core/index.json`: `latest` is `1.0.0-beta-4`, newer entries exist: `1.0.0`, `1.1.0`
- `metadata/redis.clients/jedis/index.json`: `latest` is `4.4.0`, newer entries exist: `5.0.0`, `5.0.1`
- `metadata/org.aesh/readline/index.json`: `latest` is `2.6`, newer entries exist: `3.0`, `3.1`
- `metadata/org.hibernate.orm/hibernate-core/index.json`: `latest` is `7.3.0.Final`, newer entry exists: `8.0.0.Alpha1`
- `metadata/jakarta.ws.rs/jakarta.ws.rs-api/index.json`: `latest` is `2.1.6`, newer entry exists: `3.0.0`

This system fix prevents some future javac additions from creating the issue, but existing stale indexes will not necessarily self-heal because `addLibraryAsLatestMetadataIndexJson` no-ops when the target metadata version already exists.

jvukicev:
Updated the PR branch with commit `1aa118191a` to address the blocking parser finding. The Forge parser now follows the Gradle updater shape for qualifier versions with trailing classifier payloads, so versions like `3.0.0-M5-javax` remain parseable. Added regression coverage for final-over-classifier-prerelease, newer classifier prerelease, and separated qualifier-number versions such as `1.0-alpha-8` over `1.0-alpha-7`.\n\nValidation run locally:\n- `PYTHONPATH=$PWD/forge /usr/bin/python3 -m unittest forge.tests.test_java_fail_workflow` -> 12 tests passed\n- `python3 -m py_compile forge/utility_scripts/metadata_index.py forge/ai_workflows/java_fail_workflow.py forge/tests/test_java_fail_workflow.py` -> passed\n- `git diff --check` -> passed\n\nI did not include the existing stale-index promotions from the non-blocking suggestion here; those should be handled in a separate issue/PR so this PR remains a logic-only Forge fix.

jvukicev:
Follow-up update pushed in `a83a93c35f`: the latest-selection logic is no longer javac-only. Both javac and java-run fix workflows now use the same version comparison to choose the index task: parseably newer target versions are added as latest, while historical/backfill target versions preserve the existing latest entry. This covers the java-run backfill class that caused older versions to take over `latest`.\n\nAdditional validation:\n- `PYTHONPATH=$PWD/forge /usr/bin/python3 -m unittest forge.tests.test_java_fail_workflow` -> 13 tests passed\n- `python3 -m py_compile forge/utility_scripts/metadata_index.py forge/ai_workflows/java_fail_workflow.py forge/tests/test_java_fail_workflow.py` -> passed\n- `git diff --check` -> passed

jormundur00 (approved):
Fixed the blocking findings and increased the scope of changes to fit all fix labels.

PR #6233, opened 2026-05-06
[GenAI] Add support for org.hibernate.common:hibernate-commons-annotations:5.1.2.Final using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #1521

This PR introduces tests and metadata for org.hibernate.common:hibernate-commons-annotations:5.1.2.Final, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 217995
- Cached input tokens: 2028544
- Output tokens: 24112
- Metadata entries: 6
- Test-only metadata entries: 1
- Iterations: 10
- Library coverage percentage: 39.31
- Generated lines of code: 300
- Tested library lines of code: 342

### Metadata/dynamic-access evidence

- Covered dynamic-access calls: 22
- Metadata entries: 6
- These counts are different dimensions: covered dynamic-access calls count observed call sites, while metadata entries count generated reachability-config items. Depending on the access type, a single metadata rule can cover multiple observed call sites, or no shipped rule may be required when the covered access does not target fixed library-owned metadata.
- Generated metadata rules:
  - when `org.hibernate.annotations.common.util.impl.LoggerFactory` is reached, make `org.hibernate.annotations.common.util.impl.Log_$logger.<init>(org.jboss.logging.Logger)` available for reflection
  - when `org.hibernate.annotations.common.util.impl.LoggerFactory` is reached, make `org.hibernate.annotations.common.util.impl.Log_$logger_en.<init>(org.jboss.logging.Logger)` available for reflection
  - when `org.hibernate.annotations.common.util.StandardClassLoaderDelegateImpl` is reached, make `org.hibernate.annotations.common.util.StandardClassLoaderDelegateImpl` available for reflection
- Use the call-site and metadata-rule lists together to review whether the observed dynamic-access paths are explained by the generated reachability metadata.

### Metadata Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `ee0b9ea4ac30c20819fd38ebb2bce47b599b4df1`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 22/22 covered calls (100.00%)
- Reflection: 22/22 covered calls (100.00%)

Library coverage:
- Instruction: 1520/4081 (37.25%)
- Line: 342/870 (39.31%)
- Method: 153/351 (43.59%)

[...]

PR #6228, opened 2026-05-06
[GenAI] Add support for org.wildfly.common:wildfly-common:1.5.0.Final using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #1558

This PR introduces tests and metadata for org.wildfly.common:wildfly-common:1.5.0.Final, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 280475
- Cached input tokens: 4997120
- Output tokens: 35505
- Metadata entries: 0
- Iterations: 10
- Library coverage percentage: 8.43
- Generated lines of code: 113
- Tested library lines of code: 486

### Metadata Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `42dd3895cca38d8398c1c4cbc28eb2cb5f4a0a74`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 0/8 covered calls (0.00%)
- Reflection: 0/8 covered calls (0.00%)

Library coverage:
- Instruction: 2229/28529 (7.81%)
- Line: 486/5764 (8.43%)
- Method: 101/1365 (7.40%)

### Local CI Verification

- Status: `success`
- Commands run: 19
- Fixup attempts: 1

PR #6210, opened 2026-05-06
[GenAI] Add support for tools.jackson.core:jackson-databind:3.0.0-rc2 using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #2157

This PR introduces tests and metadata for tools.jackson.core:jackson-databind:3.0.0-rc2, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 144510
- Cached input tokens: 1131008
- Output tokens: 18152
- Metadata entries: 0
- Iterations: 4
- Library coverage percentage: 24.07
- Generated lines of code: 491
- Tested library lines of code: 8296

### Metadata/dynamic-access evidence

- Covered dynamic-access calls: 16
- Metadata entries: 0
- These counts are different dimensions: covered dynamic-access calls count observed call sites, while metadata entries count generated reachability-config items. Depending on the access type, a single metadata rule can cover multiple observed call sites, or no shipped rule may be required when the covered access does not target fixed library-owned metadata.
- Use the call-site and metadata-rule lists together to review whether the observed dynamic-access paths are explained by the generated reachability metadata.

### Metadata Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `8786d48ef2736cd0978224e234226140c6deaccd`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 16/64 covered calls (25.00%)
- Reflection: 16/64 covered calls (25.00%)

Library coverage:
- Instruction: 32905/145465 (22.62%)
- Line: 8296/34466 (24.07%)
- Method: 1992/7943 (25.08%)

### Local CI Verification

- Status: `success`
- Commands run: 15
- Fixup attempts: 0

PR #6198, opened 2026-05-06
[GenAI] Add support for io.grpc:grpc-api:1.27.2 using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #1371

This PR introduces tests and metadata for io.grpc:grpc-api:1.27.2, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 28412
- Cached input tokens: 175104
- Output tokens: 4127
- Metadata entries: 0
- Test-only metadata entries: 1
- Iterations: 1
- Library coverage percentage: 0.28
- Generated lines of code: 39
- Tested library lines of code: 7

### Metadata/dynamic-access evidence

- Covered dynamic-access calls: 2
- Metadata entries: 0
- These counts are different dimensions: covered dynamic-access calls count observed call sites, while metadata entries count generated reachability-config items. Depending on the access type, a single metadata rule can cover multiple observed call sites, or no shipped rule may be required when the covered access does not target fixed library-owned metadata.
- Use the call-site and metadata-rule lists together to review whether the observed dynamic-access paths are explained by the generated reachability metadata.

### Metadata Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `6f2b25d4adc6fced3e9bd4c9df1f74664c861511`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 2/2 covered calls (100.00%)
- Reflection: 2/2 covered calls (100.00%)

Library coverage:
- Instruction: 37/10673 (0.35%)
- Line: 7/2514 (0.28%)
- Method: 3/983 (0.31%)

### Local CI Verification

- Status: `success`
- Commands run: 15
- Fixup attempts: 0

PR #6249, opened 2026-05-07
[Forge] Gate dynamic-access publication evidence

vjovanov:
## Root cause

Forge published generated `library-new-request` PRs when the workflow status was successful and line coverage was positive, but it did not treat `run_metrics["stats"]["dynamicAccess"]` as a publication-quality gate. That let PR #6228 publish even though committed stats reported `0/8 (0.00%)` dynamic-access coverage and the generated metadata was `{}`.

Forge also emitted generic dynamic-access/metadata mismatch text without durable reviewer evidence. PRs #6233, #6210, #6198, and the dynamic-access evidence side of #6177 showed cases where reviewers needed concrete covered call-site evidence plus shipped/test-only metadata-rule summaries to decide whether a mismatch was expected.

## Fix

- Adds shared dynamic-access publication thresholds to the new-library quality gate.
- Rejects new-library publication when dynamic-access coverage is at or below 20%, when covered dynamic-access calls exist with no shipped or test-only metadata entries, or when covered calls are disproportionate to metadata entries without reviewable call-site/rule evidence.
- Loads dynamic-access coverage and both shipped and test-only metadata summaries before PR creation, so unexplained mismatches fail before `gh pr create`.
- Expands the PR-body evidence section to include covered call sites, shipped metadata rules, and test-only metadata rules when mismatch evidence is reviewable.
- Reuses the same threshold for Forge human-intervention routing.

## GitHub items addressed

- Addresses PR #6228 as the clean `0/8` dynamic-access publication-gate failure.
- Addresses PRs #6233, #6210, and #6198 by preventing unexplained dynamic-access/metadata mismatches from reaching review without concrete evidence.
- Addresses the dynamic-access evidence side of PR #6177. The remaining #6177 dependency-stub/test-only classification risk is intentionally left gated for the separate follow-up noted by review.
- This is a Forge system fix only. It does not edit generated metadata, stats, or tests on the failed/generated branches.

## Validation

Passed on the clean publication branch rebased to current `origin/master`:

- `PYTHONPATH=forge python -m pytest forge/tests/test_make_pr_new_library_support.py forge/tests/test_metrics_paths.py -k 'quality or metrics or make_pr_new_library_support or dynamic_access or human_intervention'` -> `18 passed, 1 warning`
- `PYTHONPATH=forge python -m pytest forge -k 'quality or metrics or make_pr_new_library_support or dynamic_access or human_intervention'` -> `53 passed, 193 deselected, 1 warning`
- `git diff origin/master --check` -> passed

Reviewer validation from `runtime/reviews/fetch-prs-forge-coverage-metadata-evidence-forge-dynamic-access-publication-evidence-gate-forge-review-2.md` reported no blocking findings, no validation gaps, and `Ready to publish: yes`. The recorded smoke check showed #6233/#6228/#6210/#6198 blocked by this gate and #6177 not blocked by this specific gate, matching the split follow-up.

## Publication notes

Configured reviewers were `[]`, so this PR uses repository-default review routing. After this PR exists, #6233, #6228, #6210, and #6198 will get handoff comments and have `human-intervention` replaced with `human-intervention-fixed`; #6177 keeps `human-intervention` for the separate classification follow-up.

### DIFF
--- a/forge/forge_metadata.py
+++ b/forge/forge_metadata.py
@@ -90,7 +90,7 @@ from utility_scripts.metadata_index import (
     resolve_metadata_version,
     resolve_test_dir,
 )
-from utility_scripts.metrics_writer import read_pending_metrics
+from utility_scripts.metrics_writer import DYNAMIC_ACCESS_MIN_COVERAGE_RATIO, read_pending_metrics
 from utility_scripts.repo_path_resolver import (
     git_env_limited_to_repo_root,
     get_forge_subdir_name,
@@ -195,7 +195,7 @@ SCRIPT_RUN_METRICS_DIR = "script_run_metrics"
 ADD_NEW_LIBRARY_METRICS_FILE = "add_new_library_support.json"
 FIX_JAVAC_METRICS_FILE = "fix_javac_fail.json"
 FIX_JAVA_RUN_METRICS_FILE = "fix_java_run_fail.json"
-LOW_DYNAMIC_ACCESS_COVERAGE_RATIO = 0.10
+LOW_DYNAMIC_ACCESS_COVERAGE_RATIO = DYNAMIC_ACCESS_MIN_COVERAGE_RATIO
 HUMAN_INTERVENTION_LABEL_COLOR = "B60205"
 HUMAN_INTERVENTION_LABEL_DESCRIPTION = (
     "Requires manual follow-up because automated processing needs human attention"
--- a/forge/git_scripts/make_pr_new_library_support.py
+++ b/forge/git_scripts/make_pr_new_library_support.py
@@ -10,7 +10,7 @@ import re
 import shutil
 import subprocess
 import sys
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from git_scripts.common_git import (
     ensure_gh_authenticated,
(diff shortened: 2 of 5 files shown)


jormundur00 (changes requested):
Requesting changes because the "dynamic-access calls are disproportionate to metadata entries" case is being treated as a pre-PR publication blocker.

In `collect_new_library_support_quality_issues(...)`, the disproportionate call/metadata ratio is added to `quality_issues` when there is no reviewable evidence:

`forge/utility_scripts/metrics_writer.py`

That list is then used by `create_pull_request(...)` to raise before the PR is created:

`forge/git_scripts/make_pr_new_library_support.py`

I think this is the wrong behavior for this specific signal. A large gap between covered dynamic-access calls and metadata entries is suspicious, but it is not necessarily invalid: many call sites can be legitimately explained by a small number of metadata rules. This is exactly the kind of result that should be exposed to a human reviewer on the generated PR, with the call-site and metadata-rule evidence included in the PR body.

So I would keep the hard pre-PR gate for clearer quality failures, such as:

- dynamic-access coverage at or below the minimum threshold;
- covered dynamic-access calls with zero shipped/test-only metadata entries;
- zero library coverage or unsuccessful run status.

But the disproportionate call/metadata ratio should not prevent PR creation. It should instead create the PR, include the evidence section, and route/label it for human intervention so the reviewer can decide whether the metadata is actually insufficient.

As written, this can discard exactly the review surface needed to investigate the condition.

Overall, this looks like the cleaner successor to #6244 for the dynamic-access quality gate, and I think #6244 should probably be superseded by this PR. I would merge this path rather than #6244 once the ratio/evidence warning is moved out of the "refuse to create PR" path and into the human-review path.

PR #6177, opened 2026-05-06
[GenAI] Add support for io.grpc:grpc-auth:1.79.0 using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #6159

This PR introduces tests and metadata for io.grpc:grpc-auth:1.79.0, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 58631
- Cached input tokens: 1232384
- Output tokens: 11024
- Metadata entries: 3
- Test-only metadata entries: 18
- Iterations: 1
- Library coverage percentage: 49.29
- Generated lines of code: 258
- Tested library lines of code: 104

### Metadata/dynamic-access evidence

- Covered dynamic-access calls: 16
- Metadata entries: 3
- These counts are different dimensions: covered dynamic-access calls count observed call sites, while metadata entries count generated reachability-config items. Depending on the access type, a single metadata rule can cover multiple observed call sites, or no shipped rule may be required when the covered access does not target fixed library-owned metadata.
- Generated metadata rules:
  - when `io.grpc.auth.GoogleAuthLibraryCallCredentials$JwtHelper` is reached, make `java.lang.StackTraceElement[]` available for reflection
  - when `io.grpc.auth.GoogleAuthLibraryCallCredentials$MethodPair` is reached, make `java.lang.String[]` available for reflection
  - when `io.grpc.auth.GoogleAuthLibraryCallCredentials$JwtHelper` is reached, make `java.lang.reflect.Method[]` available for reflection
- Use the call-site and metadata-rule lists together to review whether the observed dynamic-access paths are explained by the generated reachability metadata.

### Metadata Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `a297c9297f77ec01586d25f1644c2564a502c5ac`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 16/16 covered calls (100.00%)
- Reflection: 16/16 covered calls (100.00%)

Library coverage:
- Instruction: 492/898 (54.79%)
- Line: 104/211 (49.29%)
- Method: 16/27 (59.26%)

### Local CI Verification

- Status: `success`
- Commands run: 15
- Fixup attempts: 0

Issue #6159, opened 2026-05-06
Support io.grpc:grpc-auth:1.79.0 service-account JWT reflection metadata


graemerocher:
## Summary

`io.grpc:grpc-auth:1.79.0` uses reflection in `io.grpc.auth.GoogleAuthLibraryCallCredentials$JwtHelper` to optimize unscoped `com.google.auth.oauth2.ServiceAccountCredentials` into `ServiceAccountJwtAccessCredentials`.

In a native image, downstream users can hit missing-method failures unless this reflective surface is registered. This was reported in micronaut-projects/micronaut-gcp#761 while using `micronaut-gcp-tracing`, which creates Stackdriver trace senders with `MoreCallCredentials.from(traceCredentials)`.

Downstream currently needs local reachability metadata for:
- `com.google.auth.oauth2.ServiceAccountCredentials` public methods, including `getQuotaProjectId()`
- `com.google.auth.oauth2.GoogleCredentials#getQuotaProjectId()`
- `com.google.auth.oauth2.ServiceAccountJwtAccessCredentials#newBuilder()`
- `com.google.auth.oauth2.ServiceAccountJwtAccessCredentials$Builder` methods used by `grpc-auth`: `build`, `setClientId`, `setClientEmail`, `setPrivateKey`, `setPrivateKeyId`, `setQuotaProjectId`

Downstream PR: micronaut-projects/micronaut-gcp#1476

### 1. Current native failure

The downstream issue reports this native-image-only warning/failure when `grpc-auth` initializes its JWT helper:

```text
java.lang.NoSuchMethodException: com.google.auth.oauth2.ServiceAccountCredentials.getQuotaProjectId()
    at java.lang.Class.getMethod(DynamicHub.java:2227)
    at io.grpc.auth.GoogleAuthLibraryCallCredentials$JwtHelper.<init>(GoogleAuthLibraryCallCredentials.java:318)
    at io.grpc.auth.GoogleAuthLibraryCallCredentials.createJwtHelperOrNull(GoogleAuthLibraryCallCredentials.java:224)
    at io.grpc.auth.GoogleAuthLibraryCallCredentials.<clinit>(GoogleAuthLibraryCallCredentials.java:53)
    at io.grpc.auth.MoreCallCredentials.from(MoreCallCredentials.java:35)
    at io.micronaut.gcp.tracing.zipkin.StackdriverSenderFactory.stackdriverSender(StackdriverSenderFactory.java:110)
```

[...]

PR #6250, opened 2026-05-07
[Forge] Fix test-only dependency stub classification

vjovanov:
## Root cause

Forge generated test-source stubs for missing dependency APIs in the real dependency namespace, for example `tests/src/io.grpc/grpc-auth/1.79.0/src/test/java/com/google/auth/oauth2/`. `SplitTestOnlyMetadataTask` treated every package discovered under `src/test/java` as test-owned, so metadata whose `type` referenced `com.google.auth.oauth2` was moved from shipped metadata into test-only metadata.

That made #6177 hide real dependency integration metadata for #6159 behind the test-only split solely because generated stubs existed in a dependency package. The prompt also did not explicitly forbid creating source stubs, fake replacements, or shadow classes for library/dependency APIs in their real packages.

## Fix

- Adds `MetadataGenerationUtils.discoverTestOnlyMetadataPackages`, a conservative split classifier that only treats packages as test-owned when their source files contain clear test-framework markers.
- Keeps `discoverTestPackages` unchanged for broader test-source package discovery, including user-code filter generation.
- Updates `SplitTestOnlyMetadataTask` to use the test-only classifier instead of all discovered test-source packages.
- Adds a split-task regression test showing dependency-package stub metadata remains shipped while real test fixture metadata moves to test-only metadata.
- Adds utility coverage proving dependency API stubs are ignored by the split classifier but still discovered by the broader package scanner.
- Tightens Forge dynamic-access rules so generated tests do not create source stubs, fake replacements, or shadow classes for library/dependency API types in their real packages.

## GitHub items addressed

- Addresses the dependency-stub/test-only classification side of PR #6177.
- Links to the original new-library request #6159 because the misplaced `com.google.auth.oauth2` metadata came from the generated `io.grpc:grpc-auth:1.79.0` support path.
- This is a Forge/build-logic system fix only. It does not edit generated metadata, stats, tests, or the failed/generated branch for #6177.

## Validation

Passed on the clean publication branch rebased to current `origin/master`:

- `git diff origin/master --check` -> passed
- `./gradlew :tck-build-logic:test --rerun-tasks --tests '*SplitTestOnlyMetadataTask*' --tests '*MetadataGenerationUtilsTests*'` -> `BUILD SUCCESSFUL`
- `PYTHONPATH=forge python -m pytest forge -k 'test_only or dependency_stub or metadata_generation or generated_test' -q` -> `2 passed, 239 deselected, 1 warning`

Reviewer validation from `runtime/reviews/fetch-prs-forge-coverage-metadata-evidence-forge-test-only-dependency-stub-classification-forge-review-2.md` reported no blocking findings, no validation gaps, and `Ready to publish: yes`.

## Publication notes

Configured reviewers were `[]`, so this PR uses repository-default review routing. After this PR exists, #6177 can be updated with the second durable Forge handoff; the earlier dynamic-access evidence handoff is #6249.

- `MetadataGenerationUtils.discoverTestOnlyMetadataPackages` walks and reads local test-source files under `tests/src/.../src/test/*`.
- `SplitTestOnlyMetadataTask` continues to read and write local reachability metadata and test-only metadata files.
- No new network, docker, or external-service access is added.

### DIFF
--- a/forge/prompt_templates/persistent/dynamic_access_rules.md
+++ b/forge/prompt_templates/persistent/dynamic_access_rules.md
@@ -6,6 +6,7 @@ Rules:
 - Do not broaden the patch into unrelated features.
 - Use upstream test sources only as behavioral examples.
 - Use documentation and source context only as API guidance.
+- Do not create source stubs, fake replacements, or shadow classes for library or dependency API types in their real packages. If a needed API is missing from the test classpath, add the correct test dependency or leave the call site unreached with an explanation.
 - Do not use reflection directly in the tests unless the public API requires it naturally.
 - Do not compile or run tests yourself. The workflow will do that externally.
 - Follow idiomatic `{test_language_display_name}` coding conventions.
--- a/tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/SplitTestOnlyMetadataTask.java
+++ b/tests/tck-build-logic/src/main/groovy/org/graalvm/internal/tck/harness/tasks/SplitTestOnlyMetadataTask.java
@@ -88,7 +88,7 @@ public class SplitTestOnlyMetadataTask extends CoordinatesAwareTask {
             throw new IllegalArgumentException("Cannot find tests directory for " + coordinate + ": " + testsDirectory);
         }
 
-        Set<String> testPackages = MetadataGenerationUtils.discoverTestPackages(testsDirectory);
+        Set<String> testPackages = MetadataGenerationUtils.discoverTestOnlyMetadataPackages(testsDirectory);
         Set<String> testResources = discoverTestResources(testsDirectory);
 
         ObjectNode libraryMetadata = requireObjectNode(objectMapper.readTree(metadataFile.toFile()), metadataFile);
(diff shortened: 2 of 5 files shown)


jormundur00 (approved):
The build-logic change looks correct to me. The old splitter treated every package under `src/test/*` as test-owned, which is too broad for generated support tests that contain helper/stub/probe classes in real library or dependency packages. Restricting the split classifier to packages with clear test-framework markers should prevent real library/dependency metadata from being moved into test-only metadata just because a source file exists under `src/test`.

The regression test covers the important shape from #6177: a dependency-package stub under `com.google.auth.oauth2` no longer causes `ServiceAccountCredentials` metadata to be moved to test-only, while actual generated test fixture metadata still moves.

Non-blocking follow-up: the repo already has checked-in test-only metadata that appears to contain real library metadata from the old classifier. Some files are mixed, so they should not be moved wholesale; only the real library entries should be split back into shipped metadata while probe/test-fixture entries remain test-only. Examples worth checking:

- `io.grpc:grpc-auth:1.79.0`
- `com.typesafe.akka:akka-protobuf-v3_2.13:2.6.21`
- `com.typesafe.akka:akka-protobuf-v3_2.13:2.8.8`
- `com.typesafe.akka:akka-protobuf-v3_3:2.9.0-M2`
- `jakarta.xml.bind:jakarta.xml.bind-api:3.0.0`
- `jakarta.xml.bind:jakarta.xml.bind-api:4.0.0`
- `org.apache.pekko:pekko-protobuf-v3_2.13:1.1.5`
- `org.apache.pekko:pekko-protobuf-v3_3:1.0.1`

I would not block this PR on that cleanup, since this PR fixes the splitter behavior going forward, but it would be good to either include that cleanup or open/link a follow-up so the already-generated misplaced metadata does not remain in the repository.

PR #6230, opened 2026-05-06
[GenAI] Add support for org.projectlombok:lombok:1.18.20 using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #1559

This PR introduces tests and metadata for org.projectlombok:lombok:1.18.20, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 186745
- Cached input tokens: 2503680
- Output tokens: 43236
- Metadata entries: 14
- Iterations: 5
- Library coverage percentage: 26.15
- Generated lines of code: 597
- Tested library lines of code: 250

### Metadata Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `6f2b25d4adc6fced3e9bd4c9df1f74664c861511`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 24/24 covered calls (100.00%)
- Reflection: 18/18 covered calls (100.00%)
- Resources: 6/6 covered calls (100.00%)

Library coverage:
- Instruction: 1310/5546 (23.62%)
- Line: 250/956 (26.15%)
- Method: 39/245 (15.92%)

### Local CI Verification

- Status: `success`
- Commands run: 15
- Fixup attempts: 0

PR #6252, opened 2026-05-07
[Forge] Block generated native-image test skips

vjovanov:
## What does this PR do?

Addresses human-intervention item #6230.

Forge generated `org.projectlombok:lombok:1.18.20` support with a successful local CI result, but the generated native-image tests could return early instead of exercising or asserting the native path. The blocked cases were introduced by a post-generation update, so prompt guidance alone was insufficient.

This adds a generated-test quality gate that rejects native-image short-circuits after generation, before strategy commit finalization, and during local CI verification. The gate reports native-image guarded `return`, empty optional/collection/null returns, assumptions/aborts, and `@DisabledInNativeImage`, while preserving accepted assertions such as `NativeImageSupport.isUnsupportedFeatureError(...)`.

Validation:
- `PYTHONPATH=$PWD/forge python -m pytest forge/tests/test_test_quality_checks.py` -> `10 passed`
- `PYTHONPATH=$PWD/forge python -m pytest forge -k 'test_quality or generated_test or native_image or dynamic_access'` -> `40 passed, 208 deselected`
- `PYTHONPATH=$PWD/forge python -m pytest forge/tests/test_workflow_strategy.py forge/tests/test_local_ci_verification.py` -> `25 passed`
- `git diff --check origin/master...HEAD` -> passed

- `forge/utility_scripts/test_quality_checks.py` scans generated test source files under `tests/src/...`.
- `forge/utility_scripts/local_ci_verification.py` inspects changed generated test source paths before running the normal local CI gates.
- No new network, docker, or external service access is added.

### DIFF
--- a/forge/ai_workflows/add_new_library_support.py
+++ b/forge/ai_workflows/add_new_library_support.py
@@ -47,7 +47,12 @@ from utility_scripts.source_context import (
     resolve_test_source_layout,
 )
 from utility_scripts.stage_logger import log_stage
-from utility_scripts.test_quality_checks import cleanup_scaffold_placeholder_tests, format_placeholder_occurrence
+from utility_scripts.test_quality_checks import (
+    cleanup_scaffold_placeholder_tests,
+    find_native_image_skip_guards,
+    format_native_image_skip_occurrence,
+    format_placeholder_occurrence,
+)
 from utility_scripts.metrics_writer import create_failure_run_metrics_output
 from utility_scripts.strategy_loader import require_strategy_by_name
 from utility_scripts.workflow_setup import resolve_graalvm_java_home, validate_repo_paths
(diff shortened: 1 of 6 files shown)


jormundur00 (changes requested):
[Medium] Non-Java generated tests can still bypass the native-image path

The new generated-test quality gate only partially covers the languages Forge can generate. Groovy scaffolded tests are not scanned at all because `TEST_SOURCE_EXTENSIONS` is limited to `.java`, `.kt`, and `.scala`, even though Forge has a Groovy scaffold flavor that writes `src/test/groovy/**/*.groovy` tests. Kotlin and Scala files are included, but the native-skip detection is still Java-centric: `SKIP_RETURN_PATTERN` requires semicolon-terminated `return;` statements, and the native-image indicator/helper extraction also expects Java-style assignment and method syntax.

That means generated Kotlin/Scala tests can still short-circuit native execution without being reported, for example:

```kotlin
if (ImageInfo.inImageRuntimeCode()) return
```

or:

```kotlin
val nativeImage = ImageInfo.inImageRuntimeCode()
if (nativeImage) return
```

Generated Groovy tests would currently be skipped by this gate entirely. Since Forge can generate Kotlin, Scala, and Groovy tests, this creates a false sense that the new gate covers all generated test sources. Please either add language-specific coverage for Kotlin/Scala/Groovy, or narrow the gate's stated/file coverage until those languages are supported.

The PR adds a useful quality gate for detecting generated tests that skip native-image execution, and the Java path looks directionally correct. The main issue is that the gate does not reliably cover all generated test languages yet.

Issue #6807, opened 2026-05-11
Update existing library: org.flywaydb:flyway-core:12.5.0


MartinHaeusler:
### Additional context and logs

I just got the following error when building a spring boot 4 project with flyway with graalvm and starting it:

```
 org.graalvm.nativeimage.MissingReflectionRegistrationError: Cannot reflectively invoke method 'public java.lang.String org.flywaydb.core.internal.configuration.extensions.PrepareScriptFilenameConfigurationExtension.getUndoFilename()'. To allow this operation, add the following to the 'reflection' section of 'reachability-metadata.json' and rebuild the native image:

  {
    "type": "org.flywaydb.core.internal.configuration.extensions.PrepareScriptFilenameConfigurationExtension",
    "methods": [
      {
        "name": "getUndoFilename",
        "parameterTypes": []
      }
    ]
  }
```

### GraalVM version

GraalVM 25.0.2

### Operating system

Linux

### CPU architecture

AARCH64 (ARM64)

vjovanov:
Human intervention needed

The automated job failed to complete automatically for `org.flywaydb:flyway-core:12.5.0`. Preserved work branch: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fvjovanov%2Fhuman-intervention%2Fissue-6807-library-update-request-org.flywaydb-flyway-core-12.5.0-d9f49194

[...]

Issue #5926, opened 2026-05-05
Update existing library: org.apache.tomcat.embed:tomcat-embed-core:11.0.18


alina-yur:
### Full Maven coordinates

org.apache.tomcat.embed:tomcat-embed-core:11.0.18

### Additional context and logs

Latest Spring Petclinic fails at runtime because Tomcat reflectively introspects protocol methods.

Tested with:
- spring-projects/spring-petclinic
- Spring Boot 4.0.3
- org.apache.tomcat.embed:tomcat-embed-core:11.0.18`
-  GraalVM 25.0.2

  ```text
MissingReflectionRegistrationError:
  Cannot reflectively invoke method
  org.apache.coyote.AbstractProtocol.getProperty(java.lang.String)
```

  After fixing that one method, the same pattern exposed more missing methods/classes:

  ```text
  org.apache.coyote.AbstractProtocol.getName()
  org.apache.coyote.http11.AbstractHttp11Protocol.isSSLEnabled()
```
  Was solved locally by registering hints:

 ```text
hints.reflection().registerType(AbstractProtocol.class,
MemberCategory.INVOKE_PUBLIC_METHODS);
  hints.reflection().registerType(AbstractHttp11Protocol.class,
MemberCategory.INVOKE_PUBLIC_METHODS);
```

  Equivalent reachability metadata would be roughly:

```text
  {
    "reflection": [
      {
        "type": "org.apache.coyote.AbstractProtocol",
        "allPublicMethods": true
      },
      {
        "type": "org.apache.coyote.http11.AbstractHttp11Protocol",
        "allPublicMethods": true
      }
    ]
  }
``

### GraalVM version

GraalVM 25.0.2

### Operating system

Linux

### CPU architecture

AMD64 (x86-64)

vjovanov:
Human intervention needed

The automated job failed to produce a valid library update for `org.apache.tomcat.embed:tomcat-embed-core:11.0.18`.

Preserved work branch: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fvjovanov%2Fhuman-intervention%2Fissue-5926-library-update-request-org.apache.tomcat.embed-tomcat-embed-core-11.0.18-8d83a35f

[...]

Issue #3899, opened 2026-04-30
Update existing library: ch.qos.logback:logback-core:1.5.26


jkuipers:
### Full Maven coordinates

ch.qos.logback:logback-classic:1.5.26
ch.qos.logback:logback-core:1.5.26

### Additional context and logs

Starting with [release 1.5.26](https://github.com/qos-ch/logback/releases/tag/v_1.5.26), Logback has introduced new version properties files that are read at initialization time. When using a native image these files aren't available as resources, causing warnings during startup:
```
|-WARN in ch.qos.logback.classic.LoggerContext[default] - Versions of logback-classic and ? are different or unknown.
```
This is especially annoying because a single warning during Logback initialization will cause *all* of its internal startup logging to become visible, so this produces a lot of noise.

For logback-classic, the file that needs to be available as a resource is `ch/qos/logback/classic/logback-classic-version.properties`. For logback-core, it's `ch/qos/logback/core/logback-core-version.properties`. 

### GraalVM version

Not relevant, but was a 25 version

### Operating system

Linux

### CPU architecture

AMD64 (x86-64)

kimeta:
Human intervention needed

The automated job failed for `ch.qos.logback:logback-classic:1.5.26` (`library-update-request`, workflow status `failure`). Preserved work branch: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fkimeta%2Fhuman-intervention%2Fissue-3899-library-update-request-ch.qos.logback-logback-classic-1.5.26-51f182a1

[...]

Issue #2110, opened 2026-04-23
org.apache.kafka:kafka-clients:4.2.0 native image still fails without oauthbearer defaults; existing 3.5.1 metadata does not resolve it


graemerocher:
## Summary

Current `org.apache.kafka:kafka-clients:4.2.0` still fails in a GraalVM native image, and the existing metadata published under `metadata/org.apache.kafka/kafka-clients/3.5.1` does not resolve it.

This came from validating Micronaut Kafka issue `micronaut-projects/micronaut-kafka#1213` against the current `6.0.x` line and then checking whether the existing reachability metadata helps.

Upstream issue links:
- Micronaut Kafka issue: https://github.com/micronaut-projects/micronaut-kafka/issues/1213
- Validation comment being followed up here: https://github.com/micronaut-projects/micronaut-kafka/issues/1213#issuecomment-4305176848

### 1. Old 3.9.1 failure is **not** the current 6.0.x failure

The older Micronaut Kafka issue comment validated `kafka-clients:3.9.1` failing in native execution with:
- `ClassNotFoundException: java.security.AccessController`
- `NoSuchMethodException: javax.security.auth.Subject.current()`

On the current Micronaut Kafka `6.0.x` line, the managed Kafka client version is `4.2.0`, and the current native failure is different.

### 2. Current `kafka-clients:4.2.0` failure

I built a minimal Micronaut app against the current Micronaut Kafka `6.0.x` source and let it create a Kafka producer on startup with:
- `security.protocol=SASL_PLAINTEXT`
- `sasl.mechanism=PLAIN`
- `sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required username="admin" password="admin-secret";`

The native executable fails before any broker handshake with:

```text
Caused by: org.apache.kafka.common.config.ConfigException: Invalid value org.apache.kafka.common.security.oauthbearer.DefaultJwtRetriever for configuration sasl.oauthbearer.jwt.retriever.class: Class org.apache.kafka.common.security.oauthbearer.DefaultJwtRetriever could not be found.
    at org.apache.kafka.common.config.ConfigDef.parseType(ConfigDef.java:789)
    at org.apache.kafka.common.config.SaslConfigs.addClientSaslSupport(SaslConfigs.java:388)
    at org.apache.kafka.clients.producer.ProducerConfig.<clinit>(ProducerConfig.java:526)
```

### 3. Existing metadata does not resolve it

[...]

PR #6848, opened 2026-05-12
Improve Forge library update target resolution

### LINKED ISSUE #6847: Improve Forge library-update-request target resolution
## Summary

Improve Forge's `library-update-request` automation so one issue produces one PR for the Maven coordinate in the issue title, while body coordinates remain contextual evidence only.

## Step-by-step changes requested

1. Resolve a single update target from the issue title coordinate.
2. Match requested versions in this order: `tested-versions`, exact `metadata-version`, `default-for` regex, then uncovered `new-version`.
3. Represent the resolved update as a structured `LibraryUpdateTarget` containing the requested coordinate, match type, matched index entry, resolved metadata/test versions, and edit paths.
4. For existing support matches, edit the matched metadata directory and matched test project while validating with the requested coordinate.
5. For uncovered versions, create new support by cloning the closest compatible baseline: same major/minor, same major, `latest: true`, then scaffold only when no usable baseline exists.
6. Use the resolved metadata version and test version everywhere paths are staged, finalized, loaded for source context, checked by local CI, and reported in metrics/PRs.
7. Pass issue-requested missing metadata evidence from the issue body into agent prompts, with explicit guidance that metadata additions need conditions, preferably `typeReached`.
8. Keep normal finalization only: `generateMetadata`, requested-coordinate tests, `splitTestOnlyMetadata`, `checkMetadataFiles`, style checks, and `generateLibraryStats`.
9. Preserve generated metadata/tests/stats/logs before failure reset so human-intervention branches contain the failed work, not only an empty checkpoint.
10. Include requested coordinate, match type, matched/resolved metadata and test versions, and validation result in the generated PR body.

## Acceptance tests

- Title coordinate wins over body coordinates.
- Requested version in `tested-versions` resolves to the existing support.
- Requested version equal to `metadata-version` resolves to that entry.
- Requested version matching `default-for` resolves to that entry.
- Uncovered requested version becomes a `new-version` target.
- New version support clones the closest compatible baseline.
- Staging/finalization uses the resolved test version.
- Failure preservation keeps generated files before reset.
- PR description includes target-resolution and validation details.

jormundur00:
## Summary

Fixes #6847.

This draft PR improves Forge's `library-update-request` automation so a single issue produces one PR for the Maven coordinate in the issue title. Coordinates in the issue body are treated as contextual evidence only.

The current policy is conservative for updates: exact `metadata-version` matches edit in place, while `tested-version` and `default-for` matches split into a requested-version target before any generated tests or metadata are added. This avoids adding dynamic-access metadata discovered for a newer requested version to an older shared metadata directory.

The PR also changes issue-requested metadata handling to be prompt-based. Forge now forwards the reporter's issue body to the agent and instructs it to infer the requested metadata, add or keep tests that exercise it through public APIs, and add `typeReached` conditions when the issue omits them. Forge no longer hardcodes parser rules for specific issue wording and no longer applies parsed metadata as a post-generation fallback.

## Step-by-step changes

1. Added `LibraryUpdateTarget` resolution in `forge/utility_scripts/metadata_index.py`.
   - Captures requested coordinate.
   - Captures match type: `tested-version`, `metadata-version`, `default-for`, or `new-version`.
   - Captures the matched `index.json` entry when one exists.
   - Captures resolved metadata/test versions and edit directories.

2. Updated library-update resolution behavior.
   - Resolution still checks `tested-versions`, exact `metadata-version`, then `default-for`.
   - Exact `metadata-version` matches edit that target in place.
   - `tested-version` and `default-for` matches now split into new requested-version support by cloning the matched support.
   - No-match requests create new requested-version support as before.

3. Added requested-version splitting for shared targets.
   - Clones metadata, tests, and stats from the matched baseline.
   - Rewrites version-specific coordinates and URLs in cloned files.
   - Adds a new `index.json` entry for the requested version and marks it latest.
   - Removes the requested version and later numeric tested versions from the old shared entry.
   - Keeps lower tested versions on the old entry; prerelease qualifiers like `1.0.1-RC1` remain below `1.0.1`.

4. Made issue-requested metadata handling prompt-based.
   - Fetches the GitHub issue body during issue lookup and claim refresh.
   - Passes the issue body into both `library-update-request` and `library-new-request` workflows.
   - Keeps the full issue body instead of keyword-filtering snippets, with a larger 50k character cap.
   - Prompts the agent to infer the reporter-requested metadata from prose, logs, and snippets.
   - Requires tests that exercise each requested metadata need through public library API paths.
   - Requires requested metadata to be included when generated metadata does not already contain it.
   - Requires appropriate metadata conditions, preferably the narrowest valid `typeReached`, when the issue does not specify conditions.
   - Removes deterministic issue-body metadata extraction and the automatic post-generation metadata merge.

5. Fixed validation and stats paths for split/shared targets.
   - Finalization validates both requested and resolved metadata coordinates when they differ.
   - Generated PR bodies list all relevant validation commands.
   - Execution metrics now write to the resolved metadata stats directory.
   - Source-context URL templates now render with the requested version, not the baseline metadata version.

6. Preserved failed generated work before reset.
   - Failed library-update runs snapshot resolved tests, metadata, index, and stats.
   - After reset to checkpoint, generated target files are restored so human-intervention branches contain useful failed work.

7. Kept validated run metrics schema unchanged.
   - Target-resolution details are not added to run metrics.
   - The workflow writes PR-only target details to `.library_update_target.json` in the metrics workspace.
   - `make_pr_improve_coverage.py` reads that sidecar when constructing the generated PR body.

8. Added a `review-library-update-request` skill.
   - Reviews generated `library-update-request` PRs for dynamic-access coverage percentage regressions between versions.
   - Requires linked issue metadata requests to be present and exercised through public library API tests.
   - Reuses relevant new-library review rules for scope, scaffold-only tests, native-image skips, version-pinned tests, package-bypassing tests, and legacy metadata files.
   - Explicitly avoids the new-library minimum dynamic-access coverage threshold.

9. Added regression coverage.
   - Title coordinate wins over body coordinates.
   - `tested-versions`, exact `metadata-version`, and `default-for` resolution are covered.
   - Shared `tested-version` and `default-for` matches split into requested-version targets.
   - Splitting trims requested/later tested versions from the old entry.
   - Uncovered versions clone the closest compatible baseline.
   - Issue-requested metadata context is forwarded and formatted as generic prompt requirements.
   - Staging/finalization use resolved paths.
   - Execution metrics use the resolved metadata stats path.
   - Source context uses the requested version.
   - Failure preservation restores generated files after reset.
   - PR body reads target details from the PR-only sidecar.

## Validation

Ran focused Forge tests:

```bash
PYTHONPATH=forge forge/.venv/bin/python -m unittest \
  forge.tests.test_schema_validator \
  forge.tests.test_library_update_target \
  forge.tests.test_make_pr_improve_coverage \
  forge.tests.test_workflow_strategy \
  forge.tests.test_forge_metadata.LibraryUpdateIssueTests \
  forge.tests.test_metrics_paths \
  forge.tests.test_source_context \
  forge.tests.test_issue_requested_metadata \
  forge.tests.test_dynamic_access_iterative_strategy
```

Result: passed, 52 tests.

Also ran Python compile checks for touched Forge modules and tests; result: passed.

Additional focused validation after switching issue-requested metadata to prompt-based handling:

```bash
PYTHONPATH=forge forge/.venv/bin/python -m unittest \
  forge.tests.test_issue_requested_metadata \
  forge.tests.test_library_update_target \
  forge.tests.test_forge_metadata.LibraryUpdateIssueTests
```

Result: passed, 18 tests.

Local read-only prompt smoke tests were run with `pi --no-tools --no-session` against representative open `library-update-request` issues. The model inferred the expected requested metadata for examples including HikariCP #1065, Tomcat #5926, Logback #3899, Flyway #6807/#1080/#424, Kafka #2110, Thymeleaf #2094, and Caffeine #436, while correctly reporting low confidence for MySQL #453 where the issue points to external metadata rather than inlining exact missing entries.

Validated the new review skill:

```bash
forge/.venv/bin/python /home/jovan/.codex/skills/.system/skill-creator/scripts/quick_validate.py   skills/review-library-update-request
```

Result: `Skill is valid!`

## Additional update: `fromJar` allowed-packages merge

`generateMetadata --agentAllowedPackages=fromJar` now preserves any existing `allowed-packages` on the matching `index.json` entry and appends package roots derived from the artifact JAR instead of replacing the existing list. This keeps cloned/update support with broader package roots, such as `ch.qos.logback`, from being narrowed to only the direct artifact package, such as `ch.qos.logback.classic`.

Additional validation:

```bash
./gradlew :tck-build-logic:test --tests org.graalvm.internal.tck.GenerateMetadataTaskTests --stacktrace --rerun-tasks
```

Result: passed.

## Additional update: resolved target edit guardrails

Forge now passes an explicit resolved edit scope into dynamic-access and reporter-requested metadata prompts, including the absolute target test project, test source root, and build file. The prompts now tell agents to edit only that resolved target and not cloned baseline or other versioned test directories.

`make_pr_improve_coverage.py` also checks for dirty paths outside the resolved target paths after staging the expected PR files and before rebasing. If a run leaves changes like an old baseline test directory edit behind, finalization now fails early with the unexpected paths instead of surfacing as a generic `git rebase` dirty-worktree error.

Additional validation:

```bash
forge/.venv/bin/python -m unittest \
  tests.test_make_pr_improve_coverage \
  tests.test_library_update_target \
  tests.test_workflow_strategy
```

Result: passed, 20 tests. Also ran `git diff --check` and Python compile checks for touched Forge modules; result: passed.

## Additional update: split tested-version suffixes

The middle-version split rule is now implemented in Forge's library-update automation path, not just applied manually to generated PRs. When automation creates a new metadata version from a shared baseline, it moves the requested version and every later tested version from the old shared entry onto the new entry.

For example, splitting `12.5.0` out of an entry with `12.3.0`, `12.4.0`, `12.6.0`, and `12.6.1` leaves `12.3.0`/`12.4.0` on the old metadata and assigns `12.5.0`/`12.6.0`/`12.6.1` to the new metadata. Earlier qualifier versions such as `1.0.1-RC1` remain on the old entry when splitting `1.0.1`.

Additional validation:

```bash
forge/.venv/bin/python -m unittest tests.test_library_update_target
```

Result: passed, 13 tests. Also ran Python compile checks for the touched Forge workflow/test modules; result: passed.

## Additional update: Forge docs

Updated the Forge workflow documentation so `library-update-request` target resolution and split behavior are described in the docs, not only in code and tests. The docs now cover title-coordinate targeting, `tested-versions`/`metadata-version`/`default-for` resolution, exact metadata-version reuse, cloned requested-version support for shared targets, and moving later tested versions when a middle version is split.

Documentation-only change; no additional tests were run for this commit.


vjovanov (approved):

vjovanov (commented):
## Overview

Rewires the `library-update-request` Forge pipeline so a single issue produces a single PR scoped to the title coordinate, with four substantive shifts:

1. **`LibraryUpdateTarget` resolution** (`utility_scripts/metadata_index.py`) classifies the requested version against `tested-versions` / `metadata-version` / `default-for` / `new-version` and surfaces resolved metadata/test paths everywhere downstream (staging, finalization, source-context URLs, local-CI verification, execution metrics).
2. **Splitting shared targets** (`improve_library_coverage.py`) — when a requested version sits in `tested-versions` or matches `default-for`, the baseline is cloned, version strings are rewritten, and the requested version (plus all later tested versions) is moved onto a new latest entry.
3. **Prompt-based reporter-requested metadata** — full issue body (≤50 k chars) is forwarded to the agent rather than parsed by hardcoded regex.
4. **Java change to `MetadataGenerationUtils`** — `fromJar` now merges with the existing `allowed-packages` instead of replacing them.

## What looks solid

- **Test coverage is genuinely thorough.** 52 new tests including title-vs-body coordinate precedence, all four match types, split semantics with later/earlier tested versions, qualifier ordering (`1.0.1-RC1` stays below `1.0.1`), failure preservation, `fromJar` merge, source-context URL rewriting, and finalization scope checks.
- **`_finalize_successful_iteration` now validates both requested and resolved coordinates** when they differ (`workflow_strategy.py:436`) — the right place to enforce that a split didn't break the older baseline.
- **`assert_no_out_of_scope_changes`** in `make_pr_improve_coverage.py` is a nice belt-and-braces check that catches the "agent edited the baseline test dir instead of the cloned one" failure mode early, with a clear error message rather than a generic rebase failure.
- **`LibraryUpdateTarget` is `@dataclass(frozen=True)`** — good immutable shape; field set is the right one to thread through.

## Concerns

**1. Issue body is passed verbatim into prompts — prompt-injection vector.**
`extract_issue_requested_metadata_context` (`forge_metadata.py:4032`) strips and truncates but doesn't sanitize. A reporter can place arbitrary "ignore previous instructions" content at the top of the issue body and it will reach the agent. The 50 k cap doesn't mitigate this. Worth considering at least a leading boundary marker like `<<<reporter-issue-body>>>` and a system-level instruction in the prompt template not to follow reporter content as instructions. Not a blocker for this PR — the agent's tool surface here is already constrained to writing repo files within a worktree — but worth a tracking issue.

**2. `_rewrite_text_files` does plain `str.replace` across the cloned tree.**
`improve_library_coverage.py:206-223`. Sorting by length DESC mitigates the obvious `1.0` vs `1.0.0` aliasing, but the function still walks every text file (including transitive `build.gradle`, `*.md`, license headers, JSON, README excerpts) and substitutes any occurrence of the baseline version string. For a baseline like `1.0` or `4.0`, version-like substrings *do* appear in unrelated content (JDK version comments, link anchors, copyright years for releases). Two-digit versions are realistic for many libraries (e.g. `commons-io:2.16`). Consider either restricting rewrites to specific filenames (`build.gradle`, `*.json`), or anchoring the search with a non-word boundary (`\b1\.0\b` regex). Not currently exercised destructively in the test fixtures, but a hazard once it meets real libraries.

**3. `prepare_library_update_target` scaffold fallback runs `./gradlew scaffold` synchronously and unconditionally.**
`improve_library_coverage.py:403`. `subprocess.run([..., "scaffold", ...], check=True)` will abort the workflow if scaffold fails, *after* the function has logged that no usable baseline was found. No catch, no metrics annotation, no surfacing to `create_failure_run_metrics_output`. If scaffold ever fails (Gradle daemon hiccup, missing artifact), this surfaces as a raw `CalledProcessError`. Worth wrapping or at least logging which path was taken.

**4. `_version_is_at_or_after` edge case: equivalent-but-different version strings.**
`1.0.1.0` vs `1.0.1` both pad to `(1,0,1,0)`. The function returns `False` (`>` is strict) for `1.0.1.0` against requested `1.0.1`, meaning a tested-version listed as `1.0.1.0` would *stay on the old entry* even though it represents the same release as the requested split point. Probably theoretical for the libraries in this repo, but noting it.

## Minor nits

- The new `review-library-update-request/SKILL.md` (213 lines) duplicates several rules verbatim from `review-library-new-request`. Could share a fragment, but not blocking.
- `clone_library_update_support` does five `_rewrite_text_files` invocations sequentially over the same trees; would be cheaper as a single walk with a combined replacements list. Performance is fine for typical library sizes.
- Two separate `from utility_scripts.issue_requested_metadata import ...` lines (`improve_library_coverage.py:25-26`) should be combined per the existing style elsewhere in the file.

## Recommendation

LGTM, with two issues worth tracking as follow-ups after merge: the rewrite-everything-in-text-files substitution scope (concern #2) and the prompt-injection envelope around the reporter issue body (concern #1). Neither is a regression vs. the previous behavior, and both are tractable as separate tickets. Test coverage and CI are convincing, and the structural improvements (single resolved target threaded through all sites, finalization scope check, failure preservation) clearly improve the workflow's robustness.

PR #6927, opened 2026-05-14
Separate root coverage updates from coverage publishing

### LINKED ISSUE #6926: Publish coverage fails when updating root COVERAGE.md
## Problem

The scheduled `Publish coverage` workflow currently tries to commit the generated root `COVERAGE.md` directly back to `master`.

That direct push is rejected by branch protection because changes to `master` must go through a pull request and required status checks are enforced.

## Expected behavior

`publish-scheduled-coverage.yml` should continue publishing generated coverage artifacts to the `stats/coverage` branch, but root `COVERAGE.md` updates should be carried through a normal PR path.

## Proposed fix

Keep the `stats/coverage` update in `publish-scheduled-coverage.yml`, and move root `COVERAGE.md` refreshes into the generated bulk-update PR flow in `verify-new-library-version-compatibility.yml`.

jormundur00:
## Summary

- keep `publish-scheduled-coverage.yml` focused on publishing generated artifacts to `stats/coverage`
- move root `COVERAGE.md` refreshes into the generated bulk-update PR workflow
- update docs and bulk-update review guidance for the new `COVERAGE.md` change path

## Validation

- `git diff --check`
- YAML parse for changed workflows
- `./gradlew generateReadmeBadgeSummary -PreadmeMetricsOutputRoot=<tmp> --stacktrace`

Fixes #6926


kimeta:
Is it intentional that the root `COVERAGE.md` on `master` is now refreshed only on the successful bulk-update path in `verify-new-library-version-compatibility.yml`, while the `COVERAGE.md` on the `stats/coverage` branch continues to be refreshed periodically by `publish-scheduled-coverage.yml`?

That means the published coverage branch can keep updating on schedule, but the root `COVERAGE.md` can remain stale if the compatibility workflow keeps failing, has no successful tested-version updates, or coverage-affecting changes land through another path such as new-library or metadata/stat fixes.

jvukicev:
> Is it intentional that the root `COVERAGE.md` on `master` is now refreshed only on the successful bulk-update path in `verify-new-library-version-compatibility.yml`, while the `COVERAGE.md` on the `stats/coverage` branch continues to be refreshed periodically by `publish-scheduled-coverage.yml`?
> 
> That means the published coverage branch can keep updating on schedule, but the root `COVERAGE.md` can remain stale if the compatibility workflow keeps failing, has no successful tested-version updates, or coverage-affecting changes land through another path such as new-library or metadata/stat fixes.

I've intentionally left it that way. I wouldn't block the update of the `stats/coverage` branch just because these `library-bulk-updates` might fail. As the `COVERAGE.md` table needs to be updated in a PR, this feels like the best solution without major changes.

A different solution would be to make `publish-coverage` generate it's own PRs, that we would create a new label for and have our automated review workflows merge it, but I'm unsure if we should introduce more labels and PRs for everything (and we would at that point probably make `publish-coverage.yml` run on a less frequent-schedule, perhaps once a day).

Either way, I'd merge this so coverage updates aren't blocked for now, and we can later think about what a better solution might be, if we want to go some other route.

kimeta (approved):

vjovanov (approved):

PR #7120, opened 2026-05-18
CI workflow hardening

matneu:
## Summary

This branch hardens CI workflow shell handling by tightening quoting, removing avoidable shell evaluation, and preventing token exposure in Git remotes.

## Changes

  - Quote workflow script paths in Docker-disabling setup and enable strict shell mode.
  - Escape sed search/replacement values in the Docker image sync workflow.
  - Use null-delimited grep/xargs handling for safer replacement across files.
  - Replace bash -c execution in the consecutive test runner with env/command arrays.
  - Validate scheduled release tags before computing the next release version.
  - Replace executable sed release tag arithmetic with explicit shell parsing.
  - Quote release tag usage in git tag, git push, and gh release create.
  - Avoid embedding GH_TOKEN in the coverage publish Git remote URL.

### DIFF
--- a/.github/workflows/create-scheduled-release.yml
+++ b/.github/workflows/create-scheduled-release.yml
@@ -83,7 +83,7 @@ jobs:
         id: set-matrix
         run: |
           LATEST_TAG=$(git tag --list | sort -V | tail -1)
-          ./gradlew generateChangedCoordinatesMatrix -PbaseCommit=$(git show-ref -s $LATEST_TAG) -PnewCommit=$(git rev-parse HEAD)
+          ./gradlew generateChangedCoordinatesMatrix -PbaseCommit="$(git show-ref -s "$LATEST_TAG")" -PnewCommit="$(git rev-parse HEAD)"
 
   release:
     needs: get-changed-metadata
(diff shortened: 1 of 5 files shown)


jormundur00 (changes requested):
Thanks for tightening the CI shell handling. Most of this looks good, especially removing the avoidable `bash -c` execution in the consecutive test runner and moving the coverage push token out of the remote URL.

I found one concrete regression risk that should be fixed before merging:

- `.github/workflows/scripts/disable-docker.sh`: adding `set -euo pipefail` makes the existing `sudo mkdir /etc/systemd/system/docker.service.d` a hard failure if that directory already exists. The script comment says it is idempotent, and `restore-docker.sh` also treats this directory as something that may already exist or be non-empty. On a runner image or local Forge environment with an existing Docker drop-in directory, this now aborts the workflow before installing the proxy config. Please change this to `sudo mkdir -p /etc/systemd/system/docker.service.d`.

Lower priority hardening note: the script still interpolates `SCRIPT_DIR` into `sudo bash -c ...`. The added quotes handle spaces, but it still asks a second shell to re-parse a generated command string. For a hardening PR, avoiding `bash -c` entirely would be cleaner, for example using `sudo tee -a /etc/inetd.conf < "$SCRIPT_DIR/discard-port.conf"` and `sudo install -D -m 0644 "$SCRIPT_DIR/dockerd.service" /etc/systemd/system/docker.service.d/http-proxy.conf`.

I did not find issues in the `run-consecutive-tests.sh`, `sync-docker-images.yml`, release tag arithmetic, or coverage credential-helper changes. I checked the patch locally with `git diff --check`, `bash -n` on the changed shell scripts, and a small smoke test for the sed escaping logic.

jormundur00 (approved):
Thanks, the follow-up commit addresses my previous findings: `disable-docker.sh` now uses `mkdir -p`, and the remaining `sudo bash -c` calls were replaced with `tee`/`install`.

I re-reviewed the updated PR locally and did not find additional issues. I checked the final diff with `git diff --check`, `bash -n` on the changed shell scripts, and a focused smoke test for the Docker image sync sed escaping with `/`, `&`, `|`, and filenames containing spaces.

PR #7121, opened 2026-05-18
Limit automated dependency issue creation per triage run

matneu:
- Caps automated dependency issue creation at 20 new issues per triage run.
  - Logs skipped dependency candidates after the cap is reached.
  - Adds a diagnostic comment to the source issue listing skipped dependency issue candidates.

### DIFF
--- a/.github/workflows/scripts/open-dependency-issues-and-link-blockers.js
+++ b/.github/workflows/scripts/open-dependency-issues-and-link-blockers.js
@@ -24,6 +24,9 @@
 const fs = require('fs');
 const path = require('path');
 
+const MAX_NEW_ISSUES = 100;
+const MAX_SKIPPED_ISSUES_IN_COMMENT = 50;
+
 /**
  * Resolves the workspace directory used to read repository files and helper scripts.
  */
@@ -805,6 +808,7 @@ module.exports = async function openDependencyIssuesAndLinkBlockers({ github, co
 
   const gaIssue = new Map([[rootGA, sourceIssueNumber]]);
   let createdCount = 0;
+  const skippedIssueCreations = [];
 
   const openIssues = await github.paginate(github.rest.issues.listForRepo, {
     owner,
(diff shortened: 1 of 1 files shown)


jormundur00 (commented):
I'm unsure if adding this cap is something we should be doing. Most bigger Maven libraries have more than 20 transitive dependencies, and supporting them also should be a priority when trying to support a requested library (as the requested library might need its transitive dependency to provide some specific metadata).

Another potential issue I see here (if we decide to go down this road) is that the current triaging is "dependency-first", as in it opens issues for the leaf nodes of a dependency tree first. This means that, with this cap, we would end up opening issues for the deepest transitive dependencies of a library, when it might be better to open the direct dependencies first (if we are under an issue number limit).

If we really MUST add a limit, for security reasons, I'd drastically increase the limit.

matneu:
What would be a reasonable limit? It's fine if the limit is high, but there should be some cap.

jvukicev:
> What would be a reasonable limit? It's fine if the limit is high, but there should be some cap.

I've done an analysis on the all current (and closed) user-authored `library-new-request` issues on the repo. The 99th percentile of the number of transitive dependencies of a library is 33, while the largest cases go for:
```
  org.apache.seata:seata-all:2.1.0                    360
  org.apache.spark:spark-sql_2.13:3.3.2                96
  org.apache.spark:spark-core_2.13:3.2.1               62
  io.quarkus:quarkus-junit:3.33.1                      40
  org.apache.hadoop:hadoop-client:2.7.3                38
  org.springframework.cloud:spring-cloud-stream-binder-rabbit:5.0.1 34
  org.jetbrains.kotlin:kotlin-maven-plugin:2.3.20      33
```

If we want to be sure we cover most of the most (ignoring this `seata-all`), I'd go for 100 as a limit. If you think that's too high, I suppose 50 works for most libs too, but we would end up with a bit more unrequested transitive dependencies.

jormundur00 (approved):

PR #7133, opened 2026-05-18
Limit workflow tokens and remove write-all permissions

matneu:
This PR removes broad write-all permissions from scheduled metadata workflows and narrows each job to the minimum GitHub scopes it needs. It also stops exposing GITHUB_TOKEN to dependency test jobs that execute third-party artifacts by replacing job-specific gh run view lookups  with the workflow run URL.

### DIFF
--- a/.github/actions/setup-native-build-tools/action.yml
+++ b/.github/actions/setup-native-build-tools/action.yml
@@ -118,6 +118,7 @@ runs:
         path: ${{ inputs.nbt-path }}
         token: ${{ inputs.github-token }}
         fetch-depth: 1
+        persist-credentials: false
 
     - id: read-version
       if: ${{ inputs.enabled-by-default == 'true' || steps.compute-ref.outputs.same-branch == 'true' }}
--- a/.github/workflows/test-all-metadata.yml
+++ b/.github/workflows/test-all-metadata.yml
@@ -6,9 +6,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  actions: read
   contents: read
-  issues: write
 
 concurrency:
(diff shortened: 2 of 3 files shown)


jormundur00 (changes requested):
The workflow permission narrowing looks directionally right, and the issue/PR creation paths still appear to have the scopes they need.

However, the dependency test jobs still expose a GitHub token through checkout. Both matrix test jobs removed `GITHUB_TOKEN` from the shell env, but `actions/checkout` persists credentials by default unless `persist-credentials: false` is set. Since these jobs execute third-party dependency tests after checkout, that still leaves a read-scoped token available through local git config, which undercuts the PR’s stated goal of not exposing `GITHUB_TOKEN` to dependency test jobs.

Please add this to the checkout steps in the dependency test jobs:

```yaml
with:
  persist-credentials: false
```

Affected spots:

- .github/workflows/test-all-metadata.yml, test-all-metadata job checkout
- .github/workflows/verify-new-library-version-compatibility.yml, test-all-metadata job checkout

Also worth checking .github/actions/setup-native-build-tools/action.yml, because its nested checkout also persists credentials by default if it runs before dependency tests.

jormundur00 (approved):
Reviewed locally. I did not find any blocking issues. The reduced workflow permissions look consistent with the remaining GitHub API, issue/PR, artifact, and branch-push operations, and disabling persisted checkout credentials in the dependency test jobs does not appear to break later steps.

Non-blocking note: failure reports now point to the workflow run instead of the exact matrix job URL after removing the gh run lookup. That is a triage-quality tradeoff rather than a correctness problem.

Checked with:\n- git diff --check origin/master...HEAD\n- actionlint on the changed workflows\n- search for remaining GITHUB_TOKEN/GH_TOKEN usage

PR #7356, opened 2026-05-20
Fixes argument injection in the interactive contribute task.

matneu:
ContributionTask now invokes external commands using the existing executable-plus-argv `GeneralUtils.invokeCommand(...)` overload instead of building space-delimited command strings. This keeps user-provided values such as coordinates and branch names as single arguments when  passed to Gradle `ExecOperations.exec()`.

 Also removes the unsafe `GeneralUtils.invokeCommand(String command, ...)` overload that used `command.split(" ")`, preventing future callers from reintroducing the same issue.


jormundur00 (approved):
No blocking findings. I reviewed the command construction changes locally and verified that build-logic tests pass with:\n\n`./gradlew -p tests/tck-build-logic test --stacktrace`\n\n`git diff --check origin/master...HEAD` also passes.

PR #7360, opened 2026-05-20
Fix GitHub Actions code injection risk

matneu:
The workflow previously interpolated issue-derived Maven coordinates directly into shell scripts and `actions/github-script` JavaScript source. This patch moves those values through environment variables and reads them at runtime instead.

 ## Changes

  - Pass extracted coordinates via `env: COORDS` for shell and `github-script` steps.
  - Read coordinates with `process.env.COORDS` inside JavaScript blocks.
  - Tighten Maven coordinate extraction and validation to:
    `^[A-Za-z0-9._-]+:[A-Za-z0-9._-]+:[A-Za-z0-9._-]+$`
  - Keep `${{ steps.extract.outputs.coords }}` only in job outputs or environment mappings, not executable script bodies.

### DIFF
--- a/.github/workflows/triage-new-issues.yml
+++ b/.github/workflows/triage-new-issues.yml
@@ -33,7 +33,7 @@ jobs:
             const hasLibraryNewRequestLabel = labelNames.includes("library-new-request");
 
             function extractCoordinatesFromTitle(title) {
-              const match = String(title || "").match(/([^:\s]+:[^:\s]+:[^\s]+)/);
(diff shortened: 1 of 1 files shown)


jormundur00 (approved):
No blocking findings. I reviewed the workflow hardening locally: checked the before/after diff, verified issue-derived coordinates are no longer interpolated into executable script bodies, exercised the stricter coordinate regex against valid and injection-shaped titles, checked existing repo coordinates against the regex, parsed the workflow YAML, and ran `git diff --check origin/master...HEAD`.\n\nNote: I did not run actionlint locally because it is not installed here; the GitHub Actions/Macaron check is passing in CI.

PR #7486, opened 2026-05-21
Fix scheduled release output size

### LINKED ISSUE #7485: Scheduled release workflow exceeds GitHub Actions output limit
The scheduled release workflow can fail in the `Get a list of changed metadata` job when the generated changed-metadata matrix is large enough to exceed GitHub Actions' per-job output limit.\n\nThe Gradle task itself succeeds, but the job fails while evaluating outputs because it publishes the full expanded matrix as a job output. The release job only needs the `none-found` boolean, so the matrix should not be exported from this workflow.

jormundur00:
## Summary
- Avoid publishing the expanded changed-metadata matrix as a job output in the scheduled release workflow.
- Keep using the existing Gradle task to detect changed metadata, but forward only the `none-found` output that the release job consumes.
- Prevent large release diffs from hitting the GitHub Actions per-job output size limit.

## Verification
- Reproduced the failing release range locally and confirmed the workflow step now writes only `none-found=false` to the real GitHub output file.
- Ran `git diff --check`.

Fixes #7485

### DIFF
--- a/.github/workflows/create-scheduled-release.yml
+++ b/.github/workflows/create-scheduled-release.yml
@@ -65,7 +65,6 @@ jobs:
     runs-on: "ubuntu-22.04"
     timeout-minutes: 5
     outputs:
-      matrix: ${{ steps.set-matrix.outputs.matrix }}
       none-found: ${{ steps.set-matrix.outputs.none-found }}
     steps:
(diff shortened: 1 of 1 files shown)


kimeta (approved):
LGTM

PR #7745, opened 2026-06-03
Split Forge local verification into generation and pre-publication tiers

### LINKED ISSUE #7743: Forge local CI verification: default `smoke` mode (skip native test matrix + Spring AOT) with progress logging
### Reduce gate

The Forge pre-PR gate `run_local_ci_verification` (`forge/utility_scripts/local_ci_verification.py`) faithfully reproduces every per-PR CI gate locally before opening a generated PR. The expensive gates make it impractically slow for the common case and for programmatic callers that just need a fast sanity check:

- the test matrix** (`_run_test_matrix_entries` → `run-consecutive-tests.sh`): for every changed `(coordinates, version)` it runs `./gradlew clean test` (full native-image build + run), looping over up to 30 versions per chunk, expanded across the `ci.json` cartesian matrix (currently `java=[25, latest-ea] × os=[ubuntu-latest] × nativeImageMode=[current-defaults, future-defaults-all]` = 4 entries per chunk), with up to 3 retries on the native run. This is where the hours go.
-  Spring AOT smoke tests** (`_run_spring_aot_verification`): clones `spring-aot-smoke-tests` and runs triaged native tests when metadata affects a Spring AOT project. Also expensive.

The cheap gates (matrix generation, `checkMetadataFiles`, index validation, checkstyle, library-stats, doc-links, docker-scan) take seconds and should always run.

### Add progress logging

There is currently no progress output, so during a long run you cannot tell which gate or which `(coordinates, version, jdk, mode)` entry is executing. We should add structured progress logging.

kimeta:
Fixes #7743.

Forge generated PRs previously reproduced the full native test matrix and Spring AOT lanes during pre-publication verification, and re-ran metadata/style/legacy checks that finalization had already established. This restructures local verification into two tiers, split along the boundary between what a single-library generation can settle on its own and what it cannot.

### Generation/finalization tier (FS-local-ci-equivalent-verification.1)
Owns library-scoped correctness. Runs the generated tests under current defaults and `future-defaults-all` on the latest GraalVM, **plus a new current-defaults lane on the GraalVM 25 toolchain** (`GRAALVM_HOME_25_0`), each wrapped by the existing Codex/Pi fixers. Finalization continues to validate metadata (`checkMetadataFiles`), derive `allowed-packages`, apply/check style, regenerate stats, and reject legacy test-only Native Image config.

### DIFF
--- a/forge/ai_workflows/core/workflow_strategy.py
+++ b/forge/ai_workflows/core/workflow_strategy.py
@@ -19,7 +19,7 @@ from ai_workflows.core.fix_post_generation_pi import (
     run_pi_post_generation_fix,
 )
 from utility_scripts.library_finalization import run_library_finalization
-from utility_scripts.workflow_setup import build_graalvm_environment
+from utility_scripts.workflow_setup import build_graalvm_environment, require_graalvm_home_env
 from utility_scripts.gradle_environment import gradle_command_environment
 from utility_scripts.gradle_test_runner import run_gradle_test_command
 from utility_scripts.library_stats import stats_artifact_dir
@@ -282,9 +282,8 @@ class WorkflowStrategy(ABC):
         # generation lanes with the same metadata/Pi fixers, because the
         # pre-publication gate (§FS-local-ci-equivalent-verification.2) no longer
         # reproduces the native test matrix.
-        # GRAALVM_HOME_25_0 presence is already preflighted at startup
-        # (forge_metadata POST_GENERATION_GRAALVM_ENV_VAR), so read it directly.
-        current_defaults_25_env = build_graalvm_environment(os.environ["GRAALVM_HOME_25_0"])
+        graalvm_home_25 = require_graalvm_home_env("GRAALVM_HOME_25_0")
+        current_defaults_25_env = build_graalvm_environment(graalvm_home_25)
         current_defaults_25_env.pop("GVM_TCK_NATIVE_IMAGE_MODE", None)
         current_defaults_25_status = run_lane(
             "current-defaults GraalVM 25 test",
--- a/forge/ai_workflows/core/workflow_strategy.py
+++ b/forge/ai_workflows/core/workflow_strategy.py
@@ -19,6 +19,7 @@ from ai_workflows.core.fix_post_generation_pi import (
     run_pi_post_generation_fix,
 )
 from utility_scripts.library_finalization import run_library_finalization
+from utility_scripts.workflow_setup import build_graalvm_environment
 from utility_scripts.gradle_environment import gradle_command_environment
 from utility_scripts.gradle_test_runner import run_gradle_test_command
 from utility_scripts.library_stats import stats_artifact_dir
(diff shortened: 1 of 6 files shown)


jormundur00 (changes requested):
Reviewed locally. This splits Forge verification into the new generation/finalization tier and a smaller pre-publication gate. The general direction looks sound, but I found two issues.

1. The reduced pre-publication retry can accept library-scoped fixups without revalidating them.

`run_local_ci_verification()` retries `_run_verification_once()` after `_run_fixup()` commits Codex changes. `_run_fixup()` can edit generated library files or shared repo files, but the default `_run_verification_once()` now only reruns the reduced pre-publication gate. If a fixup touches `metadata/<group>/<artifact>/`, `tests/src/<group>/<artifact>/`, or `stats/<group>/<artifact>/`, the earlier generation/finalization result is stale, yet Forge can still publish after only index/Docker gate success.

A good fix would be: after any local-CI fixup touches the target library's metadata, tests, or stats paths, rerun the library-scoped finalization checks for that coordinate before accepting the retry. That keeps the fast gate for pure cross-cutting fixes while preserving the two-tier verification contract when a fixup changes files owned by the generation/finalization tier.

2. Direct workflow-driver runs can now crash with a raw `KeyError` when `GRAALVM_HOME_25_0` is missing.

`_run_test_with_retry()` reads `os.environ["GRAALVM_HOME_25_0"]` directly. That env var is preflighted through `forge_metadata.py`, but the workflow drivers are also documented as direct entry points. Direct driver runs can now pass the first two lanes and then fail with `KeyError: 'GRAALVM_HOME_25_0'`. This should either be validated in the direct driver setup path as well, or `_run_test_with_retry()` should fail with the same clear `require_graalvm_home_env()` style error.

Validation run:
- `git diff --check origin/master...HEAD`
- `grund check`
- `compileall` on touched Python files
- `pytest tests/test_local_ci_verification.py tests/test_workflow_strategy.py`: 19 passed
- full `forge/tests`: 318 passed, 3 failed

The 3 full-suite failures are existing `test_forge_metadata.py` issue-claim/preflight failures that reproduce on local `master`, so I do not consider them PR regressions.

kimeta:
Thanks, I agree with the second point. We should not rely on `forge_metadata.py`
having preflighted `GRAALVM_HOME_25_0`, because workflow drivers can be invoked
directly. I will add an explicit check before the GraalVM 25 current-defaults
lane runs, so a missing or invalid `GRAALVM_HOME_25_0` fails with a clear error
instead of a raw environment lookup failure.

On the first point, I do not think we should rerun finalization inside each
post-generation recovery lane. The current order is already:

1. generate metadata
2. run the post-generation test lanes, including Codex/Pi recovery when needed
3. run shared library finalization
4. commit the iteration

So any files changed by Codex/Pi recovery are still finalized before the
iteration is committed.

There is a narrower case where a recovery made for one lane, for example
GraalVM 25 current-defaults, could theoretically break an earlier lane such as
latest current-defaults after that earlier lane has already passed. I think that
case should remain visible instead of being hidden by more local retry/finalize
cycles. It should produce a PR that clearly shows the compatibility issue, which
is the kind of signal we may need to report to Native Image. Re-running
finalization and prior lanes after every recovery would slow down the normal
generation path for a rare case and make this failure mode less explicit.

jormundur00 (changes requested):
Thanks for the update. The `GRAALVM_HOME_25_0` implementation concern is addressed: using `require_graalvm_home_env()` gives direct driver runs a clear failure mode instead of a raw `KeyError`.

I still think the local-CI fixup concern remains. My point was not about Codex/Pi recovery inside post-generation test lanes; I agree those edits are followed by shared finalization. The gap is the later pre-publication path: when the reduced pre-publication gate fails, `run_local_ci_verification()` may run `_run_fixup()`, commit Codex changes, and then retry only the reduced `_run_verification_once()` gate. Since `_run_fixup()` may edit generated library files, a fixup touching `metadata/<group>/<artifact>/`, `tests/src/<group>/<artifact>/`, or `stats/<group>/<artifact>/` can be accepted without re-running the library-scoped finalization checks for those changed files.

I think this needs a safeguard: after a local-CI fixup touches target library paths, rerun the library-scoped finalization checks for that coordinate before accepting the retry; or restrict local-CI fixups to cross-cutting files only.

One new test issue: `tests/test_workflow_strategy.py::WorkflowStrategyTests::test_post_generation_tests_run_latest_and_graalvm_25_current_defaults_lanes` now fails locally. The test sets `GRAALVM_HOME_25_0` to `/dev/graalvm-25`, but `require_graalvm_home_env()` also requires `<home>/bin/native-image`. The test should create a temp fake GraalVM home with `bin/native-image` or mock the helper.

Validation:
- `git diff --check origin/master...HEAD`: passed
- `grund check`: passed
- `compileall`: passed
- `pytest tests/test_local_ci_verification.py tests/test_workflow_strategy.py`: 18 passed, 1 failed

kimeta:
Thanks, I agree that the fixup contract should stay narrow here. I removed the
line that said local-CI fixup may update generated library files or shared
repository files, because the reduced local-CI gate should not invite broad
post-finalization edits.

For this path, I do not think we need to rerun library finalization after a
local-CI fixup. We intentionally minimized local-CI verification to the
cross-cutting pre-publication gates: changed `index.json` validation and the
allowed Docker image scan. Fixes for those gates should be minimal and targeted
to the failing gate, and they should not change the semantics of the generated
library test/native-image run.

I also moved the `GRAALVM_HOME_25_0` validation into
`resolve_graalvm_java_home()`, so direct workflow-driver runs now preflight the
GraalVM 25 toolchain at the same setup point that already resolves the ambient
`GRAALVM_HOME`/`JAVA_HOME`. The GraalVM 25 lane can then keep using the
preflighted environment value directly.

jormundur00 (approved):
Thanks, this addresses my blocking concerns. The GraalVM 25 setup path now preflights through resolve_graalvm_java_home(), and the focused workflow_strategy test issue is fixed.

I’m okay with the reduced local-CI fixup contract here: library-scoped edits from that retry can land in the generated PR and be validated by repository CI. Re-reviewed locally: diff check, grund check, compileall, and the focused pytest run passed. No further blocking findings.

PR #8196, opened 2026-06-08
Maintain latest metadata release and weekly cadence

### LINKED ISSUE #2109: Maintain a continuously updated latest release for metadata artifacts
## Problem
Metadata changes can be merged at any time, but published releases are currently created on a schedule. That means the repository can contain newer merged metadata than what is available through the most recent release download.

This creates unnecessary lag for users and automation that consume release artifacts instead of cloning the repository. Anyone depending on a stable latest release endpoint cannot reliably fetch the newest metadata right after a merge.

## Proposed enhancement
Add workflow support for a moving `latest` release/tag that is refreshed whenever metadata changes are merged into the main repository.

## Why this is needed
- Newly merged metadata should be available through the release channel immediately, not only on the next scheduled release date.
- Consumers should have a stable release name/tag that always points to the newest published metadata bundle.
- Automation built around release downloads becomes simpler because it no longer needs to discover the newest versioned tag or wait for the next scheduled release.

## Implementation note
The workflow should handle bootstrapping cleanly by ensuring an initial latest release/tag exists before any update logic tries to replace a previous one.

jormundur00:
Fixes #2109.

## Summary
- Add a `create-snapshot-release.yml` workflow that publishes a floating `SNAPSHOT` GitHub Release on the `SNAPSHOT` tag after `master` pushes when metadata changed since the previous `SNAPSHOT` tag.
- Bootstrap the first snapshot release from the most recent numbered release tag, then use the existing `SNAPSHOT` tag as the diff base for subsequent refreshes.
- Keep scheduled numbered releases independent from floating snapshot tags by selecting only semantic-version tags when diffing and computing the next numbered release.
- Change the numbered scheduled release cadence from twice per month to every Monday at 03:00 UTC so it can align with downstream repository release automation.
- Add/update Grund-backed CI and functional specs, plus the architecture overview, to document both the floating `SNAPSHOT` release and weekly numbered release cadence.

## Notes
- The snapshot release packages the repository with version `SNAPSHOT`, replaces the previous `SNAPSHOT` release/tag, publishes with title `SNAPSHOT`, and uses `--latest=false` so the newest numbered release keeps GitHub's Latest marker.
- Numbered releases continue to run `spotlessCheck`, package metadata, and require the latest completed `test-all-metadata` run to have passed unless manually dispatched.

## Verification
- `grund check`
- `git diff --check`
- Checked for stale twice-monthly/bi-weekly scheduled-release wording in the touched docs and workflows.

### DIFF
--- a/.github/workflows/create-snapshot-release.yml
+++ b/.github/workflows/create-snapshot-release.yml
@@ -73,7 +73,7 @@ jobs:
 
   release:
     needs: get-changed-metadata
-    if: needs.get-changed-metadata.result == 'success' && needs.get-changed-metadata.outputs.none-found != 'true'
+    if: needs.get-changed-metadata.result == 'success' && (needs.get-changed-metadata.outputs.none-found != 'true' || (needs.get-changed-metadata.outputs.snapshot-tag-exists == 'true' && needs.get-changed-metadata.outputs.snapshot-release-exists != 'true'))
     name: "🚀 Create snapshot release"
     runs-on: "ubuntu-22.04"
     env:
--- a/.github/workflows/create-snapshot-release.yml
+++ b/.github/workflows/create-snapshot-release.yml
@@ -11,6 +11,10 @@ on:
 permissions:
   contents: write
 
+concurrency:
+  group: create-snapshot-release
+  cancel-in-progress: true
+
 jobs:
   get-changed-metadata:
     name: "📋 Get a list of changed metadata"
--- a/.github/workflows/create-snapshot-release.yml
+++ b/.github/workflows/create-snapshot-release.yml
@@ -1,6 +1,6 @@
-# Latest release: packages metadata after master updates and publishes it on the
-# floating latest tag. §CI-create-latest-release; §FS-repository-functional-spec.4.4.
-name: "Create latest release"
+# Snapshot release: packages metadata after master updates and publishes it on the
+# floating latest tag. §CI-create-snapshot-release; §FS-repository-functional-spec.4.4.
+name: "Create snapshot release"
 
 on:
   push:
@@ -20,7 +20,7 @@ jobs:
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
       none-found: ${{ steps.set-matrix.outputs.none-found }}
-      latest-release-exists: ${{ steps.get-base-ref.outputs.latest-release-exists }}
+      snapshot-release-exists: ${{ steps.get-base-ref.outputs.snapshot-release-exists }}
     steps:
       - name: "☁️ Checkout repository"
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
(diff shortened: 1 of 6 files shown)


kimeta:
High risk: the new snapshot workflow needs a concurrency guard before merge.

Every run mutates the same global `SNAPSHOT` release/tag: it may delete the existing release/tag and then force-push `refs/tags/SNAPSHOT`. Because the workflow triggers on every `master` push, two close pushes can run concurrently. In that case an older run can delete or overwrite the snapshot produced by a newer run, leaving the floating snapshot release stale or missing.

Please add a workflow-level concurrency group for snapshot publishing, for example with a stable group such as `create-snapshot-release` and `cancel-in-progress: true`, so only the newest snapshot publication can update the shared `SNAPSHOT` tag/release.

jvukicev:
Thanks @kimeta, I agree with this. The snapshot workflow mutates a shared floating tag/release, so overlapping runs should be guarded.

I pushed a fix in f094849dc2 adding a workflow-level concurrency group with `cancel-in-progress: true`.

.github/workflows/create-snapshot-release.yml
```diff
@@ -0,0 +1,112 @@
...
+          ./gradlew spotlessCheck
+
+      - name: "🏭 Generate release artifacts"
+        run: |
+          ./gradlew package
+
+      - name: "🧹 Delete previous snapshot release and tag"
+        if: needs.get-changed-metadata.outputs.snapshot-release-exists == 'true'
+        run: |
+          gh release delete SNAPSHOT --cleanup-tag -y
```
kimeta, line 108: High risk: this recovery path assumes that a `SNAPSHOT` tag always has a matching GitHub Release. The discovery step sets `snapshot-release-exists=true` from `git rev-parse SNAPSHOT`, but this step then runs `gh release delete SNAPSHOT --cleanup-tag -y`. If a previous run is interrupted after force-pushing `refs/tags/SNAPSHOT` and before `gh release create`, or if release creation fails before the release is created, the repository is left with a `SNAPSHOT` tag but no release. The next run will take the tag as its base, reach this delete step, fail because the release does not exist, and never get to recreate the snapshot.

Please make this cleanup idempotent or track tag existence separately from release existence. For example, delete the release only when `gh release view SNAPSHOT` succeeds, and separately clean up an orphan remote tag before continuing to the recreate path.

jvukicev: Thanks, agreed. I pushed 9f6fabddcb to split `SNAPSHOT` tag detection from GitHub Release detection.

The workflow now uses `gh release view SNAPSHOT` for `snapshot-release-exists`, keeps `snapshot-tag-exists` separate, deletes the release only when it exists, and deletes an orphan `SNAPSHOT` tag separately before recreating the snapshot.

kimeta (changes requested):
Blocking issue in `.github/workflows/create-snapshot-release.yml`:

The orphan-tag cleanup is only reachable when the `release` job runs, but that job is gated on `none-found != 'true'`.

If a previous snapshot run is canceled or fails after pushing `refs/tags/SNAPSHOT` but before `gh release create SNAPSHOT`, the repository is left with a `SNAPSHOT` tag and no `SNAPSHOT` release. The next run then uses that tag as `BASE_COMMIT`; if there are no metadata changes since that tag, `generateChangedCoordinatesMatrix` reports `none-found=true`, the `release` job is skipped, and the orphan cleanup step never runs. That can leave the floating snapshot release missing indefinitely until another metadata-changing push happens.

Please make the release/repair path run when `snapshot-tag-exists == 'true' && snapshot-release-exists != 'true'`, or treat an orphan tag as needing publication before applying the `none-found` skip.

jvukicev:
Thanks, this is valid too. I pushed 79de8a928a so the release/repair job also runs when `snapshot-tag-exists == true` and `snapshot-release-exists != true`, even if the changed-metadata matrix reports `none-found=true`.

That means the orphan-tag repair path is reachable immediately after a failed/canceled snapshot publication, instead of waiting for a later metadata-changing push.

kimeta (approved):
The orphan `SNAPSHOT` tag case is fixed: the release/repair job now also runs when the tag exists but the release does not, even if the changed-metadata matrix reports `none-found=true`.

I rechecked the workflow and docs, and ran `grund check` plus `git diff --check` against the updated head. CI is green as well.

Non-blocking follow-up: `create-snapshot-release.yml` still exports the full `matrix` job output even though the release job only uses `none-found`, while `create-scheduled-release.yml` intentionally captures the Gradle output in a temp file and forwards only `none-found`. It would be cleaner to mirror that pattern here, but I do not think it should block this PR.

PR #8299, opened 2026-06-10
Improve test-all-metadata failure reporting

### LINKED ISSUE #8298: Stop opening issues with test-all-metadata.yml workflow runs
`test-all-metadata.yml` currently opens issues for every failure. We don't want this as, more often than not, these issues are GraalVM regressions that will get picked up by our `forge/` workflows and waste resources generating an unfixable problem.

A better solution would be to stop generating these issues, but improve the logging in the workflow itself. Currently, the "All metadata tests have passed job" just fails (but the other batched jobs pass), so inspecting the action itself is tedious. We should make it so the workflow keeps artifacts of failed tests (also making the batch jobs fail in the process), and then fail the aggregate job if any of the batch jobs failed.

Lastly, we should reduce the cadence of these tests to run once a week, on sunday, so it runs one day before attempting to create a scheduled release (preventing regressions). As it is now, most of our github actions time is spent on running this workflow, so we should reduce our resource usage.

jormundur00:
## Summary
- reduce test-all-metadata to a weekly Sunday schedule
- stop opening GitHub issues for full-sweep metadata failures
- upload failure-log and aggregate-result artifacts, then fail affected matrix batches

Fixes #8298

## Testing
- git diff --check
- grund check
- ./gradlew spotlessCheck --stacktrace

### DIFF
--- a/.github/workflows/test-all-metadata.yml
+++ b/.github/workflows/test-all-metadata.yml
@@ -1,11 +1,11 @@
-# Scheduled full metadata sweep (every 3 days): the 85-batch matrix from
-# generateMatrixBatchedCoordinates (§TCK-test-harness.7) runs the full test lane and files one
-# issue per failed (library, version). §CI-test-all-metadata; §FS-repository-functional-spec.5.3.
+# Weekly Sunday metadata sweep: 85 batches from generateMatrixBatchedCoordinates
+# (§TCK-test-harness.7) run tests, upload failure artifacts, and fail affected batches.
+# §CI-test-all-metadata; §FS-repository-functional-spec.5.3.
 name: "Test all metadata (one version per test suite)"
 
 on:
   schedule:
-    - cron: "0 2 */3 * *"
+    - cron: "0 2 * * 0"
   workflow_dispatch:
 
 permissions:
@@ -77,6 +77,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: "🧪 Run '${{ matrix.coordinates }}' tests"
+        id: run-tests
         run: |
           set -euo pipefail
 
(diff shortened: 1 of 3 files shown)


kimeta:
Non-blocking: removing the per-library `fails-*` issue creation makes sense because those labels feed Forge and many full-sweep failures are likely GraalVM regressions rather than actionable metadata repairs. I do think artifacts-only may be hard to track over time, though.

Could we add one aggregate tracking issue for a failed `test-all-metadata` run? For example, the `process-results` job could open a single aggregated issue titled like `[Tracking] test-all-metadata failures from <run date>` with the workflow run URL, aggregate artifact name, and a table of failed `group:artifact:version`, failure type, and matrix environment. 

Another option is still to create per-library issues, but use a non-Forge label/title pattern and no `fails-*` classification labels. That would preserve searchability per coordinate while avoiding automated Forge execution.

jvukicev:
> Non-blocking: removing the per-library `fails-*` issue creation makes sense because those labels feed Forge and many full-sweep failures are likely GraalVM regressions rather than actionable metadata repairs. I do think artifacts-only may be hard to track over time, though.
> 
> Could we add one aggregate tracking issue for a failed `test-all-metadata` run? For example, the `process-results` job could open a single aggregated issue titled like `[Tracking] test-all-metadata failures from <run date>` with the workflow run URL, aggregate artifact name, and a table of failed `group:artifact:version`, failure type, and matrix environment.
> 
> Another option is still to create per-library issues, but use a non-Forge label/title pattern and no `fails-*` classification labels. That would preserve searchability per coordinate while avoiding automated Forge execution.

I don't think creating aggregate issues is a good solution, as that way there is no way to reliably map "already open" issues. If there is a fail anywhere, every workflow run would open one issue, which will become cluttered if they are not fixed fast (and with Graal release waiting, more often than not a fix won't be present in between runs).

kimeta (approved):

PR #8439, opened 2026-06-12
Split library-update tested-version aliases

### LINKED ISSUE #8438: Implement library-update tested-version alias splitting
Library-update PRs can generate tests for one metadata version while the same index entry still represents later tested-version aliases. If the generated JVM tests stop being compatible with a later alias, local publication should preserve progress for the passing prefix and split the successor range instead of waiting for remote CI to fail.

This issue tracks implementing the Forge publication behavior for that case:

- sweep tested-version aliases with Java tests before PR eligibility for library-update publications
- split the index entry at the first later alias failure
- copy baseline metadata and tests for the successor range
- create and park the successor library-update issue until the current PR merges
- release that follow-up issue to Todo after the splitting PR merges

kimeta:
## Motivation

A `library-update` coverage-improvement run regenerates JVM tests for a metadata
entry that often covers **several tested-version aliases** at once (e.g.
`["1.1", "1.2", "1.3"]`). The freshly generated tests can pass on the entry's own
version yet quietly stop compiling or running against a *later* alias — and today
nothing catches that before the PR becomes eligible, so the entry ships claiming
support it no longer has.

This PR teaches Forge to detect that break and **split the entry at the first
alias that fails**, keeping the generated progress for the passing prefix while
preserving the repo's existing baseline support for everything after it.

## What it does

Before a changed-tests library-update branch becomes PR-eligible, Forge runs a
**Java-only alias sweep** — `javaTest` with `GVM_TCK_LV` set to each alias in
index order, stopping at the first failure. This is deliberately narrower than
full CI (no native-image matrix) since it only needs to catch JVM test
incompatibility. On the first later-alias failure it splits the index entry:

| | `metadata-version` | `tested-versions` | `latest` | source |
|---|---|---|---|---|
| **Current entry** | unchanged | passing prefix | dropped if it moves | generated progress (this PR) |
| **Successor entry** | first failing alias | failing alias + all later | inherited if split entry had it | baseline metadata/tests copied from the PR base commit |

It then opens a parked `library-update-request` issue for the successor range,
holds it in **In Progress** so the queue can't claim it early, and references it
from the PR via a machine-readable `Forge-Unblocks-Issue:` trailer. Once this PR
merges, Forge releases that issue to **Todo** for normal processing.

If the *first* alias fails there is no passing prefix to keep, so publication
fails instead of splitting.

## Spec & wiring

- New behavior contract: §FS-library-update-tested-version-split
- PR-body trailer / follow-up linking: §GIT-pr-body
- Hooks into publication through a new `before_verification` step in
  `publish_branch`, so the split lands before local CI decides eligibility —
  one code path for "verified diff → pushed branch" is preserved.

Fixes #8438

### DIFF
--- a/forge/docs/functional-spec.md
+++ b/forge/docs/functional-spec.md
@@ -317,6 +317,41 @@ paths. If any shared repository file changed, the PR must be labeled
 the repository-level paths that require maintainer review, following
 §FS-human-intervention-policy.
 
+### FS-library-update-tested-version-split: Library-update tested-version split
+
+For a `library-update-request` publication that changes generated tests for an
+existing metadata entry, Forge must run a Java-only compatibility sweep before
+the branch becomes PR-eligible (§FS-local-ci-equivalent-verification). The sweep
+executes `javaTest` for the changed metadata coordinate with `GVM_TCK_LV` set to
+each tested-version alias in the index entry, in index order, and stops at the
+first failing alias. This is intentionally narrower than full CI: it catches
+generated JVM test code that is incompatible with later aliases without paying
+the native-image matrix cost locally.
+
+If the first alias fails, Forge must fail publication instead of splitting,
+because there is no passing prefix for the current update. If a later alias
+fails, Forge must split the index entry at that first failing alias. The
+generated metadata version keeps only the passing prefix. A successor entry uses
+the failing alias as its `metadata-version`, preserves that alias and all later
+aliases as its `tested-versions`, and receives `latest: true` when the split
+entry previously carried `latest: true`.
+
+The successor entry must preserve the repository's pre-generation support for
+the failing range. Forge must copy the metadata directory and test directory
+from the PR base commit entry that originally covered the failing alias, using
+that original entry's `metadata-version` and `test-version` if present, into
+`metadata/<group>/<artifact>/<failing-version>` and
+`tests/src/<group>/<artifact>/<failing-version>`. The current PR then keeps the
+new generated progress for the passing prefix while retaining baseline support
+for the successor range.
+
+When Forge creates such a split, it must also create a follow-up
+`library-update-request` issue for the successor metadata version and place that
+issue in `In Progress`. The current PR must reference the follow-up issue but
+must not close it. After the current PR merges, Forge releases the follow-up
+issue by clearing assignees and moving its project status to `Todo`, making the
+successor update eligible for normal processing.
+
 ### FS-human-intervention-policy: Human intervention policy
 
 The `human-intervention` label is a maintainer follow-up signal, not a generic
--- a/forge/docs/functional-spec.md
+++ b/forge/docs/functional-spec.md
@@ -319,39 +319,40 @@ the repository-level paths that require maintainer review, following
 
 ### FS-library-update-tested-version-split: Library-update tested-version split
 
-A `library-update-request` entry often covers several tested-version aliases at
-once (for example `["1.1", "1.2", "1.3"]`). When a coverage-improvement run
-regenerates the JVM tests for such an entry, the new tests can pass on the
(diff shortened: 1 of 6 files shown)


forge/docs/functional-spec.md
```diff
@@ -320,6 +317,46 @@ paths. If any shared repository file changed, the PR must be labeled
 the repository-level paths that require maintainer review, following
 §FS-human-intervention-policy.
 
+### FS-library-update-tested-version-split: Library-update tested-version split
+
+For a `library-update-request` publication that changes generated tests for an
```
vjovanov: Can this be in more direct speech? I am strugling to understand.

vjovanov: PR description is much cleaner.

kimeta: I agree that the spec changes were generally vague. Changed the wording to make it clearer.

jormundur00 (changes requested):
Thanks for the changes. I still see one blocking issue.

`ensure_alias_split_follow_up_issue()` searches for an existing open issue by `successor_coordinates`, but it does not exclude the currently claimed issue. If the original update issue is for the alias that becomes the successor side of the split, `_find_existing_open_issue_number()` can return that same issue instead of creating a distinct follow-up.

For example, if the claimed issue is for `group:artifact:1.3`, the entry has tested versions `[1.1, 1.2, 1.3, 1.4]`, and the alias sweep first fails at `1.3`, then `successor_coordinates` is also `group:artifact:1.3`. The PR body would then contain both `Fixes: #<current>` and `Forge-Unblocks-Issue: #<current>`, so merging closes the issue that Forge is supposed to keep parked and release to `Todo` for the successor work.

Please make the follow-up issue lookup exclude `current_issue_number` or otherwise guarantee that alias splits always reference a distinct successor issue.

kimeta:
Replying to the review above: I dug through the target-resolution path and I think the issue is narrower than the example suggests.

For the normal `library-update-request` case where the requested version is only present as a tested-version alias, Forge splits before generation. Example:

```json
{
  "metadata-version": "1.1",
  "tested-versions": ["1.1", "1.2", "1.3", "1.4"]
}
```

If the claimed issue is for `group:artifact:1.3`, `prepare_library_update_target()` first splits that into roughly:

```json
{
  "metadata-version": "1.1",
  "tested-versions": ["1.1", "1.2"]
},
{
  "metadata-version": "1.3",
  "tested-versions": ["1.3", "1.4"]
}
```

So the alias sweep for the generated target sees only `["1.3", "1.4"]`. If `1.3` fails, it is the first alias and publication fails because there is no passing prefix to keep. If `1.4` fails, the successor issue is for `group:artifact:1.4`, which is distinct from the current issue.

The same is true for a correctly anchored existing entry like:

```json
{
  "metadata-version": "1.3",
  "tested-versions": ["1.3", "1.4"]
}
```

The problematic case is only when the index bucket is already misordered, for example:

```json
{
  "metadata-version": "1.3",
  "tested-versions": ["1.1", "1.2", "1.3", "1.4"]
}
```

Then the sweep can pass `1.1` and `1.2`, fail at `1.3`, and produce `successor_coordinates == group:artifact:1.3`, which is also the claimed issue coordinate.

We do have a few rare current cases with this shape, mostly qualifier/snapshot/version-alias cases introduced by bulk tested-version updates. They should not exist under the intended index structure: the metadata-version should be the bucket anchor, and bulk compatibility testing should not place tested-version aliases before that anchor. I opened #8511 to track that bulk-update bug separately.

jormundur00 (approved):
Re-reviewed locally. The previous wording/spec concern is addressed, and I did not find another blocker.

The self-follow-up edge case still exists: a misordered tested-version range can still make the successor coordinate match the current issue and record the current issue as its own follow-up. Per the discussion, this is accepted here and will be handled in a follow-up.

PR #8454, opened 2026-06-12
Add metadata for net.sf.jasperreports:jasperreports and com.github.librepdf:openpdf

keepprogress:
## What does this PR do?

Adds reachability metadata and TCK tests for two related PDF/reporting libraries:

- **`net.sf.jasperreports:jasperreports:6.21.5`** — covers the headless report lifecycle: programmatic report design, report compilation, compiled-report serialization round-trip (`JRSaver`/`JRLoader`, the mechanism behind precompiled `.jasper` resources), report filling, and PDF export.
  - Reflection: report compilers resolved by name (`JRJdtCompiler`/`JRJavacCompiler`), text-measurer factories, extension-registry factories listed in `jasperreports_extension.properties`, exporter filter factories, the classic PDF producer factory, `NamedEnum.getByName(String)` lookups for PDF export configuration enums, and the dynamic proxies built for `PdfExporterConfiguration`/`PdfReportConfiguration`.
  - Serialization: the complete object graph of a compiled report (entries are conditioned on both `JRSaver` and `JRLoader`, because `typeReached` is evaluated at run time and saving happens before `JRLoader` is ever reached).
  - Resources: `default.jasperreports.properties`, `jasperreports_extension.properties`, and the `jasperreports_messages` bundle.
- **`com.github.librepdf:openpdf:1.3.32`** (the PDF backend JasperReports 6.21.5 depends on) — covers the built-in Type1 "standard 14" fonts: all 14 AFM metric resources, `glyphlist.txt`, and the `error_messages/*.lng` localization catalogs, exercised by creating fonts, writing a PDF document, and resolving a localized message.

Both test projects add `-H:+AddAllCharsets` (same as the existing okhttp/mssql-jdbc/jdom tests) because the built-in Type1 fonts default to the `Cp1252` encoding, which is not in the default native-image charset set and cannot be expressed as reachability metadata.

The report in the JasperReports test only uses direct field/parameter expressions, so the compiled report is evaluated through direct expression evaluation (available since JasperReports 6.20) without any runtime bytecode generation — the only report-compilation mode that can work in a native image. The JRXML (XML template) parsing path and the AWT-based default text measurer are intentionally out of scope: the former needs a much larger reflection surface (a follow-up candidate), and the latter depends on desktop/JNI support that is not library metadata.

The metadata mirrors what we run in production: a Spring Boot 3.5 native image serving real JasperReports PDFs, where these exact registrations were validated on real hardware.

None — both test projects are fully in-memory (`ByteArrayOutputStream`/`ByteArrayInputStream`), with no network, no Docker, and no file-system access.


vjovanov:
Seems that we didn't register something for the JNI metadata and the test fail. @keepprogress would you like to look into it or us?

Issue #8486, opened 2026-06-15
Forge: resume failed runs from the phase that failed (run continuation)


kimeta:
## Motivation

When a Forge run hits a **logical** failure today, it preserves the work branch and
labels the issue `human-intervention` (`§FS-human-intervention-policy`), but any
retry **regenerates the result from scratch** — re-spending tokens and compute on
phases (setup, fix, exploration) that already succeeded, and re-discovering work
that was already committed on the preserved branch.

Run continuation closes that gap: a later run (or a maintainer) re-enters the
workflow **at the phase that failed**, using the preserved branch as the starting
state. This directly serves two existing goals:

- `§GOAL-shorten-issue-to-shipped-metadata`, which already requires Forge to
  *"preserve enough evidence for maintainers or later Forge runs to continue
  without rediscovering the same problem."*
- `§GOAL-minimize-generation-cost`, by not paying again for phases that passed.

Continuation is **additive** to human intervention, not a replacement: a logical
failure still preserves the branch and applies the `human-intervention` label
exactly as today. The marker is extra machine-readable state that rides the same
preserved branch.

## Phase model

A run is an ordered sequence of phases, classified by *shape* (shape decides how it
resumes):

| Phase | Shape | Resume action |
| --- | --- | --- |
| `setup` | discrete | Skip sub-steps already marked done (`preflightDone`, `setupDone`); else rerun. |
| `fix` | continuous | Continue from the preserved branch at the recorded `iteration`. |
| `explore` | continuous | Rerun; the regenerated dynamic-access report self-prunes to uncovered classes, and `exhaustedClasses` prevents retrying abandoned classes. |
| `finalization` | discrete | If entered but not completed, redo from the preserved branch. |
| `publication` | discrete (remote) | If `isPushed`, skip to PR find-or-create on the stored `branch`; else run the full publication pipeline. |

Unifying invariant: **the preserved branch HEAD is the cursor.** The committed tree
is the source of truth, so the marker never stores commit hashes — only the logical
state a rebuild cannot recover.

## Contracts

- **Store only non-reconstructable state.** The dynamic-access coverage report is
  regenerated by rebuilding the tree; covered classes reappear in the regenerated report. The only
  exploration state worth keeping is `exhaustedClasses` (a fresh report cannot tell an
  uncovered-but-abandoned class from an uncovered-but-untried one).
- **Gitignored during the run, force-added only on the preservation branch.** Same
[...]

PR #8487, opened 2026-06-15
Forge: add run-continuation spec (WF-forge-run-continuation)

kimeta:
## What

Spec-first, docs-only PR that adds `WF-forge-run-continuation` — the contract for
**resuming a failed Forge run at the phase that failed** instead of regenerating the
whole result from scratch — and documents the `ContinuationMarker` it relies on.

No behavior change yet. This is step 1 of the plan in
oracle/graalvm-reachability-metadata#8486; writing the spec before the code so the
implementation has a stable ID to cite.

## Why

A logical failure already preserves the work branch and labels the issue
`human-intervention` (`§FS-human-intervention-policy`), but a retry today
regenerates everything — re-spending tokens and compute on phases that already
succeeded. Continuation lets a later run re-enter at the failed phase, which is
exactly what `§GOAL-shorten-issue-to-shipped-metadata` ("later Forge runs continue
without rediscovering the same problem") and `§GOAL-minimize-generation-cost`
already ask for. It is **additive** to human intervention, not a replacement.

## Contents

- `forge/docs/workflows/continuation.md` — new `WF-forge-run-continuation`: the phase
  model (continuous `fix`/`explore` vs discrete `finalization`/`publication`), the
  marker lifecycle (gitignored during the run, force-added onto the preserved branch
  only — the same mechanic as the chunked exhaust report), and the resume flow.
- `forge/docs/functional-spec.md` — one sentence in `§FS-human-intervention-policy`
  linking the preserved branch to the continuation marker.
- `forge/docs/README.md` — index entry for the new workflow doc.

Scope: docs only, three files, no code touched. `grund check` passes.

### Open question

- Separate from the marker, the PR branch today carries the many per-class
  exploration checkpoint commits (publication carries the working `HEAD`, no squash).
  Do we also want a squash step for a single clean PR commit? Orthogonal to
  continuation — flagging for visibility.
- The driver/engine unification (one config-driven driver shell + a composite
  `fix → explore` engine) would make resume cleaner but is a larger refactor; tracked
  separately, not in this spec.

### DIFF
--- a/forge/docs/README.md
+++ b/forge/docs/README.md
@@ -41,6 +41,7 @@ its IDs in that one file.
 - [grund.md](grund.md) — `GRUND-forge-motivation`: why Forge exists.
 - [goals.md](goals.md) — `GOAL-forge-direction` and the outcome goals beneath it.
 - [functional-spec.md](functional-spec.md) — `FS-forge-functional-spec`: top-level functional spec and the workflow-spec catalog.
+- [continuation.md](continuation.md) — `FS-forge-run-continuation`: resuming a failed run at the phase that failed, and the continuation marker contract.
 - [architecture.md](architecture.md) — `AR-forge-architecture`: control plane, workflow boundaries, and extension points.
 - [agent.md](agent.md) — `AR-agent-api`: the agent API and its Pi implementation.
 - [do-work.md](do-work.md) — `DW-do-work-loop`: the long-running worker loop.
--- a/forge/docs/continuation.md
+++ b/forge/docs/continuation.md
@@ -1,0 +1,130 @@
+# FS-forge-run-continuation: Run continuation and resume
+
+Run continuation lets a later Forge run re-enter a failed issue run at the phase
+that failed, instead of regenerating the whole result from scratch. It is a
+cross-cutting capability across the Forge workflow system
+(§WF-forge-workflow-system) and builds on the workflow engine's ownership of run
+state (§WF-forge-workflow-engine): the engine
+already advances per-iteration progress as commits and reverts to the last good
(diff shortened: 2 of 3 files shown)


jormundur00 (changes requested):
Thanks for writing this spec first. I found one implementer-facing contradiction that should be resolved before this becomes the contract.

Finding:

1. `forge/docs/continuation.md`: The marker rules contradict each other about `publication.branch`. The prose says anything rebuildable is omitted and specifically says “the publication branch name is recomputed from the coordinate,” but the example marker stores `publication.branch`, and the field rules say `publication.branch` is stored so resume targets the original branch namespace regardless of which identity resumes. Please pick one contract: either branch is reconstructable and omitted, or it is persisted marker state.

Local validation:

- `git diff --check origin/master...pr/8487`: passed
- `git merge-tree ... origin/master pr/8487`: clean
- `./gradlew --quiet checkDocLinks`: passed
- `./gradlew --quiet spotlessCheck`: passed
- `FS-forge-run-continuation` and cited IDs resolve with `grund`

Full `grund check` is currently blocked by pre-existing outdated `AGENTS.md` grund init blocks on current `origin/master`, not by this PR.

jormundur00 (approved):
Approved. The previous blocker is resolved: publication.branch is now consistently documented as persisted marker state rather than recomputed state. I reran the local docs/style checks and did not find any remaining blocking issues in the committed diff.\n\nNon-blocking cleanup: the PR title/body still mention WF-forge-run-continuation and forge/docs/workflows/continuation.md, while the committed docs use FS-forge-run-continuation in forge/docs/continuation.md.

PR #8494, opened 2026-06-16
Run metadata generation inside throwaway Incus VMs

### LINKED ISSUE #8247: Feature: Support Incus VM for metadata forge
### Is your feature request related to a problem?

Currently it is hard to generate tests as they will open windows, delete files, blow up the /tmp folder, inject stuff to the home folder etc.

### Describe the solution you'd like

We need a config file and setup instructions for Incus so we can test in piece. Docker is not possible as we spawn docker images with 3rd party tools.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

kimeta:
Closes #8247

## What this does

Forge generates and **runs** tests for arbitrary third-party libraries, and that generated code is untrusted — it can open windows, fill `/tmp`, write into `$HOME`, and start its own Docker containers. Running it straight on the operator's machine is risky and messy. This adds an opt-in **`--incus`** mode that runs each issue inside a fresh, single-use virtual machine and destroys it afterward, so nothing a run does can reach the host; behaviour, branches, and outputs are otherwise identical to a normal run. §FS-forge-vm-isolated-execution §AR-forge-vm-runner-boundary §ROADMAP-forge-incus-vm-runner

## How a run works

A base image bakes the expensive, generic setup once; each issue gets a throwaway VM cloned from it copy-on-write, so launches stay cheap and every run starts from an identical clean state:

```
forge-base image  ────►  fresh VM per issue
                                           ├─ seed: gh token, git identity, codex/pi creds
                                           ├─ run the normal per-issue workflow
                                           ├─ stream logs ──► host: forge/logs/issue-<n>/
                                           └─ destroy VM
```

## What's baked vs injected

| Baked into `forge-base` (generic, non-secret) | Injected per run (secret / run-specific) |
|---|---|
| GraalVM, Docker, GitHub CLI, Node + codex/pi, native-image toolchain (gcc), Python package, shallow clone of the local repo | GitHub token, git identity, codex/pi credential files, issue number, strategy, log destination |

The published image stays secret-free; the runner pushes credentials and run-specific state into each VM at launch (§AR-forge-vm-runner-boundary).

## Decisions made during implementation

| Decision | Why it fits |
|---|---|
| **VM, not container** | Untrusted generated code plus nested Docker; a separate guest kernel is a real boundary, not a shared-kernel costume |
| **Fresh, single-use VM per run** | Reproducible, no state bleed between issues; copy-on-write keeps launches cheap |
| **Bake generic tooling, inject secrets at launch** | Published image stays secret-free; credentials rotate without rebuilding it |
| **Bake a shallow copy of the *local* repo** | The VM runs the code under development, not upstream `master`; `--depth 1` avoids copying ~1.2 GB of git history (started as a runtime mount+fetch, simplified to a build-time clone) |
| **Seed only the small codex/pi config files** | The real `~/.codex` / `~/.pi` dirs are ~2.2 GB of session state; only a few KB of auth/provider config is actually needed |
| **Source `/etc/environment` in the exec wrapper** | `incus exec` skips PAM, so the baked GraalVM env vars weren't loaded otherwise |

## Host prerequisites

The operator prepares the host once; `--incus` preflights the essentials and fails fast rather than installing anything. Bringing this up on a real machine surfaced what's needed, now documented in the README:

| Requirement | Why |
|---|---|
| QEMU + KVM, then restart Incus | VMs need a hypervisor and the daemon must detect it |
| `incusbr0` bridge | VM networking (NAT, outbound) |
| Storage pool on a roomy disk | VM disks need tens of GB |
| `DOCKER-USER` accept rules for `incusbr0` | Docker's `FORWARD` default-DROP otherwise blocks the VM's internet |

## Networking (outbound + iptables)

The VMs need only **outbound** traffic — `apt`, `git`, Docker pulls, the LLM gateway — NAT'd through `incusbr0`; no **inbound** is exposed, which is also good for isolation. On a host that also runs Docker, Docker sets the iptables `FORWARD` chain to default-DROP and only allows its own bridge, so the Incus bridge's forwarded packets are silently dropped (DNS resolves, every connection times out). Allowing `incusbr0` through the `DOCKER-USER` chain (`-i`/`-o incusbr0 -j ACCEPT`) fixes it.

## Running it

```bash
./incus/build-base-image.sh           # one-time, per host
python3 forge_metadata.py --run-work-queues --incus
```

### DIFF
--- a/forge/.gitignore
+++ b/forge/.gitignore
@@ -7,6 +7,7 @@ logs/
 fixture-e2e-logs/
 .pending_metrics.json
 .continuation-marker.json
+incus/forge.env
 
 ##############################
 ## Agent state
--- a/forge/README.md
+++ b/forge/README.md
@@ -104,18 +104,70 @@ and fails fast if anything is missing. The runner itself is tracked in
 §ROADMAP-forge-incus-vm-runner.
 §FS-forge-vm-isolated-execution §AR-forge-vm-runner-boundary
 
-### Install and initialize Incus (one-time, on the host)
+### Prepare the host (one-time)
+
+Forge never installs or configures Incus; do this once. Each step is a real
+prerequisite — VM support, networking, storage, and Docker coexistence — and
+`--incus` preflights the most important ones and fails fast if they are missing.
+
+**1. Incus with VM support.** `--incus` launches real virtual machines
+(`incus launch --vm`), which require QEMU and KVM (`/dev/kvm` present; the CPU
+must expose `vmx`/`svm`). Install both Incus and QEMU:
 
 ```console
-# Install Incus from your distribution's packages, then:
-sudo incus admin init --minimal          # storage pool + default network
+sudo apt install incus qemu-system-x86       # package names for Debian/Ubuntu
+# If Incus was already running when QEMU was installed, restart it so the daemon
+# detects the VM backend — otherwise launches fail with
+# "QEMU command not available for CPU architecture":
+sudo systemctl restart incus
+```
+
+**2. Initialize Incus and its bridge network.**
+
+```console
+sudo incus admin init --minimal          # storage pool + managed incusbr0 bridge
 sudo usermod -aG incus-admin "$USER"     # drive Incus without sudo; re-login to apply
 incus list                               # should print an empty table
+```
 
-# Apply the checked-in profile the runner launches every VM with:
+The checked-in profile attaches each VM's NIC to `incusbr0` (the bridge
+`init --minimal` creates); if your bridge has another name, edit the `eth0`
+device in `incus/forge.profile.yaml`.
+
+**3. Apply the Forge profile.**
+
+```console
 incus profile create forge
 incus profile edit forge < incus/forge.profile.yaml
 ```
+
+**4. Give the VMs room on disk.** The base image plus per-run VMs (GraalVM
+installs, Gradle caches, Docker images, native-image scratch) need tens of GB.
+If your root filesystem is small, put the Forge VMs on a roomier disk by creating
+a storage pool there and pointing the `forge` profile's root device at it:
+
+```console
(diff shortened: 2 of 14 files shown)


kimeta:
Review notes from the Incus runner pass:

**Hard problem: host queue accounting counts unclaimed VM attempts as processed**

In `forge_metadata.py`, `--incus` enqueues `dispatch_issue_to_vm(...)` before the host owns the issue claim, then immediately increments `processed_count`. If the inner VM later loses the claim race, finds the project item no longer `Todo`, or otherwise cannot claim the issue, the host still counts that queue item toward the work limit and can stop without processing another claimable issue. This is different from host mode, where `processed_count` increments only after `claim_issue_for_processing(...)` succeeds.

This also conflicts with the new architecture text, which describes the host dispatcher as claiming work before launching the per-run VM. A safer shape is probably: keep host-side claiming and queue-limit semantics unchanged, then invoke an inner VM mode that processes the already-claimed issue/work item without trying to claim it again.

**Medium problem: VM runs the checkout baked into the base image, not the current worker checkout**

The base-image build clones the operator's local repository into `/root/graalvm-reachability-metadata`, and `run_issue_in_vm()` later executes the inner `forge_metadata.py --issue-number N` from that baked checkout. The runner does not refresh that checkout before starting generation; `_seed_credentials()` even notes that there is no checkout to refresh.

That means `do-work.sh` can self-update the host checkout while Incus runs keep using stale Forge code from the last base-image rebuild. The failure mode is subtle: every VM is fresh, but fresh from an old template. This can silently miss fixes to claiming, routing, prompts, publication, metrics, or verification until the operator rebuilds `forge-base`.

The implementation should refresh/reset the VM checkout to the intended current ref before invoking the inner workflow, or mount/copy the current host checkout into the VM per run.

PR #8499, opened 2026-06-16
Evidence-gate human-intervention scanner issue creation

### LINKED ISSUE #8500: Human-intervention scanner opens trackers without current system evidence
The human-intervention scanner currently turns each clustered pattern into a tracker too early. That creates issues for stale groups, already-fixed prompt behavior, one-off PR bugs, or library-specific failures.

It should only open a tracker after a generated investigation task finds current repository evidence for a live system cause. Groups without that evidence should be recorded and cancelled.

kimeta:
Closes #8500

## What this does

The scanner was treating every clustered human-intervention pattern as an issue candidate. That made stale groups and already-fixed prompt problems look like live system work. This changes the template to match §FS-human-intervention-resolution: the scan pass only groups observed causes, then generated investigation tasks search the current checkout and open an issue only when current code, prompt, workflow, or policy evidence still supports a live system cause.

## How issue creation is gated

The workspace now separates the flow into explicit phases:

```text
sleep -> queue-check -> scan -> generated investigation task
                                -> issue-dispatch -> open | cancelled
```

`queue-check` owns the cheap GitHub availability probe. `scan` writes structured group files with `problematic_items` and `common_problems`, drops groups below `min_group_size`, and generates one Rhei task per supported group. The investigation task uses a reusable prompt from `prompt-templates.yaml`, writes a decision-bearing findings file, and only `Decision: open-issue` can reach the opener.

## What changes for scanner users

`min_group_size` defaults to `2`, `investigator_agent` is configurable separately from the scanner and opener, and investigation prompts receive deterministic workspace-rooted paths for group files and issue bodies. Stale, already-fixed, one-off, library-specific, or unsupported groups now cancel without touching GitHub.


jormundur00 (changes requested):
I found one blocking issue in the Rhei template wiring.\n\n configures the  state with , but the Rhei runtime schema I checked only consumes ;  is silently ignored. I instantiated the template locally and rendered a representative generated investigation task with local Rhei, and the task had an empty  field.\n\nThat means the investigation agent never receives the decision contract from , including the required  line and issue-body path. As a result, the evidence gate will not reliably produce the artifacts that  requires. Please either inline this as  in the state, or add actual Rhei support/tests for  if that is intended to be a new runtime feature.

jormundur00 (changes requested):
Correction to my previous review body, which was mangled by shell quoting.

I found one blocking issue in the Rhei template wiring.

forge/.agents/rhei/templates/human-intervention-scanner/states.yaml configures the investigate state with prompt_template, but the Rhei runtime schema I checked only consumes instructions; prompt_template is silently ignored. I instantiated the template locally and rendered a representative generated investigation task with local Rhei, and the task had an empty instructions field.

That means the investigation agent never receives the decision contract from forge/.agents/rhei/templates/human-intervention-scanner/prompt-templates.yaml, including the required Decision: line and issue-body path. As a result, the evidence gate will not reliably produce the artifacts that issue-dispatch requires. Please either inline this as instructions in the state, or add actual Rhei support/tests for prompt_template if that is intended to be a new runtime feature.

kimeta:
> Correction to my previous review body, which was mangled by shell quoting.
> 
> I found one blocking issue in the Rhei template wiring.
> 
> forge/.agents/rhei/templates/human-intervention-scanner/states.yaml configures the investigate state with prompt_template, but the Rhei runtime schema I checked only consumes instructions; prompt_template is silently ignored. I instantiated the template locally and rendered a representative generated investigation task with local Rhei, and the task had an empty instructions field.
> 
> That means the investigation agent never receives the decision contract from forge/.agents/rhei/templates/human-intervention-scanner/prompt-templates.yaml, including the required Decision: line and issue-body path. As a result, the evidence gate will not reliably produce the artifacts that issue-dispatch requires. Please either inline this as instructions in the state, or add actual Rhei support/tests for prompt_template if that is intended to be a new runtime feature.

`prompt-templates.yaml` is a new feature in the upcoming release of `rhei`. This is tested locally and works fine.

jormundur00 (approved):

PR #8556, opened 2026-06-19
Fix Forge publication pending metrics resume

### LINKED ISSUE #8555: Forge publication deletes pending metrics before PR creation
Forge publication can lose `forge/.pending_metrics.json`, the transient PR-publication handoff, before publication has fully finished.

In issue #8508, generation succeeded and durable `stats/.../execution-metrics.json` was written. Publication then started, read `forge/.pending_metrics.json`, and later failed after the run had already reached the publication phase. Subsequent resumable runs re-entered publication directly and skipped the workflow driver, but the transient pending metrics sidecar was no longer a reliable input.

The subtle part is that durable `execution-metrics.json` is not a complete replacement for `.pending_metrics.json`: PR-only fields such as `post_generation_intervention`, `local_ci_verification`, and `library_update_alias_split` may exist only in pending metrics. Reconstructing from execution metrics alone would produce a lossy PR body and can hide the details reviewers need.

Expected behavior:

- Publication cleanup may remove resume helper artifacts from the PR branch, but it must not delete local inputs still needed by the active PR creation step.
- Resumable publication must preserve enough state to rebuild `.pending_metrics.json` from the durable execution-metrics entry plus local-only PR extras.
- `ContinuationMarker` should carry the exact durable metrics pointer and only the extra local PR fields needed for publication resume.
- Forge should not guess from the latest durable execution-metrics entry when marker-backed publication metrics are absent, because that loses local-only PR fields.
- Successful generation should publish directly instead of requiring a fail-then-resume cycle.

jormundur00:
### Summary

- keep `forge/.pending_metrics.json` available as a local publication input while removing it from PR branch contents
- add `ContinuationMarker.publicationMetrics`, which stores the exact durable execution-metrics pointer plus local-only PR extras
- reconstruct missing pending metrics during publication resume from `execution-metrics.json` plus marker extras, without falling back to lossy latest-metrics guessing
- snapshot pending-metrics extras into the marker before failed work is preserved, including post-generation intervention, local CI verification, and library-update alias split data
- document the pending-metrics/continuation lifecycle and add a fixture-backed publication-continuation E2E scenario

### Testing

- `./.venv/bin/python -m unittest tests.test_forge_metadata.FinalizeSuccessfulIssueTests tests.test_metrics_paths tests.test_pr_publication tests.test_schema_validator tests.test_make_pr_new_library_support tests.test_make_pr_improve_coverage tests.test_make_pr_not_for_native_image tests.test_make_pr_ni_run_fix tests.test_local_ci_verification`
- `GRAALVM_HOME="$JAVA_HOME" GRAALVM_HOME_25_0="$JAVA_HOME" GRAALVM_HOME_LATEST_EA="$JAVA_HOME" .venv/bin/python forge_metadata.py --fixture-testing --issue-number 9107 --reachability-metadata-path ..`
- continuation marker schema validation for fixture issue `9107`
- `git diff --check`

`grund check` still reports the pre-existing AGENTS.md v2 init-block warnings in root and `forge/AGENTS.md`.

Fixes #8555

### DIFF
--- a/forge/docs/continuation.md
+++ b/forge/docs/continuation.md
@@ -67,6 +67,14 @@ classes re-appear in the regenerated report.
   "newVersion": null,
   "libraryUpdateRoute": null,
   "libraryPreparationPreflight": null,
+  "publicationMetrics": {
+    "library": "com.acme:widget:1.4.0",
+    "timestamp": "2026-06-18T14:44:56.782450Z",
+    "extras": {
+      "post_generation_intervention": { "stage": "future-defaults-all" },
+      "local_ci_verification": { "status": "passed" }
+    }
+  },
   "phases": {
     "setup":        { "status": "completed", "preflightDone": true, "setupDone": true },
     "fix":          { "status": "skipped",   "iteration": null },
@@ -105,6 +113,13 @@ because the marker never enters a successful run's publication staging
 - `libraryPreparationPreflight` records dispatcher preflight output so resume
   can skip the preflight agent while still passing the original advisory setup
   context back to the workflow driver.
+- `publicationMetrics` records the committed per-library execution-metrics
+  entry (`library` plus `timestamp`) and only the local-only PR fields needed to
+  reconstruct `.pending_metrics.json` during publication resume. Durable metrics
+  remain the source for normal cost, token, coverage, and status evidence
+  (§FS-forge-run-metrics); the marker carries local extras such as
+  `post_generation_intervention`, `local_ci_verification`, and
+  `library_update_alias_split` when they exist.
 - Improve-coverage runs write the original `.baseline-stats.json` into the
   resolved test directory during setup. Resume treats that as preserved worktree
   state and reuses it instead of recomputing the baseline after generation may
@@ -140,8 +155,15 @@ does the PR making. Opening the pull request is the workflow's
 completion — a marker only exists for a run that failed before that point, so
 continuation never reaches a state with the pull request already open.
 In publication resume mode, Forge creates a clearance commit before it publishes
-the PR branch. The clearance commit deletes resume helper artifacts: the
-continuation marker, pending metrics, and human-intervention logs.
+the PR branch. The clearance commit deletes resume helper artifacts from the
+branch: the continuation marker and human-intervention logs. Pending metrics are
+a local PR-publication input, so they must remain readable until PR creation
+finishes even when the cleanup removes them from the branch index.
+If that transient pending file is missing in a later publication resume, Forge
+reconstructs it from the durable execution-metrics entry referenced by
+`publicationMetrics` and overlays the marker's `extras`. A publication marker
+without `publicationMetrics` is incomplete: Forge does not guess from the latest
+durable execution-metrics entry because that would lose local-only PR fields.
 
 ## 4. Relationship to human intervention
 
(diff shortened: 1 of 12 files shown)


kimeta:
I think this fix points at a broader contract issue around pending metrics.

If publication can reconstruct `forge/.pending_metrics.json` from `stats/.../execution-metrics.json`, then the PR publication path should probably not depend on `.pending_metrics.json` as a separate handoff file in the first place. It would be cleaner for publication to resolve the selected run metrics directly from the durable execution-metrics store, with one shared resolver for the “current issue/run” entry.

The reconstruction fallback also risks losing local-only data. In particular, `post_generation_intervention` is currently present in pending metrics for the PR body, but it is stripped from the committed execution-metrics entry. If a publication failure happens after pending metrics are deleted, reconstructing from execution metrics can recover basic metrics/status but not the post-generation intervention details. We should either preserve that intervention record somewhere durable enough for publication resume, or avoid deleting the original local handoff until publication is fully done.

So I’m fine with a small recovery fix, but I do not think reconstructing `.pending_metrics.json` from `execution-metrics.json` should become the long-term model. It would be better to remove/narrow `.pending_metrics.json` and make publication consume durable metrics directly, while explicitly preserving post-generation intervention data when it exists.

jvukicev:
Thanks, this should address that contract issue now.

What changed:

- Publication no longer treats durable `execution-metrics.json` as a complete replacement for `.pending_metrics.json`.
- `ContinuationMarker` now has `publicationMetrics`, which stores the exact durable metrics entry (`library` + `timestamp`) and only the local-only PR extras needed to rebuild pending metrics.
- The preserved extras currently include `post_generation_intervention`, `local_ci_verification`, and `library_update_alias_split` when present.
- On publication resume, Forge reconstructs `.pending_metrics.json` from the referenced durable execution-metrics entry plus those marker extras.
- I removed the execution-metrics-only fallback, so Forge does not guess from latest durable metrics and produce a lossy PR body when marker-backed publication metrics are absent.
- Failure preservation snapshots the latest pending metrics into the marker before committing the preserved branch, so fields appended after initial metrics writing, such as local CI verification, are kept.

I also added fixture issue `9107` to exercise publication-only continuation without running a generation workflow. The fixture E2E verifies that the reconstructed PR body includes both post-generation intervention and local CI verification sections.

kimeta:
Small fixture detail: `forge/fixture_github_issues/library-new-request-publication-continuation.yaml` uses `status: "success-with-intervention"` in the durable execution-metrics entry, but Forge's run status value is `success_with_intervention` with underscores.

The fixture may still exercise the preview path because it does not run the live new-library quality gate, but using the real status value would make this fixture more representative of the production metrics shape.

kimeta:
One small cleanup from the earlier version: it looks like `_publication_run_metrics_coordinate` and `load_latest_execution_metrics` are now unused after switching to marker-backed `publicationMetrics`.

I would remove them, especially `load_latest_execution_metrics`, since the updated design intentionally avoids latest-metrics guessing and requires the continuation marker to point at the exact durable metrics entry.

jvukicev:
Addressed both, thanks.

- Updated the fixture execution-metrics status from `success-with-intervention` to the production value `success_with_intervention`.
- Removed the dead `_publication_run_metrics_coordinate` helper.
- Removed `load_latest_execution_metrics` entirely, since publication resume now requires the marker to point at the exact durable metrics entry and should not guess from latest metrics.

Re-ran the focused unit batch, `git diff --check`, continuation-marker schema validation for fixture `9107`, and the fixture-backed publication-continuation E2E for issue `9107`.

kimeta (approved):

PR #8582, opened 2026-06-23
Add layered native-image test

Zeavee:
Adds a manual Layered tests workflow for running TCK native-image tests with a cached shared Native Image base layer.

The PR introduces Gradle tasks to build/validate the base layer and run layered test variants, plus a GitHub composite action that caches the layer by exact GraalVM/native-image version, OS/arch, native-image mode, ci.json, and TCK build logic inputs. Existing CI workflows are not changed to run layered tests automatically. The new lane is manual-only for now.

For now, we still fail some libraries, but many are already passing.


Zeavee:
Some examples of workflow run can be found here https://github.com/Zeavee/graalvm-reachability-metadata/actions/workflows/layered-tests.yml

vjovanov (approved):

PR #8714, opened 2026-06-27
Add reachability metadata for io.confluent:kafka-schema-serializer:8.3.0

### LINKED ISSUE #8564: Support for io.confluent:kafka-schema-serializer:8.3.0
### Reason for requesting the library

8.3.0 is the latest version. Unfortunately Confluent posts on their own maven repository first

https://packages.confluent.io/maven/io/confluent/kafka-schema-serializer/8.3.0/

mvanhorn:
## What does this PR do?

This adds GraalVM reachability metadata for `io.confluent:kafka-schema-serializer:8.3.0`, resolving the new-library request in #8564. Today, native-image builds that integrate Kafka clients with Confluent Schema Registry fail unless the user hand-maintains reflection config, because the serializer infrastructure loads its configuration, its subject- and context-name strategies, and its schema-id serdes reflectively at runtime. Shipping the metadata here means those applications build natively out of the box.

The reporter (jessecoddington) already runs this exact reflection configuration in production and posted it on the issue, so it serves as the verified basis for the entries rather than a guess. The metadata stays deliberately narrow: it registers only the classes that are reflectively reached and exercised by the test, and every entry is gated on `condition.typeReached` so it does not bloat images. Concretely, it covers the abstract serde and config base classes, the subject-name strategies (topic, topic-record, record, associated, and the reference strategies), the null context-name strategy, and the schema-id serializers and deserializers.

The contribution mirrors the existing `metadata/org.apache.kafka/kafka-clients` layout: a per-artifact index, the versioned reachability metadata, a registration entry in the framework list (re-sorted by artifact per the contributing guide), a Gradle test harness under `tests/src`, and the generated `stats/io.confluent/kafka-schema-serializer/8.3.0/stats.json`. Following the maintainer (kimeta) human-intervention note on the issue, the harness talks to an in-memory `mock://` Schema Registry instead of a live Kafka or Schema Registry service, so it needs no external infrastructure.

On testing: every new and changed JSON file parses and validates against the repository's own schemas (the library index, reachability metadata, library stats, and framework-list schemas). The harness does an Avro serialize/deserialize round trip through the mock registry, which reaches the abstract serde, the config, and the schema-id serdes, and it serializes once per configured subject-name strategy. Every strategy is loaded the way Confluent loads it, by naming the class in configuration and letting the serde config instantiate it: the subject-name strategies through `KafkaAvroSerializerConfig.key/valueSubjectNameStrategy()`, the context-name strategy through `contextNameStrategy()`, and the reference-subject-name strategies through `KafkaProtobufSerializerConfig.referenceSubjectNameStrategyInstance()`, which is the public config that defines `reference.subject.name.strategy`. Both config classes extend `AbstractKafkaSchemaSerDeConfig`, so reaching them satisfies the `typeReached` condition on every strategy entry. Verified against GraalVM for JDK 25: `nativeTest` is green at 11/11, and deleting a single strategy entry from the metadata fails exactly the case that loads it.

Contribution requires the Oracle CLA; the signing gate is handled downstream.

The test resolves the library from the Confluent Maven repository (scoped to the `io.confluent` group) and otherwise talks only to an in-memory `mock://` Schema Registry, so there is no live network, Kafka broker, or Docker dependency.

- Dependency resolution: [`build.gradle`](https://github.com/oracle/graalvm-reachability-metadata/blob/master/tests/src/io.confluent/kafka-schema-serializer/8.3.0/build.gradle)
- Mock-registry serialization paths: [`KafkaSchemaSerializerTest.java`](https://github.com/oracle/graalvm-reachability-metadata/blob/master/tests/src/io.confluent/kafka-schema-serializer/8.3.0/src/test/java/io_confluent/kafka_schema_serializer/KafkaSchemaSerializerTest.java)

Fixes #8564


kimeta:
Hey @mvanhorn, thanks for the contribution! CI fails in `nativeTest`: all seven `strategyTypesAreReflectivelyInstantiable` cases fail. The metadata itself is fine, the problem is that every entry is gated on `typeReached: AbstractKafkaSchemaSerDeConfig`, and that test never reaches that class, so the registrations stay inactive. Do you want to fix this, or should I do it?

mvanhorn:
Thanks for the precise diagnosis, that was exactly it. Removed the `typeReached: AbstractKafkaSchemaSerDeConfig` condition from all seven strategy registrations so they are active for direct reflective instantiation. Everything else in the metadata is untouched.

One file, 21 deletions, JSON validated. I could not run the Gradle validation or `nativeTest` locally since the Gradle 9.1 wrapper is not cached here and my JDK lacks `native-image`, so CI will be the real confirmation that the seven `strategyTypesAreReflectivelyInstantiable` cases now pass.

kimeta:
Thanks for the quick turnaround, I should have been clearer in my first comment.

Every entry in this repository must carry a `condition.typeReached`, by functional-spec: registration has to be gated on actual reachability so consumers do not pay image size for metadata they never exercise. Dropping the condition from the seven strategy entries makes them unconditional for every consumer.

The condition was correct, the test is what needs to change. `strategyTypesAreReflectivelyInstantiable` reflects on the strategies directly from test code, so it never reaches `AbstractKafkaSchemaSerDeConfig` and the condition never fires. Please restore the conditions and exercise the strategies through the library's own API instead, the way `serializesWithEachSubjectNameStrategy` already does; configuration drives the reflective load, which is both how Confluent uses these classes and what makes the condition fire.

Separately, `Validate library stats JSON` fails because the PR has no `stats/io.confluent/kafka-schema-serializer/8.3.0/stats.json`, which is required for every metadata directory. Run `./gradlew generateLibraryStats -Pcoordinates=io.confluent:kafka-schema-serializer:8.3.0` and commit the result, but do it last: it runs the test lane and measures coverage against the final metadata and test.

mvanhorn:
@kimeta restored `condition.typeReached` on all seven strategy entries.

`strategyTypesAreReflectivelyInstantiable` is gone. The tests now name each strategy in configuration and let `AbstractConfig.getConfiguredInstance` load it from a subclass of `AbstractKafkaSchemaSerDeConfig`, so the superclass condition actually fires. Reference-subject strategies go through `KafkaProtobufSerializerConfig` because that's where `reference.subject.name.strategy` lives in 8.3.0.

`stats.json` is in, generated last.

kimeta (approved):

PR #8860, opened 2026-06-29
[GenAI] Add support for com.rabbitmq.client:amqp-client:0.8.0 using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #7103

This PR introduces tests and metadata for com.rabbitmq.client:amqp-client:0.8.0, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 563570
- Cached input tokens: 5624832
- Output tokens: 45469
- Metadata entries: 85
- Test-only metadata entries: 1
- Iterations: 7
- Library coverage percentage: 17.21
- Generated lines of code: 308
- Tested library lines of code: 4882

### Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `1c55bcf510a4ce9e84492d4d2350315d9eef14f2`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 21/21 covered calls (100.00%)
- Reflection: 19/19 covered calls (100.00%)
- Resources: 2/2 covered calls (100.00%)

Library coverage:
- Instruction: 19513/120330 (16.22%)
- Line: 4882/28372 (17.21%)
- Method: 1695/8231 (20.59%)

### Local CI Verification

- Status: `success`
- Commands run: 2
- Fixup attempts: 0

PR #8789, opened 2026-06-28
[GenAI] Add support for dev.profunktor:redis4cats-effects_3:1.7.2 using gpt-5.5

vjovanov:
## What does this PR do?

Fixes: #4885

This PR introduces tests and metadata for dev.profunktor:redis4cats-effects_3:1.7.2, enabling support for this library.

Summary:
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 352098
- Cached input tokens: 3338240
- Output tokens: 29474
- Metadata entries: 0
- Iterations: 6
- Library coverage percentage: 29.44
- Generated lines of code: 393
- Tested library lines of code: 219

### Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `495672c318b0fc9e4429f334c582286782220a41`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Library coverage:
- Instruction: 2230/15520 (14.37%)
- Line: 219/744 (29.44%)
- Method: 261/1898 (13.75%)

### Local CI Verification

- Status: `success`
- Commands run: 2
- Fixup attempts: 0

PR #8891, opened 2026-07-01
Persist Docker declarations through generation and repair missing allow-list pins at finalization

### LINKED ISSUE #8890: Docker declaration files dropped between preflight and PR cause human-intervention
## Problem

Generated new-library PRs whose tests use Docker are being labeled `human-intervention` because the Docker declaration files are incomplete or missing, even though preflight already computed the right images.

Two failure shapes, same root cause area:

- **Allow-list pin is dropped.** Preflight's deterministic `docker_image` setup writes `tests/tck-build-logic/src/main/resources/allowed-docker-images/Dockerfile-<slug>` into the worktree (`"result": "applied"`), but the generation commit (`_commit_test_sources`) only stages the library's own `tests/src/<g>/<a>/<v>` subtree, and the iteration loop runs `git reset --hard <checkpoint>`. The shared-tree pin is never committed and is wiped by the first reset — so it never reaches the PR. Example: #8860 (`rabbitmq:4.1`).
- **`required-docker-images.txt` is never written by preflight.** For a new library the test dir does not exist until `scaffold` runs, so preflight cannot write `required-docker-images.txt` and demotes it to advisory guidance the agent often skips. With the file missing, `_coordinate_uses_docker` returns false, `pullAllowedDockerImages` is skipped entirely, and the test then fails under the network-disabled gate. Example: #8789 (`valkey/valkey:7.2.6`), where the agent used the image in source and even recorded the `Dockerfile-valkey` target in its plan, yet neither declaration file landed.

## Expected

Images that a test needs should end up declared in both places deterministically:
- `required-docker-images.txt` for the library, and
- an allow-list `Dockerfile-<slug>` — so `pullAllowedDockerImages` pre-pulls them and the network-disabled test gate passes.

This is fully specified data (preflight already knows the images; the test source states them), so it should not require human follow-up.

Relevant specs: declared Docker images `§FS-repository-functional-spec.5.2`; deterministic preflight setup `§forge/ORCH-forge-orchestration-spec.1.1`; finalization `§forge/FS-local-ci-equivalent-verification`.

kimeta:
Closes #8890

## What this does

New-library PRs with Docker-backed tests kept landing in `human-intervention` because their Docker declaration files went missing even though preflight had already computed the right images. Two mechanisms were at play: the allow-list pin preflight writes to the shared `allowed-docker-images/` tree was dropped by the library-scoped scaffold commit and the iteration `git reset --hard` (e.g. #8860's `rabbitmq:4.1`), and `required-docker-images.txt` was never written by preflight at all because a new library's test dir does not exist until `scaffold` runs, so it silently became advisory work the agent skipped (e.g. #8789's `valkey/valkey:7.2.6`). The data is fully determined — preflight knows the images and the test source states them — so this is fixed deterministically rather than left for a human.

## The three changes

- **Preflight also writes `required-docker-images.txt`.** `_apply_docker_image_setup` now declares the image in `tests/src/<g>/<a>/<v>/required-docker-images.txt` alongside the allow-list `Dockerfile-<slug>`, so a `docker_image` setup produces both declarations (`§forge/ORCH-forge-orchestration-spec.1.1`). Before the test dir exists it stays a no-op, matching the advisory-fallback contract.
- **Pins are committed after branch creation.** The new-library driver re-applies the deterministic setup once the scaffolded test dir exists, then stages the shared `Dockerfile-<slug>` targets into the `Scaffold <library>` commit. Because that commit is the reset checkpoint, `git reset --hard` during iteration preserves them instead of wiping them.
- **Finalization repairs a still-missing pin.** The `pullAllowedDockerImages` gate is preserved; on failure it now pins any image declared in `required-docker-images.txt` but absent from the allow-list (slug derived, `FROM <image>`), commits it, and retries the pull once before falling through (`§forge/FS-local-ci-equivalent-verification`). The image string comes from the declaration file, so no source parsing is involved.

## What each case needs

| Case | Fixed by |
|------|----------|
| #8860 — declared, allow-list pin dropped | commit-after-branch; finalization repair as backstop |
| #8789 — preflight knew the image, no `required-docker-images.txt` | preflight write + committed pin |
| Agent-declared image, pin still missing at finalization | deterministic repair on the pull gate |

Both declarations land in both places, so `pullAllowedDockerImages` pre-pulls the image and the network-disabled test gate (`§FS-repository-functional-spec.5.2`) runs clean. The only residual case — a test that uses an image and declares nothing — is left to the agent instruction and the runtime-observation backstop, unchanged here.

### DIFF
--- a/forge/ai_workflows/drivers/add_new_library_support.py
+++ b/forge/ai_workflows/drivers/add_new_library_support.py
@@ -545,8 +545,12 @@ def main(argv=None):
         # Re-apply deterministic docker setup now that the scaffolded test dir exists: this
         # writes `required-docker-images.txt` and re-pins the allow-list Dockerfiles. Stage the
         # shared pins explicitly so the library-scoped scaffold commit (the reset checkpoint)
-        # preserves them instead of dropping them on the next `git reset --hard`.
-        apply_library_preparation_setup(library_preparation_preflight, reachability_repo_path)
+        # preserves them instead of dropping them on the next `git reset --hard`. Restricted to
+        # docker so dependency edits stay advisory, matching the model context already rendered
+        # pre-scaffold (which lists them as pending work).
+        apply_library_preparation_setup(
+            library_preparation_preflight, reachability_repo_path, only_kinds={"docker_image"}
+        )
         docker_setup_targets = [
             os.path.join(reachability_repo_path, item["target"])
             for item in (library_preparation_preflight or {}).get("applied_setup") or []
--- a/forge/ai_workflows/drivers/add_new_library_support.py
+++ b/forge/ai_workflows/drivers/add_new_library_support.py
@@ -52,6 +52,7 @@ from utility_scripts.continuation_marker import (
 from utility_scripts.dynamic_access_exhaust_report import resolve_workflow_exhaust_report
 from utility_scripts.gradle_environment import gradle_command_environment
 from utility_scripts.library_preparation_preflight import (
+    apply_library_preparation_setup,
     prepare_library_preparation_preflight,
 )
 from utility_scripts.metadata_index import is_not_for_native_image, write_not_for_native_image_marker
@@ -541,8 +542,18 @@ def main(argv=None):
     directory_path = module_dir
     index_json_path = os.path.join(reachability_repo_path, "metadata", package, artifact, "index.json")
     if not resume_existing_tree:
+        # Re-apply deterministic docker setup now that the scaffolded test dir exists: this
+        # writes `required-docker-images.txt` and re-pins the allow-list Dockerfiles. Stage the
+        # shared pins explicitly so the library-scoped scaffold commit (the reset checkpoint)
+        # preserves them instead of dropping them on the next `git reset --hard`.
+        apply_library_preparation_setup(library_preparation_preflight, reachability_repo_path)
+        docker_setup_targets = [
+            os.path.join(reachability_repo_path, item["target"])
+            for item in (library_preparation_preflight or {}).get("applied_setup") or []
+            if isinstance(item, dict) and item.get("kind") == "docker_image" and item.get("target")
+        ]
         # Add generated files to git and commit; record commit hash (do not use it)
-        subprocess.run(["git", "add", directory_path, index_json_path], check=False)
+        subprocess.run(["git", "add", directory_path, index_json_path, *docker_setup_targets], check=False)
         subprocess.run(["git", "commit", "-m", f"Scaffold {library}"], check=False, capture_output=True, text=True)
     checkpoint_commit_hash = subprocess.check_output(["git", "rev-parse", "HEAD"], text=True).strip()
     log_stage("scaffold", f"Checkpoint at {checkpoint_commit_hash}")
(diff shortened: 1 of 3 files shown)


jormundur00 (changes requested):
I found two issues that should be addressed before merge.

1. The new retry path in `forge/utility_scripts/local_ci_verification.py` does not actually repair the common "same image name, newer tag" case.

`_repair_missing_allowed_images()` skips repair whenever `Dockerfile-<slug>` already exists, even if it points at the wrong image tag. The relevant branch is:

- `_repair_missing_allowed_images()` lines 966-968 in this PR
- retry call site at lines 778-787

I reproduced this locally with:
- `required-docker-images.txt` containing `postgres:17`
- existing `tests/tck-build-logic/src/main/resources/allowed-docker-images/Dockerfile-postgres` containing `FROM postgres:16`

The helper returned `[]` and left `Dockerfile-postgres` unchanged, so the retried `pullAllowedDockerImages` would still fail. In other words, this does not repair the tag-bump case; it only creates a new Dockerfile when no slugged file exists yet.

2. The agent still receives stale preflight context after the new post-scaffold reapply.

`prepare_library_preparation_preflight()` builds `library_preparation_preflight_context` before scaffold and that snapshot is passed into `StrategyClass`. After that, this PR re-applies deterministic setup again once the scaffolded test dir exists. That means setup that only becomes applicable after scaffold, especially dependency edits in `build.gradle`, is still described to the model as pending advisory work even though Forge has already applied it. That risks duplicate or conflicting edits in generated files.

I did not run the full Forge Python test suite in this environment because the local worktree here does not have the required Python test dependencies installed, but I did reproduce the first issue directly in an isolated local repo.

kimeta:
@jormundur00 — thanks, both good catches. Handled them differently:

**2. Stale preflight context after the reapply — fixed.**
You're right. The reapply called the general setup function, so it re-applied `dependency` edits too, while the model's context was frozen pre-scaffold and still listed them as pending. I've scoped the post-scaffold reapply to `only_kinds={"docker_image"}` (new param on `apply_library_preparation_setup`, which now preserves prior `applied_setup` results for kinds it doesn't touch). Docker pins are re-applied as before; dependencies stay advisory and consistent with the already-rendered context. No context rebuild needed.

**1. Newer-tag case not repaired — intentional, and fixed at the root instead.**
Your read of the code is correct: the repair skips when `Dockerfile-<slug>` exists with a different tag. But that skip is deliberate and shouldn't change. The allow-list holds exactly one `FROM` per slug (`DockerUtils.imageNameFromFile` throws otherwise), and each slug file is shared by every library using that image — e.g. all three postgres consumers pin the same `postgres:18-alpine`. If the repair rewrote `Dockerfile-postgres` from `:16` to `:17`, every other library pinned to `:16` would then fail validation; clobbering just relocates the break. A real tag conflict isn't auto-resolvable — it's a decision about the repo's single shared tag, so it correctly falls through to human-intervention.

So rather than repair a conflict at finalization, I've prevented it at generation: the preflight prompt now instructs the agent to reuse the exact tag already pinned under `allowed-docker-images/` when one exists, instead of introducing a new one. If a conflicting tag still slips through, the pull gate fails and it goes to human-intervention — which is the right outcome.

Pushed in `4308b86a9e`.

jormundur00 (approved):
Reviewed the latest head commit 4308b86a9e9e67fb10b6fac1721b97009db992e8 locally. The two prior issues look addressed: the post-scaffold reapply is now scoped to docker-only setup, so it no longer mutates dependency state behind the already-rendered preflight context, and preflight now instructs the agent to reuse an existing allow-listed Docker tag instead of introducing a conflicting shared slug tag. I did not find any remaining blocking issues in the updated implementation.

PR #8900, opened 2026-07-01
Migrate model provider from oca to openai-codex

### LINKED ISSUE #8899: Migrate model provider from oca to openai-codex
We migrated from the `oca` provider to `openai-codex`, so models have different names (the `oca/` prefix is dropped).

kimeta:
We migrated from the `oca` provider to `openai-codex`, so models have different names (the `oca/` prefix is dropped).
This also adds logging of review-pr cost.

Fixes #8899


jormundur00 (changes requested):
Two issues block this for me:

1. `DEFAULT_REVIEW_MODEL` now defaults to `gpt-5.6-terra` in `forge_metadata.py`, but the local Codex CLI in this environment does not recognize that model cleanly. Running `codex exec --json -m gpt-5.6-terra ...` here emits: `Model metadata for `gpt-5.6-terra` not found. Defaulting to fallback metadata; this can degrade performance and cause issues.` So the new default already lands on a fallback path in the runtime Forge is expected to use.

2. The review-model migration is internally inconsistent across entrypoints. `forge_metadata.py` now defaults PR review / failure analysis to `gpt-5.6-terra`, but `forge/do_up_to_date_work.sh` still defaults `FORGE_REVIEW_MODEL` to `gpt-5.4`. That means direct `forge_metadata.py --review-pr ...` runs and the standard do-work loop will review the same PRs with different models unless the env var is explicitly set.

kimeta:
> Two issues block this for me:
> 
> 1. `DEFAULT_REVIEW_MODEL` now defaults to `gpt-5.6-terra` in `forge_metadata.py`, but the local Codex CLI in this environment does not recognize that model cleanly. Running `codex exec --json -m gpt-5.6-terra ...` here emits: `Model metadata for `gpt-5.6-terra` not found. Defaulting to fallback metadata; this can degrade performance and cause issues.` So the new default already lands on a fallback path in the runtime Forge is expected to use.
> 2. The review-model migration is internally inconsistent across entrypoints. `forge_metadata.py` now defaults PR review / failure analysis to `gpt-5.6-terra`, but `forge/do_up_to_date_work.sh` still defaults `FORGE_REVIEW_MODEL` to `gpt-5.4`. That means direct `forge_metadata.py --review-pr ...` runs and the standard do-work loop will review the same PRs with different models unless the env var is explicitly set.

You need to update codex-cli so it actually has this models available.

jormundur00 (approved):

PR #8931, opened 2026-07-02
Fix stale JDK guidance and make review prompt more exact

### LINKED ISSUE #8930: Bad JDK version instructions and reviewer doesn't always follow the script's rules
Two related problems in the generation/review automation:

### 1. Stale JDK version in contributor guidance
`AGENTS.md` (and the `CLAUDE.md` symlink) still states `JAVA_HOME set to JDK 21`, but the default test JVM moved to **25** (#8396). Because that stale `21` is the only JDK constant in the agent's context, generators regenerate test `build.gradle` files with a hardcoded `JavaLanguageVersion.of(21)` / `jvmToolchain(21)` instead of the `tck.testJvmVersion.get()` getter. `explicitTestJvmVersion(...)` then scrapes that literal and pins those tests to 21, so the metadata is validated on the wrong JVM even though CI is green.

### 2. Reviewer blocks on rules that aren't in the script
The review agent sometimes requests changes based on self-formed test-quality, test-scope, or "end-user behavior" judgments that are not backed by any enumerated rule in the applicable review skill. This produces false `human-intervention` labels — including blocking tests that already exercise the library's own (relocated/shaded) types.

kimeta:
- Point contributor guidance at JDK 25 and convert three drifted Scala/Kotlin test fixtures to `tck.testJvmVersion.get()` so they track the resolved matrix JVM instead of a hardcoded 21.
- Make `build_review_prompt` only request changes for concrete violations of enumerated review rules, and add the `pi_gpt-5.6` sol/luna/terra strategies.

Fixes #8930


jormundur00 (approved):
LGTM

PR #8950, opened 2026-07-03
Stop writing singular test file metrics

### LINKED ISSUE #8949: Stop writing singular test file metrics
### Problem

Forge writes a singular `artifacts.test_file` path into each execution-metrics entry. A run can create or update multiple test files, so this field is incomplete and can point to an arbitrary helper source rather than a meaningful test.

The version-controlled test tree and pull-request diff already provide the authoritative set of test files. The singular metric is therefore redundant and misleading.

### Proposed change

- Stop writing `artifacts.test_file` in new success and failure metrics.
- Remove the obsolete test-file discovery and workflow plumbing.
- Keep the schema property optional so historical execution metrics remain valid.
- Continue writing `artifacts.metadata_file`.

kimeta:
### Summary

- stop writing the singular `artifacts.test_file` field in new Forge execution metrics
- remove the obsolete test-source scan and its workflow plumbing
- keep the deprecated schema property optional so historical records remain valid
- document the version-controlled test tree and PR diff as the authoritative test-file set

Fixes #8949

### DIFF
--- a/forge/DEVELOPING.md
+++ b/forge/DEVELOPING.md
@@ -294,7 +294,6 @@ that persistent layer was configured without printing the instruction text.
       library_version="1.2.3",
       agent=None,  # Agent implementation exposing token counters
       global_iterations=3,
-      tests_root="/abs/path/to/tests/root",
       strategy_name="basic_iterative",
       status="success",
   )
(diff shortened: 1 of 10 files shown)


jormundur00 (approved):
Approved.

Verified locally:
- New records omit the singular `artifacts.test_file` field; historical records containing it remain schema-valid.
- All 2,064 checked-in execution-metrics files validate against the updated schema.
- Focused Forge tests passed (14 tests), as did Python compilation and diff whitespace checks.

PR #8975, opened 2026-07-05
Fix input-token cost billed as zero

kimeta:
## Bug

There is a bug in the token-cost accounting: the full-rate input charge is dropped, so run costs are under-reported.

The biller computes full-rate input tokens as `total_tokens_sent - cached_input_tokens`, assuming cached reads are folded into the input total. In fact `total_tokens_sent` (full-rate input) and `cached_input_tokens` (cache reads) are **separate, non-overlapping** counters. With any real cache reuse `cached_input_tokens > total_tokens_sent`, so the subtraction clamps billable input to `0` and the full-rate input cost silently becomes `$0`.

Concretely, a run reporting 212,117 input / 1,833,984 cached / 23,679 output on `gpt-5.6-terra`:

| Component | Tokens | Rate/1M | Reported | Correct |
|---|---|---|---|---|
| Full-rate input | 212,117 | $2.50 | **$0.0000** | $0.5303 |
| Cached input | 1,833,984 | $0.25 | $0.4585 | $0.4585 |
| Output | 23,679 | $15.00 | $0.3552 | $0.3552 |
| **Total** | | | **$0.8137** | **$1.3440** |

### DIFF
--- a/forge/tests/test_metrics_paths.py
+++ b/forge/tests/test_metrics_paths.py
@@ -17,6 +17,8 @@ from ai_workflows.drivers.java_fail_workflow import JAVAC_CONFIG, resolve_fix_me
 from utility_scripts.library_stats import load_library_stats_entry, resolve_stats_file_path
 from utility_scripts.metrics_writer import (
     append_execution_metrics,
+    calc_model_session_cost,
+    collect_token_usage_metrics,
     count_metadata_entries,
     count_test_only_metadata_entries,
     create_run_metrics_output_json,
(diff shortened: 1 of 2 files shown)


jormundur00 (changes requested):
The durable run-metrics path is fixed, but a second active cost path remains incorrect.

`calc_model_session_cost()` still computes full-rate input as `max(input_tokens - cached_input_tokens, 0)`. `forge_metadata.extract_codex_token_usage_summary()` calls that helper, so Codex review token summaries continue to under-report cost whenever cached reads exceed full-rate input.

Reproduced with the PR example: durable metrics report `$1.3440`, while the review summary still reports `$0.8137`. Please make the helper bill the two non-overlapping counters independently and add regression coverage for both paths.

jormundur00 (approved):
Approved on re-review.

The follow-up fixes `calc_model_session_cost()` to bill full-rate and cached input independently, resolving the remaining Codex-summary under-reporting path. Regression tests cover both the shared helper and durable metrics path.

Verified on a clean merge with current `origin/master`: the reported example now yields `$1.3440` through both paths; focused metrics/schema tests passed (16 tests).

PR #8979, opened 2026-07-06
Default Pi provider to openai-codex

### LINKED ISSUE #8978: Pi runs intermittently fail with missing azure-openai-responses key
Forge strategies that omit `provider` make Pi resolve the bare model name against its full catalog. Model names can exist in Pi's builtin catalog under other providers — `gpt-5.6-terra` also resolves to `azure-openai-responses` — so resolution is nondeterministic and runs intermittently fail with:

```
PiRpcError: Pi RPC `prompt` failed: No API key found for azure-openai-responses.
```

The same model works when a provider is pinned (interactive Pi uses the configured default provider). Forge should pin a default provider instead of relying on ambiguous name resolution.

kimeta:
## What this does

Fixes: #8978

Strategies without an explicit `provider` let Pi resolve the bare model name against its full catalog (§forge/AR-agent-api). Newly introduced model names also exist in Pi's builtin catalog under other providers, `gpt-5.6-terra` also matches `azure-openai-responses`, so resolution was nondeterministic and runs intermittently failed with `No API key found for azure-openai-responses`.

`PiAgent` now falls back to `openai-codex` when the strategy gives no provider, so `--provider` is always passed and resolution is deterministic. Strategies with an explicit `provider` (e.g. `openrouter`) are unaffected.

### DIFF
--- a/forge/ai_workflows/agents/pi_agent.py
+++ b/forge/ai_workflows/agents/pi_agent.py
@@ -14,6 +14,9 @@ from utility_scripts.gradle_test_runner import run_gradle_test_command
 from utility_scripts.pi_logs import build_pi_log_path
 
 
+DEFAULT_PI_PROVIDER = "openai-codex"
+
+
 @Agent.register("pi")
 class PiAgent(Agent):
     """Agent adapter that drives Pi through its RPC mode."""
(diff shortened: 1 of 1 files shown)


jormundur00 (approved):
Approved.

Verified locally:
- Default Pi provider produces `--provider openai-codex`.
- Explicit provider overrides remain unchanged.
- Syntax and whitespace checks pass.
- The Forge test suite has five pre-existing failures that reproduce on `origin/master`; this PR introduces none.

PR #8984, opened 2026-07-06
Ignore gosu Go-stdlib false positives in docker scan

### LINKED ISSUE #8985: gosu Go-stdlib false positives block new docker images in the vulnerability scan
## Problem

The docker-image vulnerability gate (`checkAllowedDockerImages` / `GrypeTask`, §CI-scan-docker-images) cannot accept a **new** allowed image whose only High/Critical findings come from `gosu`.

`gosu` (`/usr/sbin/gosu`) is a ~200-line setuid helper that drops privileges and execs the container entrypoint at startup. It is statically linked against an old Go stdlib, so grype attributes every Go stdlib CVE for that version to the binary — even though gosu exercises none of the vulnerable code paths (`net/http`, `crypto/tls`, `crypto/x509`, `net/url`, `encoding/gob`, archive parsing). These are unreachable false positives, and gosu ships that stdlib upstream so there is nothing to patch.

Because gosu is bundled in essentially every official DB/broker image, this is not image-specific. Existing allowed images (`postgres`, `mariadb`, `nats`, …) escape it only because PRs scan **changed** images and those were grandfathered in when clean. A newly added gosu-bearing image cannot pass:

1. grype flags it as High/Critical (e.g. `rabbitmq:4.1` → 55 findings, all from `/usr/sbin/gosu`);
2. `GrypeTask` then compares it against a master baseline via `git show origin/master:<new-Dockerfile>`, which crashes with exit 128 because the Dockerfile does not exist on master yet.

Concrete blocker: the `com.rabbitmq.client:amqp-client:0.8.0` support PR (#8860) needs `rabbitmq:4.1`, whose only High/Critical findings are gosu stdlib noise, and the scan job fails as above.

## Expected

The gosu Go-stdlib false positives should be suppressed repo-wide so the scan gate reflects genuinely reachable vulnerabilities, while real CVEs in gosu's own package or in any other image package remain reported. Adding a gosu-bearing test image should not require the vulnerability scan to be manually overridden.

kimeta:
Closes #8985

## What this does

The docker-image vulnerability gate (`checkAllowedDockerImages` / `GrypeTask`, [§CI-scan-docker-images](docs/ci.md)) rejects any newly-added allowed image that grype reports as High/Critical. When investigating why the `amqp-client` support PR could not go green, every one of `rabbitmq:4.1`'s **55 High/Critical findings** turned out to come from a single binary, `/usr/sbin/gosu`,  statically linked against an outdated Go `1.22.2` stdlib. gosu is a ~200-line setuid helper that only drops privileges and execs the container entrypoint at startup; it never touches the vulnerable stdlib code paths (`net/http`, `crypto/tls`, `crypto/x509`, `net/url`, `encoding/gob`, archive parsing), so these are unreachable false positives. gosu ships that stdlib upstream, so there is nothing to patch on our side. This adds a repo-root `.grype.yaml` that grype auto-loads to ignore exactly those findings.

## Why it blocks the gate today

gosu is bundled in essentially every official DB/broker image, so this is not rabbitmq-specific — `postgres`, `mariadb`, and others carry the same noise. Existing allowed images escape it only because PRs scan **changed** images and those were grandfathered in when clean. A *new* gosu-bearing image cannot pass: grype flags it, `GrypeTask` then compares against a master baseline via `git show origin/master:<new-Dockerfile>`, and that crashes with exit 128 because the file does not exist on master yet.

## The rule

```yaml
ignore:
  - package:
      name: stdlib
      location: "/usr/sbin/gosu"
```

Scoped to the Go stdlib artifact **on the gosu binary only**. A real CVE in gosu's own package, or in any other package in any image, is still reported — verified locally: `rabbitmq:4.1` drops to `0` High/Critical while `postgres:18-alpine` still reports its unrelated `42`.

### DIFF
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -1,0 +1,19 @@
+# grype configuration for the docker-image vulnerability gate (§CI-scan-docker-images).
+#
+# grype auto-loads this file from the repository root, so it applies to the
+# `checkAllowedDockerImages` / `GrypeTask` scan both on PRs (changed images) and
+# on the weekly all-images run.
+#
+# gosu (/usr/sbin/gosu) is a ~200-line setuid helper that only drops privileges
+# and execs the container entrypoint at startup. It is statically linked against
+# an old Go stdlib, so grype attributes every Go stdlib CVE for that version to
+# the binary even though gosu exercises none of the vulnerable code paths (no
+# net/http, crypto/tls, crypto/x509, net/url, encoding/gob, or archive parsing).
+# These are unreachable false positives in test-only fixture images, and gosu
+# ships that stdlib upstream so there is nothing to patch. Scoped to the Go
+# stdlib artifact on the gosu binary only: a real CVE in gosu's own package or
+# in any other image package is still reported.
+ignore:
+  - package:
+      name: stdlib
+      location: "/usr/sbin/gosu"
--- a/.grype.yaml
+++ b/.grype.yaml
@@ -4,16 +4,183 @@
 # `checkAllowedDockerImages` / `GrypeTask` scan both on PRs (changed images) and
 # on the weekly all-images run.
 #
-# gosu (/usr/sbin/gosu) is a ~200-line setuid helper that only drops privileges
-# and execs the container entrypoint at startup. It is statically linked against
-# an old Go stdlib, so grype attributes every Go stdlib CVE for that version to
-# the binary even though gosu exercises none of the vulnerable code paths (no
-# net/http, crypto/tls, crypto/x509, net/url, encoding/gob, or archive parsing).
-# These are unreachable false positives in test-only fixture images, and gosu
-# ships that stdlib upstream so there is nothing to patch. Scoped to the Go
-# stdlib artifact on the gosu binary only: a real CVE in gosu's own package or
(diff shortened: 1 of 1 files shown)


jormundur00 (changes requested):
The rule in  ignores every Grype finding for  at , with no CVE, version, or expiry constraint. That can hide a future Go standard-library vulnerability reachable by gosu itself, weakening the weekly all-images security gate. Please scope the ignore to the specific confirmed false-positive CVE(s), or provide a VEX-style justification per finding.

jvukicev:
Correction to my review: the rule is in .grype.yaml line 16 and matches package stdlib at /usr/sbin/gosu. It ignores every Grype finding for that package/path, with no CVE, version, or expiry constraint. Please scope the ignore to the specific confirmed false-positive CVE(s), or provide a VEX-style justification per finding.

jormundur00 (approved):
Re-reviewed the updated commit locally. The prior concern is resolved: every ignore is now pinned to a specific vulnerability ID, stdlib version go1.22.2, and /usr/sbin/gosu, so it cannot suppress a newly disclosed CVE or the same finding in another binary. The Grype configuration loads successfully.

PR #8869, opened 2026-06-29
[GenAI] Add support for org.hsqldb:hsqldb:2.7.3 using gpt-5.5 (chunked dynamic-access)

kimeta:
## What does this PR do?

Refs: #701

This PR introduces tests and metadata for org.hsqldb:hsqldb:2.7.3, enabling support for this library.

Summary:
- Chunked dynamic-access: yes
- Chunk class threshold: 15
- Current chunk class count: 15
- Processed dynamic-access classes: completed=0, skipped=0, exhausted=15, failed=0
- Strategy: dynamic_access_main_sources_pi_gpt-5.5
- Agent: pi
- Model: gpt-5.5
- Input tokens: 2184720
- Cached input tokens: 27712512
- Output tokens: 252260
- Metadata entries: 190
- Test-only metadata entries: 31
- Iterations: 49
- Library coverage percentage: 25.25
- Generated lines of code: 1541
- Tested library lines of code: 0

### Forge

- Forge monitored branch: `master`
- Forge branch: `master`
- Forge commit hash: `5e46b70a65d88c3d6a7e147dfe7f55d91b4bbf7c`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 0/82 covered calls (0.00%)
- Reflection: 0/69 covered calls (0.00%)
- Resources: 0/13 covered calls (0.00%)

Library coverage:
- Instruction: 91103/360737 (25.25%)
- Line: N/A
- Method: 2992/10234 (29.24%)

### Local CI Verification

- Status: `success`
- Commands run: 2
- Fixup attempts: 0

PR #8987, opened 2026-07-07
Match line-less dynamic access via agent configuration origins

kimeta:
## What this does

Dynamic-access coverage marks a call site covered only when its frame carries a source line that JaCoCo saw executed. Jars compiled without `LineNumberTable` — e.g. `org.hsqldb:hsqldb:2.7.3`, where 0 of 685 classes have one — produce frames like `org.hsqldb.Foo.bar(Unknown Source)` and a line-less JaCoCo report, so coverage is hardwired to **0/N** no matter what the tests exercise. That starved the dynamic-access strategy of guidance and produced misleading PR stats (the closed #8869).

This adds a fallback for exactly that case (§TCK-test-harness.8, [docs/tck.md](docs/tck.md)): when the dynamic-access report has call sites but **none** carries a line number, the harness collects the `native-image-agent` **configuration-origin tree** from a JVM `javaTest` run and matches each statically reported call site against the observed origin paths. Line-based matching stays the primary path and is unchanged for jars with line info.

## How the fallback works

| step | where | what happens |
| --- | --- | --- |
| gate | `LibraryStatsSupport.lineMatchingImpossible` | true iff `totalCalls > 0` and every call site is line-less — data-driven, no per-library config |
| collect | `AbstractLibraryStatsTask.maybeCollectAgentOrigins` | runs the JVM-only `javaTest` lane with `-agentlib:native-image-agent=config-output-dir=...,experimental-configuration-with-origins` (gated by the new `dynamicAccessOriginsOutput` property; JaCoCo disabled for that run); any failure degrades gracefully to today's behaviour |
| candidates | `loadAgentOriginCandidates` | statically reported call sites, taken from the reports as-is (the reports only contain library call sites; the coverage side only consults the result for line-less library frames) |
| match | `parseAgentOrigins` / `matchAgentOriginPath` | streams the compressed origin tree keeping only the current branch in memory; on every configuration-bearing path, a call site is covered when its **exact tracked API** appears on the path with its **caller class#method** above it — nearest caller wins, and a nested tracked API stops the caller walk |
| consume | `parseDynamicAccessFile` | line-less site covered ⇔ its exact `(trackedApi, callerClass, callerMethod)` triple was matched |

Matching on the full origin path is what makes delegated JDK calls attributable: for `Class#getResource` the agent registers the configuration on the delegated `ClassLoader#getResource` frame, so any immediate-caller scheme sees `caller_class=java.lang.Class` and can never credit the library caller. The origin path retains the whole chain (`lib caller → Class#getResource → ClassLoader#getResource[config]`), so the matcher walks up from the tracked API to the nearest reported caller. The walk stopping at nested tracked APIs prevents the reverse error — the delegated `ClassLoader#getResource` hop can never be credited to a caller that actually invoked `Class#getResource`. Matching is per exact tracked-API signature plus caller class **and method**, so coverage cannot leak across APIs sharing a method name or across call sites sharing a class.

Both consumers are wired: `generateDynamicAccessCoverageReport` (strategy guidance) and `computeVersionStats` (published PR stats), so the numbers agree. Existing signatures are untouched; the matched call-site set threads through new overloads defaulting to `Set.of()`.

## What it changes in practice

- `org.hsqldb:hsqldb:2.7.3`: **0/82 → 63/82** in a full Forge run (#9010), including every delegated-resource call site the generated tests exercise — e.g. `Class#getResource ← org.hsqldb.lib.FileAccessRes.isStreamElement` and the deep `ResourceBundle#getBundle ← ResourceBundleHandler.getBundle` chain whose configuration lands 20+ frames down in `ServiceLoader` machinery.
- Precision spot-check: `Method#invoke ← ResourceBundleHandler.getBundle` (same caller method, unfired API) correctly stays uncovered.
- Regression tests: a realistic delegated `Class#getResource` origin fixture asserts nearest-caller attribution and the nested-tracked-API guard (`agentOriginsMatchDelegatedApiToNearestReportedCaller`), plus tracked-API parameter-spacing normalization.
- Jars with line info are untouched: the agent run and origin matching only ever start when line matching is impossible.

### DIFF
--- a/docs/tck.md
+++ b/docs/tck.md
@@ -129,8 +129,13 @@ carries a source line and JaCoCo covered that line. Some jars are compiled
 without a `LineNumberTable`, so every dynamic-access frame is line-less
 (`Unknown Source`) and line-based matching can never succeed. In that case only
 — the report has call sites but none carries a line number — the harness
-collects a `native-image-agent` trace from a JVM `test` run and falls back to
-matching a call site by the tracked-API method name plus the caller class: a
-call site is covered when a trace event for the same JDK entry-point method name
-fired from exactly that caller class. Line-based matching stays the primary path
-and is unchanged for jars that carry line information.
+collects `native-image-agent` configuration origins from a JVM `test` run and
+falls back to matching the complete origin path against the statically reported
+call sites. The harness streams the agent's compressed origin tree and marks a
+line-less call site covered when the same path contains both its exact tracked
+API and its caller class and method, with the caller preceding the API. It keeps
+only the current origin-tree branch and the matched call-site identities in
+memory. If multiple reported callers precede the API, the nearest one wins; a
+nested tracked API ends the caller search so delegated JDK calls do not cover a
+different API with the same method name. Line-based matching stays the primary
+path and is unchanged for jars that carry line information.
--- a/docs/tck.md
+++ b/docs/tck.md
@@ -123,3 +123,14 @@ These emit the GitHub Actions matrices the workflows consume, all driven by
 | `generateDynamicAccessCoverageReport`, `analyzeExternalLibraryDynamicAccess` | Dynamic-access coverage reporting (§FS-repository-functional-spec.4.5). |
 | `generateLibraryStats`, `listTopCoordinatesByMetric`, `generateTopCoordinatesByMetricMatrix`, `generateReadmeBadgeSummary`, `generateDependencyGraph` | Produce and query the stats mirror, README badge inputs, and dependency graphs that feed the coverage dashboard (§CI-publish-scheduled-coverage). |
 | `package` | Zip the `metadata/` directory into the release artifact consumed by native-build-tools (§FS-repository-functional-spec.4). |
+
+Dynamic-access coverage normally marks a call site covered when its stack frame
+carries a source line and JaCoCo covered that line. Some jars are compiled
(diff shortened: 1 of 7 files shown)


jormundur00 (changes requested):
### Blocking issue

The agent trace's `caller_class` does not reliably identify the class containing the original dynamic-access call site. `LibraryStatsSupport` marks a line-less site covered only when that field equals the dynamic-access frame class.

For example, on GraalVM JDK 25, calling `Class#getResource` from an application class produced a trace event with `function: "getResource"` and `caller_class: "java.lang.Class"`: the event is emitted from the JDK's `Class` -> `ClassLoader` delegation, not from the application caller. The existing fixture already models `java.lang.Class#getResource` with an application frame (`com.example.Bar.lookup`). Consequently, every such line-less resource call site remains uncovered even when exercised, so the fallback does not meet its all-call-site coverage contract.

Please correlate using provenance that preserves the original caller, or constrain the fallback to APIs for which the trace caller is demonstrably the original caller, and add a regression test using an actual agent trace.

jormundur00 (approved):
Approved. Reviewed the line-less dynamic-access fallback locally: focused LibraryStatsSupport tests and Checkstyle passed, the diff is whitespace-clean, and a GraalVM agent probe confirmed the origins option emits reachability-metadata-origins.txt. The fallback gate and origin-path matching are consistent with §TCK-test-harness.8.

PR #9041, opened 2026-07-13
Add Codex fallback for metadata validation

### LINKED ISSUE #9040: Add Codex fallback for metadata validation
When deterministic allowed-package repair cannot resolve a checkMetadataFiles failure, run one Codex metadata fix, then rerun validation. This preserves the deterministic repair for ordinary missing packages while recovering rare generated-metadata schema failures.

kimeta:
Summary:
- retain deterministic missing allowed-package repair
- run up to three Codex metadata repair attempts for remaining checkMetadataFiles failures
- limit each Codex invocation to 20 minutes
- write combined Codex output to a printed coordinate-scoped metadata-fix log
- rerun metadata validation after each repair

Tests were not run per request.

Fixes #9040

### DIFF
--- a/forge/docs/workflows/workflow-drivers.md
+++ b/forge/docs/workflows/workflow-drivers.md
@@ -245,7 +245,8 @@ re-implement the status merge at their call sites.
 Per-coordinate finalization repairs missing `allowed-packages`
 deterministically. If `checkMetadataFiles` still fails, Forge runs up to three
 Codex metadata fix and validation attempts before failing the run. Each Codex
-invocation is limited to 20 minutes.
+invocation is limited to 20 minutes, and its combined output is written to a
+printed coordinate-scoped metadata-fix log path (§FS-durable-generation-logs).
 
 Run metrics flow through the shared writer
 (`metrics_writer.write_workflow_run_metrics`): it appends the run entry to the
--- a/forge/docs/workflows/workflow-drivers.md
+++ b/forge/docs/workflows/workflow-drivers.md
@@ -242,6 +242,11 @@ finalization succeeds, otherwise the finalization status becomes the run
 status. Drivers must not call the private finalization internals and must not
 re-implement the status merge at their call sites.
 
+Per-coordinate finalization repairs missing `allowed-packages`
+deterministically. If `checkMetadataFiles` still fails, Forge runs one Codex
+metadata fix, then reruns test-only metadata splitting and validation before
+failing the run.
+
 Run metrics flow through the shared writer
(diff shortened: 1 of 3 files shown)


jormundur00 (changes requested):
I found two blocking issues in the new Codex fallback:

1. After Codex edits the repository, finalization reruns only `checkMetadataFiles`. It does not rerun `splitTestOnlyMetadata`, even though #9040 explicitly requires the post-repair split. This can leave test-only entries in shipped metadata. The legacy test-config policy is also checked only before Codex, so configuration introduced by the agent can escape that gate. Please rerun the split and post-split policy checks after each successful Codex attempt, before accepting validation.

2. `_run_codex_check_metadata_fix` invokes `codex exec` directly and bypasses the existing metadata-fix infrastructure. That loses exact GraalVM/JAVA_HOME pinning and version instructions, durable task-scoped logs, and the bounded timeout. The fallback can therefore reproduce under a different GraalVM, leave no diagnostic transcript, or block finalization indefinitely. Please reuse or generalize the shared Codex runner rather than duplicating a raw subprocess invocation.

Please also add focused `test_library_finalization.py` coverage for initial validation failure, successful Codex recovery with the post-repair split, retry exhaustion, and the three-attempt bound. The current tests never enter the new fallback branch.

kimeta:
@jormundur00 Thanks. I added a 20-minute timeout to the Codex process.

I don’t think another splitTestOnlyMetadata run is necessary. It already runs before validation, and
#9040 only requires rerunning validation after the repair. This Codex invocation is specifically
scoped to fixing checkMetadataFiles, it should not regenerate metadata or introduce test-only entries.

The subprocess also inherits the existing process environment. The workflow noralizes `GRAALVM_HOME`
and `JAVA_HOME` before finalization, so they are already available to Codex. Since this is schema
validation rather than Native Image failure reproduction, the specialized GraalVM-pinned metadata-fix
runner does not seem appropriate here.

Unit tests for this type of change isn't necessary.

jormundur00 (approved):
Thanks for the clarification. I agree that this fallback is scoped to repairing checkMetadataFiles, so another splitTestOnlyMetadata run and specialized native-image/GraalVM pinning are unnecessary here. I also do not consider focused unit tests required for this change.

The 20-minute timeout addresses the unbounded-process concern. One non-blocking concern remains: this direct codex exec invocation does not print or record a task-scoped log path, as required by the Forge logging contract. Please capture combined Codex output under the coordinate metadata-fix log directory and print that path when the attempt starts.

PR #9060, opened 2026-07-16
Preserve dynamic-access class checkpoints

### LINKED ISSUE #9062: Dynamic-access rollback loses committed class progress
## Problem

Iterative dynamic-access exploration commits accepted progress after each resolved or partially resolved class. If a later dynamic-access report refresh or native-test gate fails, the workflow can return failure and reset the feature branch all the way to the scaffold commit.

Failed-run preservation still keeps exploration information such as processed or exhausted classes. A resumed run can therefore treat earlier classes as already handled even though scaffold rollback removed the generated tests that justified that state. This loses useful work and leaves the resume checkpoint inconsistent with the preserved exploration information.

For example:

1. Class A gains coverage and its tests are committed.
2. Exploration moves to class B.
3. The report refresh or native-test gate fails while processing class B.
4. The workflow resets to scaffold, removing the class A commit.
5. Failed-run continuation preserves class A as already processed, so resume may skip work whose generated tests no longer exist.

## Expected behavior

After iterative exploration starts, failure recovery should reset to the latest committed class checkpoint. That commit is the last coherent point from which exploration can resume because it keeps accepted class progress and the corresponding exploration information aligned while discarding only later uncommitted work.

The scaffold commit remains the recovery point when no class has been committed. Bulk dynamic-access exploration keeps its existing scaffold-checkpoint behavior.

kimeta:
## What this does

Fixes: #9062

An iterative dynamic-access run could commit several completed classes, then lose its refreshed coverage report or fail a native gate and reset the entire branch to scaffold. That made resumable runs preserve exploration state while throwing away the corresponding generated tests.

The class loop already defines the durable unit of progress, so failure recovery now follows that same boundary. The latest class commit is the last coherent checkpoint from which exploration can resume: it keeps accepted class progress and the corresponding exploration information aligned while discarding only later uncommitted work. This updates [§WF-dynamic-access-fallback-and-failure](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/workflows/dynamic-access.md#WF-dynamic-access-fallback-and-failure).

## Recovery boundary

| Situation | Recovery checkpoint |
| --- | --- |
| No class committed yet | Scaffold commit |
| Resolved or partial class committed | Latest class commit |
| Later report refresh or native gate fails | Latest class commit |

### DIFF
--- a/forge/ai_workflows/core/dynamic_access_iterative_strategy.py
+++ b/forge/ai_workflows/core/dynamic_access_iterative_strategy.py
@@ -806,7 +806,10 @@ class DynamicAccessIterativeStrategy(WorkflowStrategy):
             f"native-test gate {result.status} for {class_name} after {result.iterations_used} cycles",
             indent_level=2,
         )
-        self._commit_library_metadata(f"Native metadata for {class_name}")
+        self._commit_test_sources(f"Native-test gate fixes for {class_name}")
+        self._latest_class_checkpoint = self._commit_library_metadata(
+            f"Native metadata for {class_name}"
+        )
         return True
 
     def _refresh_report_after_gate(self, class_name: str):
@@ -1021,8 +1024,13 @@ class DynamicAccessIterativeStrategy(WorkflowStrategy):
             text=True,
         ).strip()
 
-    def _commit_library_metadata(self, message: str) -> None:
-        """Stage and commit durable library metadata created by the class gate."""
+    def _commit_library_metadata(self, message: str) -> str:
+        """Stage and commit durable library metadata created by the class gate.
+
+        Returns the resulting `HEAD` SHA so a passing gate can advance the
+        whole-phase recovery checkpoint past the verified test and metadata
+        commits, per §WF-dynamic-access-fallback-and-failure.
+        """
         metadata_version = resolve_metadata_version(
             self.reachability_repo_path,
             self.group,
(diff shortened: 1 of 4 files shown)


jormundur00 (changes requested):
Found one blocking issue in the checkpoint recovery change:\n\nIn chunked runs,  is updated to the test/metadata commit, but  then creates a newer commit for the durable exhaust report. If a later failure reaches the whole-phase reset, it resets to the stale checkpoint and drops that completed-class-state commit. That reintroduces the continuation/checkpoint mismatch this PR is intended to fix.\n\nPlease advance the recovery checkpoint after the exhaust-report commit as well (ideally return its SHA from the helper), and add a chunked failure regression test.

forge/ai_workflows/core/dynamic_access_iterative_strategy.py
```diff
@@ -463,7 +466,7 @@ def flush_native_test_batch(reason: str, active_class_name: str | None = None) -
                 if updated_class is None:
                     if current_report.covered_calls > previous_report.covered_calls:
                         self._print_dynamic_access_detail("result: resolved", indent_level=2)
-                        self._commit_test_sources(
+                        self._latest_class_checkpoint = self._commit_test_sources(
```
jvukicev, line 469: The stale value is `_latest_class_checkpoint`: it is set before `_mark_chunk_class_completed()` creates the exhaust-report commit. A later whole-phase reset therefore removes that newer completed-class state. Please advance `_latest_class_checkpoint` after the exhaust-report commit too.

jormundur00 (approved):

PR #9121, opened 2026-07-28
JNA: add reflection metadata for direct mapping (Native.register)

ArloL:
## What does this PR do?

Adds the reflection metadata JNA needs for **direct mapping**
(`Native.register`), and a test that covers it. The entries come from
`./gradlew generateMetadata -Pcoordinates=net.java.dev.jna:jna:5.8.0`.

The existing metadata covers interface mapping (`Native.load`), where the
libffi call descriptors are built in native code. Direct mapping builds them in
Java and instantiates several types reflectively, none of them registered, so a
native image using it fails at run time:

```
java.lang.NoSuchMethodException: com.sun.jna.Structure$FFIType.<init>()
    at com.sun.jna.Klass.newInstance(Klass.java:48)
    at com.sun.jna.Structure.newInstance(Structure.java:1976)
    at com.sun.jna.Structure$FFIType.<clinit>(Structure.java:2077)
```

```
java.lang.NoSuchMethodException: com.sun.jna.NativeLong.<init>()
    at com.sun.jna.Klass.newInstance(Klass.java:48)
    at com.sun.jna.NativeMappedConverter.defaultValue(NativeMappedConverter.java:65)
    at com.sun.jna.Native.register(Native.java:1965)
```

The first is `Structure$FFIType`'s static initializer, which creates the
pre-defined libffi types through `Klass.newInstance()` and then reads their
fields — and the `size` field is itself a `Structure$FFIType$size_t` built the
same way. The second is `NativeMappedConverter`, which instantiates the
`NativeMapped` parameter type to derive its default value.

No existing test uses direct mapping, so this adds one that registers `atol`
and `labs` and calls them; `labs` takes and returns `NativeLong`, which is what
routes registration through `NativeMappedConverter`. Removing any of the three
metadata entries makes it fail.

Not a regression from a recent release — the reflective code is unchanged since
5.8.0, and the test fails on 5.8.0 too.

Two deliberate deviations from the raw agent output. The conditions are
normalised to `com.sun.jna.Native`, which the surrounding entries already use,
because the agent's computed condition is not stable across runs —
`Structure$FFIType$size_t` was recorded under `IntegerType` and
`NativeMappedConverter` in one run and under `Native` and `Structure` in the
next. And the agent's bare `Structure$FFIType` entry is dropped, since the
entry carrying the members registers the type as well.

Verified on GraalVM CE 25.0.2 (aarch64): `./gradlew test` passes for 5.8.0 and
5.18.1, plus `checkMetadataFiles` and `checkstyle`. Reverting only the metadata
change makes the new test fail while the existing one still passes. 5.19.1 also
passes, but I left `tested-versions` alone — that is a separate change, and
5.19.x still hits the unrelated `future-defaults-all` failure in #7741.

- `tests/.../net_java_dev_jna/jna/DirectMappingTest.java` binds the system C
  library via `Native.register(DirectMappedCLibrary.class, "c")` and calls
  `atol` and `labs` — the same dependency the existing test already has through
  `Native.load("c")`. No network, Docker or filesystem access.

### DIFF
--- a/metadata/net.java.dev.jna/jna/5.8.0/reachability-metadata.json
+++ b/metadata/net.java.dev.jna/jna/5.8.0/reachability-metadata.json
@@ -167,6 +167,18 @@
       "condition": {
         "typeReached": "com.sun.jna.Native"
       },
+      "type": "com.sun.jna.NativeLong",
+      "methods": [
+        {
+          "name": "<init>",
+          "parameterTypes": []
+        }
+      ]
+    },
+    {
+      "condition": {
+        "typeReached": "com.sun.jna.Native"
+      },
       "type": "com.sun.jna.NativeMapped",
       "jniAccessible": true,
       "methods": [
@@ -255,6 +267,32 @@
       "condition": {
         "typeReached": "com.sun.jna.Native"
       },
+      "type": "com.sun.jna.Structure$FFIType",
+      "fields": [
+        {
+          "name": "alignment"
+        },
+        {
+          "name": "elements"
+        },
+        {
+          "name": "size"
+        },
+        {
+          "name": "type"
+        }
+      ],
+      "methods": [
+        {
+          "name": "<init>",
+          "parameterTypes": []
+        }
+      ]
+    },
+    {
+      "condition": {
+        "typeReached": "com.sun.jna.Native"
+      },
       "type": "com.sun.jna.Structure$FFIType$FFITypes",
       "jniAccessible": true,
       "fields": [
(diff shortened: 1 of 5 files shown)


jvukicev:
Hello @ArloL , thanks for looking to contribute to the repository!

Before we can merge this PR, I would ask you to please regenerate this libraries `stats.json` (located at https://github.com/oracle/graalvm-reachability-metadata/blob/master/stats/net.java.dev.jna/jna/5.8.0/stats.json) and push the regenerated file to this PR in case dynamic access coverage changed with the new tests. You can do this by running the `generateLibraryStats` gradle task (see https://github.com/oracle/graalvm-reachability-metadata/blob/master/docs/DEVELOPING.md). While this "should" be functional on Graal `25.0.x`, I'd recommend using a `25.1.x`+ Graal, as it has all the dynamic access detection features.

If you are unable to use `25.1.x` for some reason, feel free to ping me and I will generate and push this file for you.

ArloL:
I just pushed the updated stats.json. Hope it's correct :)

kimeta (approved):

jormundur00 (approved):
Thanks for the contribution, LGTM.

PR #9248, opened 2026-08-04
Execute the JVM tests against the Crema JVM, weekly and on demand

### LINKED ISSUE #9247: Run the library test suites on the Crema JVM to find run-time class loading gaps
## Motivation

Crema is Native Image's run-time class loading VM (`-H:+RuntimeClassLoading`), shipped as a native-image-built `libjvm` that the ordinary `java` launcher can select. This repository already carries ~2184 library test projects exercising real-world reflection, serialization, resource loading, and service lookup — which makes it an unusually good corpus for finding where Crema cannot yet run real code.

The goal here is **finding Crema bugs**, not gating metadata. Nothing this sweep reports should block a release.

## How the JDK is obtained

GraalVM early-access builds ship `lib/svm/macros/jvm-library/native-image.properties` and `lib/graalvm/svm-libjvm.jar`, and already list `-svm KNOWN` in `lib/jvm.cfg`, but deliberately omit the built library. One `native-image --macro:jvm-library` produces `lib/svm/libjvm.so`, after which `java -svm` reports `Substrate 64-Bit Server VM`. Listing `-svm` first in `jvm.cfg` makes that the JDK's default VM, so no caller needs a VM selection flag.

Verified locally against `graalvm-jdk-25i3-25.0.4-ea.01` (`GRAALVM_VERSION=25.3.4-dev`): the macro build takes under three minutes. Note the public `graalvm-ce-dev-builds` artifacts do **not** ship the macro — only builds including the `svmjava` component do.

## Findings already produced by standing this up

1. **Gradle cannot run on Crema.** `AbstractMethodError` at `org.gradle.launcher.cli.BuildActionsFactory.getBuildStartTime`, with `com.oracle.svm.core.reflect.CremaMethodAccessor.invoke` frames in the stack. This is why the sweep runs Gradle on a stock JDK and moves only the forked test workers to Crema.
2. **`-ea` is rejected at VM startup**: `Error: The option '-ea' is not supported by Native Image`. Gradle enables assertions on every test worker by default, so every coordinate fails identically before a single test runs. This one blocks the whole sweep and is likely the highest-value fix — it blocks any JVM test framework, not just this repository.
3. `-javaagent` is accepted but ignored, so JaCoCo silently yields empty coverage rather than failing.
4. With assertions cleared, `org.yaml:snakeyaml:2.0` passes its JVM tests on Crema — so the corpus does run once past (2).

## Scope

- A composite action that builds the Crema JDK from the macro and selects the VM.
- A dispatch-only workflow that sweeps the `javaTest` lane, attributes failures per coordinate, and publishes NDJSON plus logs.
- Harness support for pointing the JVM test lane at a JDK that cannot host Gradle.

No Crema-specific JVM flags are added anywhere: a library that fails because Crema rejects an argument the lane normally passes is a finding, not something to work around.

vjovanov:
Sweeps the repository's library test suites on Crema — Native Image's run-time class loading VM — to find where it cannot yet execute real code. Bug-finding aimed at Crema, not a metadata gate: it never blocks a release.

## What is here

- **`.github/actions/setup-crema-jdk`** — builds `lib/svm/libjvm.so` from the `jvm-library` macro that GraalVM EA builds ship but leave unbuilt, then lists `-svm` first in `lib/jvm.cfg` so a plain `java` from that JDK is Crema and no caller needs a VM flag. Every Crema build option comes from the macro and from `svm-libjvm.jar`'s own `native-image.properties`; the action adds none. It fails loudly when the macro is missing, because a silently-HotSpot run would report a meaningless clean pass.
- **`.github/workflows/test-all-metadata-crema.yml`** — weekly (Saturday 02:00, clear of the Sunday metadata sweep and the Monday release) and on demand. The JDK is built **once** and shared with all 85 shards as an artifact; per-shard builds cost ~10 minutes each, about 14 hours of runner time. One JDK serves both lanes — the action leaves `jvm.cfg.hotspot-default` beside the Crema-default `jvm.cfg`, so a shard picks its VM by choosing between two files — and every shard re-asserts the VM identity after unpacking rather than trusting the build job.
- **`GVM_TCK_TEST_JAVA_HOME`** (§TCK-test-harness.3.1) — points the forked JVM test workers at a JDK that need not be able to host Gradle. Unset by default, so nothing changes for existing runs.

## Why the workers and not Gradle

Gradle cannot run on Crema: `AbstractMethodError` at `org.gradle.launcher.cli.BuildActionsFactory.getBuildStartTime`, with `CremaMethodAccessor.invoke` frames in the stack. So Gradle stays on a stock JDK and only the test workers move. The variable selects an interpreter and never adds, removes, or rewrites a JVM argument — that is what keeps a failure attributable to the JDK rather than to the harness.

## The one deliberate deviation

Crema rejects Gradle's default `-ea` at VM startup (`The option '-ea' is not supported by Native Image`), which fails every worker before a single test runs. `GVM_TCK_TEST_DISABLE_ASSERTIONS` / the `disable-assertions` input clears it so the sweep can reach the failures behind that one. It defaults to **on**, because leaving `-ea` in place yields one finding repeated 2184 times instead of a survey; set it false to re-check whether the blocker still exists. With assertions off, tests that verify via `assert` pass vacuously — a green coordinate means Crema ran the code, never that the library is supported.

## Validation

All runs on a fork, since `workflow_dispatch` requires the workflow on a default branch.

| Run | Scope | Result |
| --- | --- | --- |
| [30983741655](https://github.com/vjovanov/graalvm-reachability-metadata/actions/runs/30983741655) | 1 coordinate, `vm=both` | ✅ 3/3 jobs — matrix, crema lane, hotspot lane |
| [30987005574](https://github.com/vjovanov/graalvm-reachability-metadata/actions/runs/30987005574) | full sweep, 85 shards | ✅ 80 shards passed; 1 lost to Maven Central HTTP 429; 4 cancelled |
| [30991829016](https://github.com/vjovanov/graalvm-reachability-metadata/actions/runs/30991829016) | build-once, `vm=both` | ✅ 4/4 jobs — matrix, JDK build, crema lane, hotspot lane |
| [30995271162](https://github.com/vjovanov/graalvm-reachability-metadata/actions/runs/30995271162) | full sweep, 85 shards | ✅ 81 shards passed, matrix + JDK build passed; 1 lost to HTTP 429; 3 cancelled by a later dispatch |

Evidence the runs measured what they claim:

- `native-image --macro:jvm-library` produced a 302–359 MiB `libjvm.so`; build options are attributed in the log to `origin(s): macro option 'jvm-library@…'`.
- The crema lane reports `Substrate 64-Bit Server VM`; the hotspot lane reports `Java HotSpot(TM) 64-Bit Server VM`. Both guards are enforced, so a shard cannot silently measure the wrong VM.
- Running the JVM lane with `-ea` intact fails at VM startup, which is independent proof that workers really execute on Crema.
- The JDK is unpacked to a different absolute path than it was built at, and both lanes run correctly from it.

**Roughly 2000 library coordinates pass their JVM test suites on Crema.** The only failures in either sweep were Maven Central rate limiting, not Crema defects.

## Known limitations

- **Maven Central HTTP 429.** 85 shards resolving concurrently draw rate limiting, which fails a shard during dependency resolution and loses the ~26 coordinates behind it. The latest commit stretches Gradle's retry backoff on both retry layers via `GRADLE_USER_HOME/gradle.properties`; an earlier `GRADLE_OPTS` form was measured to have no effect, because that variable configures the client JVM while resolution runs in the daemon. **The corrected form has not yet been validated under load.** Caching the Gradle dependency cache across shards and capping `max-parallel` would address the cause rather than the symptom; neither is included here.
- **Failures are not classified.** A 429 landing mid-coordinate rather than at shard startup would be recorded as a Crema failure. Both observed 429s hit at startup and produced no findings, so current results are clean, but failure entries deserve a glance before being filed upstream.
- **No HotSpot baseline in the sweep runs above.** The `vm` input supports `crema`, `hotspot`, or `both`; the full sweeps ran `crema` only. A baseline lane distinguishes Crema defects from coordinates that fail under any VM.

Fixes #9247


dougxc (approved):

jormundur00 (approved):

.github/actions/setup-crema-jdk/action.yml
```diff
@@ -0,0 +1,105 @@
...
+      shell: bash
+      env:
+        INPUT_DEFAULT_VM: ${{ inputs.default-vm }}
+      run: |
+        set -euo pipefail
+
+        CREMA_HOME="$JAVA_HOME"
+        MACRO="$CREMA_HOME/lib/svm/macros/jvm-library/native-image.properties"
+
+        # Fail loudly rather than silently producing a HotSpot-only JDK: a sweep that reports a clean
```
gilles-duboscq: Do we really need that? `native-image --macro:jvm-library` already fails loudly if this file is missing:
```
Error: Unknown name in option specification: macro:jvm-library
Available macro options are:
    --macro:native-image-agent-library
    --macro:native-image-diagnostics-agent-library
...
```
It exits with exit code 20 and we have `set -e` above.

gilles-duboscq (approved):

.github/actions/setup-crema-jdk/action.yml
vjovanov: Good point — `native-image` does exit non-zero on an unknown macro and `set -e` catches it. Dropped the explicit check.

PR #9295, opened 2026-08-13
Drain issue queues by priority tier

### LINKED ISSUE #9294: Priority ordering applies only within a fetched scan batch
Forge ranks issues by priority only *inside* one fetched scan batch. `get_prioritized_issues_with_label` pulls a single page of `DEFAULT_ISSUE_SCAN_BATCH_SIZE` (25) issues with a plain label query and sorts those 25 by `high-priority` / `priority` / neither. Priority is therefore a tie-breaker among 25 arbitrary issues, not a queue order.

The practical consequences:

- A `high-priority` issue that sits at position 40 of a queue is picked up only after Forge has worked through 39 issues of ordinary work. With one issue claimed per run, that is many runs of latency on exactly the work that was marked urgent.
- The queue is sorted `updated desc`, so an old but urgent issue drifts further back every time anything else in the queue is touched.
- The scan window is capped at GitHub Search's first `GITHUB_SEARCH_MAX_RESULTS` (1000) matches of one flat query. A `high-priority` issue past that cap is unreachable regardless of its label.

Forge should drain the queue by urgency tier instead: every `high-priority` issue first, then every `priority` issue, then the rest — across the whole queue rather than within a batch, matching the ordering §FS-forge-issue-resolution-goal already promises.

Scans started from a random offset (the mode that spreads concurrent runners across a queue) should keep their current flat behaviour.

kimeta:
Closes #9294

## What this does

Priority was only a tie-breaker inside one fetched batch: `get_prioritized_issues_with_label` pulled 25 issues with a flat label query and sorted those 25. A `high-priority` issue at position 40 of a queue waited behind 39 ordinary ones, and — because the flat query is capped at GitHub Search's first `GITHUB_SEARCH_MAX_RESULTS` matches — one past position 1000 was unreachable no matter its label. This makes the scan drain the queue tier by tier instead, which is the ordering [§FS-forge-issue-resolution-goal](forge/docs/functional-spec.md) already promised and [§ORCH-forge-orchestration-spec](forge/docs/orchestration-scripts.md) now describes.

## How a queue is drained

Three searches, consumed in order, each excluding the labels of the tiers above it so an issue carrying both labels is served exactly once, in its highest tier:

| Tier | Query |
|---|---|
| `high-priority` | `label:"high-priority"` |
| `priority` | `label:"priority" -label:"high-priority"` |
| regular | `-label:"high-priority" -label:"priority"` |

A tier is left only once its search returns nothing, so its result window is now capped independently rather than sharing one flat window. `IssueQueueScanState` carries the tier index and the offset within it; an empty return means every tier is drained, so the scan loop stops on that signal alone.

With 13 `high-priority` and 32 `priority` issues, a scan walks 13 → 25 → 7 → regular in batches of 25. Because pagination fetches 100 at a time and slices locally, that is one API call per tier under 100 issues — the tier-boundary probes re-slice a page the search cache already holds.

Scans started from a random offset, which spread concurrent runners across a queue, keep paging the flat unfiltered query.

### DIFF
--- a/forge/README.md
+++ b/forge/README.md
@@ -32,6 +32,7 @@ Common options:
 - `--parallelism N`: run up to `N` issue workflows in parallel. Maximum: 4.
 - `--review-limit N`: process up to `N` PR review tasks per label per cycle.
 - `--random-offset`: start new-library issue scans at a random offset instead of the newest issues first.
+- `--priority {high,priority,normal}`: process only the selected issue priority tier.
 - `--user-requested-only`: fetch only user-requested issue queue items, excluding configured automation and maintainer authors.
 - `--once`: run a single update/work cycle through `do_up_to_date_work.sh` and exit.
 - `--stop`: ask all Forge `do-work` loops for the current user to exit by creating `~/.metadata-forge-stop`.
--- a/forge/do_up_to_date_work.sh
+++ b/forge/do_up_to_date_work.sh
@@ -21,6 +21,7 @@ LIBRARY_UPDATE_WORK_STRATEGY_NAME="${FORGE_LIBRARY_UPDATE_STRATEGY_NAME:-}"
 WORK_LABEL="${FORGE_WORK_LABEL:-library-new-request}"
 WORK_LIMIT="${FORGE_WORK_LIMIT:-1}"
 RANDOM_WORK_OFFSET="${FORGE_RANDOM_WORK_OFFSET:-0}"
+PRIORITY_TIER=""
 PARALLELISM="${FORGE_PARALLELISM:-1}"
 REVIEW_LABEL="${FORGE_REVIEW_LABEL:-}"
 REVIEW_LIMIT="${FORGE_REVIEW_LIMIT:-1}"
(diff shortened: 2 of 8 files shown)


vjovanov:
How do I use it?

kimeta:
> How do I use it?

It is a default mechanism. Should we put it under the flag?

vjovanov:
Can I say to run only priority or high_priority?

kimeta:
> Can I say to run only priority or high_priority?

Yes, this is now supported with `--priority`. The accepted values are `high`, `priority`, and `normal` (so use `high`, not `high_priority`).

For the regular worker:

```bash
./do-work.sh --priority high
./do-work.sh --priority priority
./do-work.sh --priority normal
```

For a single queue directly:

```bash
python3 forge_metadata.py --label library-new-request --limit 1 --priority priority
```

Without `--priority`, Forge keeps the default behavior and drains all tiers in order: high, priority, then normal.

vjovanov (approved):

PR #9297, opened 2026-08-14
Add reachability metadata for Flyway Oracle module

graemerocher:
The Flyway Oracle database module currently requires application-provided native-image reflection configuration for its database type and configuration extension.

This adds the missing metadata under the Flyway Oracle module for:
- org.flywaydb.database.oracle.OracleDatabaseType
- org.flywaydb.database.oracle.OracleConfigurationExtension

The existing Flyway core metadata already covers the core baseline and SLF4J logging classes, so those entries are intentionally not duplicated here.

### DIFF
--- a/metadata/org.flywaydb/flyway-database-oracle/11.14.1/reachability-metadata.json
+++ b/metadata/org.flywaydb/flyway-database-oracle/11.14.1/reachability-metadata.json
@@ -8,6 +8,26 @@
       "type": "org.flywaydb.database.oracle.OracleConfigurationExtension",
       "allDeclaredConstructors": true,
       "allPublicMethods": true
+    },
+    {
+      "condition": {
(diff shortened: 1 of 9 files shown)


kimeta:
Hey @graemerocher, thanks for contributing!

Before we can merge this, the PR needs two things:

- A test for the new library. Every metadata directory here must be justified by a test that exercises the entries it ships. `tests/src/org.flywaydb/flyway-database-postgresql` is a good example to look at.

- Library coverage stats. Once the test is in place, please run `./gradlew generateLibraryStats -Pcoordinates=org.flywaydb:flyway-database-oracle:11.14.1` and commit the resulting files under `stats/`.

Would you like to take these on, or would you prefer that we add the test and stats?

graemerocher:
please add the test and stats

thanks

vjovanov (approved):

PR #9484, opened 2026-08-21
Fix compatible baseline selection for version backfills

### LINKED ISSUE #9481: Bug: Forge selects a cross-major latest suite as the baseline for version backfills
### Describe the bug

Forge resolves the baseline for repair and missing-version library-update workflows from the metadata entry marked `latest`, even when that entry belongs to a different major version than the requested version. This can turn a metadata/index repair into an unrelated compilation repair against an incompatible test suite.

Issue #9408 reproduces the problem for `io.netty:netty-common:4.1.132.Final`. The index marks `5.0.0.Alpha1` as `latest`, while the compatible Netty 4.1 metadata and tests live under `4.1.115.Final`. Forge therefore invoked the native-image repair as `5.0.0.Alpha1 -> 4.1.132.Final`.

`fixTestNativeImageRun` attempted metadata generation by running the Netty 5 test suite against 4.1.132. Compilation failed with 40 errors because classes under `io.netty.util.internal.chmv8` used by that suite do not exist in 4.1.132, so no `reachability-metadata.json` was generated and Codex repair was skipped. The later `library-update-request` compatibility probe selected the same Netty 5 baseline and incorrectly routed the issue to `fix_javac_fail` for the same compilation errors.

Relevant implementation points:

- `forge_metadata.load_current_metadata_version` selects the entry marked `latest` for direct repair queues.
- `ai_workflows/drivers/library_update_router.py::_latest_supported_entry` selects the latest usable suite for missing requested versions.
- The requested version may be a historical backfill or another maintained version line, so repository-wide latest is not necessarily a compatible predecessor.

### Steps to reproduce

1. Use an index where `io.netty:netty-common:5.0.0.Alpha1` is marked `latest` and `4.1.115.Final` owns the supported 4.1.x test suite.
2. Process a `fails-native-image-run` or missing-suite `library-update-request` for `io.netty:netty-common:4.1.132.Final`.
3. Observe that Forge chooses `5.0.0.Alpha1` as the baseline.
4. Observe `compileTestJava` fail on the copied Netty 5 tests before metadata repair can begin.

### Expected behavior

Forge should select a baseline compatible with the requested version line. For #9408 it should choose the `4.1.115.Final` entry/test suite, not the Netty 5 alpha entry. If no compatible baseline can be selected deterministically, Forge should stop with an actionable routing error instead of probing an unrelated major-version suite.

### Acceptance criteria

- Define the baseline-selection behavior in the most-specific Forge workflow spec before changing code.
- Direct repair queues and the missing-version library-update router use one shared compatible-baseline resolver.
- The resolver considers exact tested-version ownership, `default-for`, version ordering/line compatibility, and test-suite availability before falling back.
- A next-major prerelease marked `latest` is not selected for a stable historical backfill from another major line.
- Add fixture/unit coverage for the Netty 5 latest / Netty 4.1.132 backfill shape.
- Log the selected baseline and the reason it was selected.

vjovanov:
### Summary

- resolve version-backfill baselines from exact index ownership, then the nearest compatible same-line support
- use the shared resolver for direct repair queues, missing-version routing, and coverage backfills
- reject cross-major latest suites and fail with an actionable error when no compatible suite exists
- treat test-version strictly as a test-directory alias, not as a supported library release
- log the selected baseline coordinate and selection reason

### Testing

- python3 -B -m unittest tests.test_library_update_target tests.test_library_update_router tests.test_forge_metadata.LibraryUpdateIssueTests
- ruff check --no-cache utility_scripts/metadata_index.py tests/test_library_update_target.py
- grund check
- git diff --check
- hermetic Forge fixture E2E for issue 9108

Fixes #9481

### DIFF
--- a/forge/ai_workflows/drivers/improve_library_coverage.py
+++ b/forge/ai_workflows/drivers/improve_library_coverage.py
@@ -26,7 +26,7 @@ import re
 import shutil
 import subprocess
 import sys
-from typing import Any, Callable
+from typing import Any
 
 if __package__ in (None, ""):
     sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
@@ -64,6 +64,7 @@ from utility_scripts.metadata_index import (
     is_newer_than_latest_metadata_version,
     load_index_entries,
     resolve_library_update_target,
+    resolve_version_backfill_baseline,
 )
 from utility_scripts.metrics_writer import count_metadata_entries, count_test_only_metadata_entries, create_failure_run_metrics_output
 from utility_scripts.source_context import (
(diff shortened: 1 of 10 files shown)


kimeta:
I think only the `library-update-request` router actually needs the compatible-baseline resolver. The `fails-*` labels should mean the *newest* version failed somewhere, so `latest` is the correct baseline there more or less by definition; a `fails-native-image-run` issue for a version below `latest` looks like a bug in whatever filed it, not something baseline resolution should paper over. Wiring the resolver into `build_claim_metadata` risks hiding that signal.

vjovanov:
Addressed in 3d76bce37e. The compatible-baseline resolver now stays scoped to missing-version `library-update-request` routing. Direct `fails-*` claim metadata again uses the index entry marked `latest`, and the workflow spec plus regression test now preserve that producer-contract signal.

kimeta (approved):

PR #9504, opened 2026-08-24
Update layered tests exclude lists after GraalVM update

Zeavee:
This PR enables additional layered tests after the latest GraalVM nightly build.


Zeavee:
@vjovanov I have a new update for the layered exclude list. I might have them every couple weeks, is it ok to ping each time? Or is there another workflow we could use?

vjovanov (approved):

vjovanov:
Fine to ping me.

PR #9532, opened 2026-08-24
Restructure docs into per-kind spec and architecture homes

kimeta:
## What this does

The docs had drifted away from the rule that every grund kind gets its own home.
`README.md` files in `docs/functional-spec/` and `docs/architecture/` — in both
namespaces — had grown into the specs themselves: the root functional-spec README
was 509 lines of declarations, the Forge one 906. Meanwhile Forge kept four kind
homes (`DW`, `STRAT`, `ORCH`, `GIT`) loose in `docs/`, so the prefix on an ID no
longer told you which directory to open, and a reader landing on a README could
not tell whether it was an index or the source of truth.

This restructures both namespaces so a README is always navigational and never
citable, moves every kind home under the spec or architecture directory it
belongs to, and drops the deprecated-API rule from the test contract along the
way.

## Where the kinds live now

Every README became a pure index — a table of the IDs in that directory plus a
one-line description per file — and the content moved into a real kind home that
declares one top-level ID:

| Namespace | New home | Declares |
| --- | --- | --- |
| root | [`docs/functional-spec/functional-spec.md`](docs/functional-spec/functional-spec.md) | §FS-repository-functional-spec |
| root | [`docs/architecture/architecture.md`](docs/architecture/architecture.md) | §AR-repository-architecture |
| forge | [`forge/docs/functional-spec/functional-spec.md`](forge/docs/functional-spec/functional-spec.md) | §forge/FS-forge-functional-spec |
| forge | [`forge/docs/architecture/architecture.md`](forge/docs/architecture/architecture.md) | §forge/AR-forge-architecture |

Forge's loose kind homes moved under the directory that matches what they hold,
with `forge/grund.toml` updated to match:

```
docs/do-work.md                 -> docs/architecture/do-work.md
docs/orchestration-scripts.md   -> docs/architecture/orchestration-scripts.md
docs/git-scripts.md             -> docs/architecture/git-scripts.md
docs/strategies.md              -> docs/functional-spec/strategies.md
docs/benchmarking.md            -> docs/functional-spec/benchmarking.md
docs/workflows/                 -> docs/functional-spec/workflows/
```

`docs/` in the Forge namespace now holds only the four documents that answer a
question about Forge as a whole: why (`GRUND`), where (`GOAL`), what is planned
(`ROADMAP`), and how it is tested end to end (`E2E`). `code-coverage-signal-quality.md`
is deleted — a design note from the coverage work that declared no ID and had
been overtaken by §forge/WF-code-coverage-improvement.

## Setup is checked by its artifacts, not by its report

§forge/WF-forge-workflow-drivers.1 specified `NeuralSetupResult` as a
schema-validated JSON document that `check_setup` validated on read and *then*
confirmed against the worktree. That makes the agent's account of its work the
primary artifact and the tree a second opinion, which is backwards under
§PRCPL-verify-inputs.

Each setup step already writes a real artifact to a place fixed by the coordinate
and the run context, so `check_setup` now opens those directly:

| Step | Artifact | Properly generated when |
| --- | --- | --- |
| Artifact URL population | `metadata/<group>/<artifact>/index.json` | it parses and the entry carries its four URLs |
| Source context | the directory that entry names | it exists and holds the requested sources |
| Preparation preflight | the run's preflight JSON | it parses and validates against its schema |
| Applied actions | test `build.gradle`, `allowed-docker-images` | each decided action is present in the tree |

`NeuralSetupResult` is reduced to the set of those paths — enough for continuation
state and metrics to point at the same files — and `check_setup` never takes a
path from it that it can derive itself, so nothing the agent returns can move
where the check looks. §forge/ROADMAP-forge-algorithmic-then-neural-setup and
§forge/AR-forge-architecture are updated to match.

## Open questions

The directory layout here is a working proposal, not a settled convention — the
FS/AR split in particular forced a coin flip on a few docs that are both. Let's
talk it through before this ossifies by being merged.


.github/actions/detect-file-changes/detect-file-changes.js
```diff
@@ -1,5 +1,6 @@
-// Lightweight in-repository file-change filter for GitHub pull request workflows.
 // §CI-detect-file-changes.
+
```
vjovanov, line 2: This is a bug in grund. It believes this is an inline comment and then breaks it up.

This is the highest priority issue to fix.

docs/architecture/README.md
```diff
@@ -1,81 +1,46 @@
...
+exposes and CI is that same task surface driven by GitHub Actions — both are
+architecture. They keep their own prefixes: a `TCK-` or `CI-` ID is not an `AR-`
+ID, and each is declared in its own home file above rather than anywhere in this
+folder.
+
+| ID | Subject |
+| --- | --- |
+| [§AR-repository-architecture](architecture.md#ar-repository-architecture-repository-architecture-and-high-level-overview) | Repository architecture and high-level overview |
+| [§AR-build-infrastructure](build-infra.md#ar-build-infrastructure-build-infrastructure) | Build infrastructure |
+| §TCK-test-harness | Test harness (TCK) task groups |
```
vjovanov: Make them all links. New grund should catch this?

Aha, it doesn't as it is linked from above.

docs/architecture/README.md
```diff
@@ -1,81 +1,46 @@
...
-- **Schemas everywhere.** Metadata, index, stats, and the framework list each
-  validate against a vendored JSON schema. Minor and patch schema bumps stay
-  backward-compatible so older consumers keep working; a major bump represents a
-  change native-build-tools must adopt, released in lockstep with the plugin
-  (§METADATA-suite, §FS-repository-functional-spec.4).
-- **Separation of automation.** Forge is an independently checked subproject in
-  its own grund namespace; from the repository's side it is just another
-  contributor bound by the same constraints and gates
-  (§forge/AR-forge-architecture, §FS-repository-functional-spec.4).
+| `AR` | this folder | The structural map and the build wiring. |
```
vjovanov: Do you even need this?

grund.toml
```diff
@@ -25,64 +25,113 @@ number_pattern = "\\d+"
...
+
+[[kinds]]
+kind = "action"
+folder = ".github/actions"
+citable = false
+title = "Composite actions the workflows reuse"
+
+# Places that are listed, not walked: content that ships verbatim or is generated.
+[[kinds]]
+kind = "metadata-suite"
```
vjovanov: Not really a suite, right?

grund.toml
```diff
@@ -129,5 +175,11 @@ must = ["FS|METADATA|GOAL"]
...
 
+[citations.workflow]
+should = ["CI|FS|AR|TCK|METADATA|TESTS"]
+
+[citations.action]
+should = ["CI|FS|AR"]
+
 [citations.code]
-must = ["FS|AR|TCK|E2E|CI|METADATA|TESTS"]
+must = ["FS|AR|TCK|CI|METADATA|TESTS"]
```
vjovanov: Code should not cite tests and METADATA, nor the CI.
```toml
must = ["FS|AR"]
```

grund.toml
```diff
@@ -129,5 +175,11 @@ must = ["FS|METADATA|GOAL"]
 [citations.SKILL]
 must = ["FS|CI|METADATA|TESTS"]
 
+[citations.workflow]
+should = ["CI|FS|AR|TCK|METADATA|TESTS"]
+
+[citations.action]
+should = ["CI|FS|AR"]
```
vjovanov: This should be an error. You can't cite non-citable kinds.

grund.toml
vjovanov: Or even better `must  = ["FS"]` and `should = ["AR"]`

vjovanov: Does this complain in grund? You can't cite a non-citable thing.

AGENTS.md
```diff
@@ -150,12 +148,13 @@ Declarations are heading lines `# FS-user-login: …` in markdown. In a code doc
...
 - **AR** must cite FS or GOAL.
 - **TCK** must cite FS or AR or GOAL.
-- **E2E** must cite FS or TCK or AR.
-- **CI** should cite FS or TCK or METADATA or TESTS or E2E or GOAL.
+- **CI** should cite FS or TCK or METADATA or TESTS or GOAL.
 - **METADATA** must cite FS or GOAL.
 - **TESTS** must cite FS or METADATA or GOAL.
-- **SKILL** must cite FS or CI or METADATA or TESTS.
-- **code** (any file outside a kind home) must cite FS or AR or TCK or E2E or CI or METADATA or TESTS.
+- **skills/** must cite FS or CI or METADATA or TESTS.
```
vjovanov: Skills shouldn't cite anything, right? They are shipped to people?

AGENTS.md
```diff
@@ -150,12 +148,13 @@ Declarations are heading lines `# FS-user-login: …` in markdown. In a code doc
...
+- **.github/workflows/** should cite CI or FS or AR or TCK or METADATA or TESTS.
```
vjovanov: Maybe AR only?

AGENTS.md
```diff
@@ -150,12 +148,13 @@ Declarations are heading lines `# FS-user-login: …` in markdown. In a code doc
 - **FS** should cite GOAL or FS.
```
vjovanov, line 144: This table is the key, it essentially explains how the memory is organized. It needs to be accurate.

AGENTS.md
```diff
@@ -109,12 +101,18 @@ A `§<ID>` is a pointer to a fact, not a file path. Resolve it with `grund` and
 - [PRCPL](docs/principles.md): Cross-cutting principles for how the repository works
 - [FS](docs/functional-spec): Repository functional behavior and contributor-facing requirements
 - [AR](docs/architecture): Repository architecture and build infrastructure
-- [TCK](docs/tck.md): Test harness (TCK): task groups
-- [E2E](docs/e2e.md): Infrastructure end-to-end tests (testInfra/testAllInfra)
-- [CI](docs/ci.md): Recurring CI workflows and composite actions
-- [METADATA](docs/metadata.md): The metadata/ suite: shipped reachability metadata
-- [TESTS](docs/tests.md): The tests/ suite: tests that justify the metadata
-- [SKILL](skills): Agent review and automation skills
+- [TCK](docs/architecture/tck.md): Test harness (TCK): task groups
```
vjovanov: Why is this not just AR?

AGENTS.md
```diff
@@ -109,12 +101,18 @@ A `§<ID>` is a pointer to a fact, not a file path. Resolve it with `grund` and
...
+- [CI](docs/architecture/ci.md): Recurring CI workflows and composite actions
```
vjovanov: Same here?

AGENTS.md
```diff
@@ -109,12 +101,18 @@ A `§<ID>` is a pointer to a fact, not a file path. Resolve it with `grund` and
...
+- [METADATA](docs/functional-spec/metadata.md): The metadata/ suite: shipped reachability metadata
```
vjovanov: This is just FS-metadata?

AGENTS.md
```diff
@@ -109,12 +101,18 @@ A `§<ID>` is a pointer to a fact, not a file path. Resolve it with `grund` and
...
+- [TESTS](docs/functional-spec/tests.md): The tests/ suite: tests that justify the metadata
```
vjovanov: This as well FS-tests?

kimeta: Yes I agree for all above, those should not have their specific kinds.

AGENTS.md
kimeta: Hmm, good question. I think it depends. These concrete skills are used by our fixing and reviewing agents in generation workflows. They cite the functional spec to express all the rules that agents need to respect for generation to succeed. Only skill that I think should be relevant for users is [fix-missing-reachability-metadata](https://github.com/oracle/graalvm-reachability-metadata/tree/master/skills/fix-missing-reachability-metadata), and it does not cite anything.

AGENTS.md
kimeta: AR and FS?

grund.toml
kimeta: CI is here a kind that is citable, when we turn all of it into the AR and FS, they will dissapear.

grund.toml
kimeta: Agreed. METADATA and TESTS is here part of FS.

kimeta: This is currently citable, that's why it is not complaining.

docs/architecture/README.md
kimeta: No, all of them will become AR.

vjovanov (approved):

PR #9537, opened 2026-08-25
[GenAI] Add support for org.apache.calcite:calcite-core:1.35.0 using gpt-5.6-terra (chunked dynamic-access)

graalvmbot:
## What does this PR do?

Refs: #446

This PR introduces tests and metadata for org.apache.calcite:calcite-core:1.35.0, enabling support for this library.

Summary:
- Chunked dynamic-access: yes
- Chunk class threshold: 15
- Current chunk class count: 15
- Processed dynamic-access classes: completed=16, skipped=0, exhausted=1, failed=0
- Strategy: dynamic_access_main_sources_pi_gpt-5.6-terra
- Agent: pi
- Model: gpt-5.6-terra
- Input tokens: 401753
- Cached input tokens: 4433920
- Output tokens: 33156
- Metadata entries: 579
- Test-only metadata entries: 23
- Iterations: 8
- Library coverage percentage: 3.5
- Generated lines of code: 125
- Tested library lines of code: 4470

### Forge

- Forge monitored branch: `master`
- Forge branch: `master`
- Forge commit hash: `8639c823fbd9d6e8088ce9784ceea81507cdba92`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 22/121 covered calls (18.18%)
- Reflection: 21/115 covered calls (18.26%)
- Resources: 1/6 covered calls (16.67%)

Library coverage:
- Instruction: 30098/708772 (4.25%)
- Line: 4470/127745 (3.50%)
- Method: 886/21213 (4.18%)

### Local CI Verification

- Status: `success`
- Commands run: 2
- Fixup attempts: 0

Forge-Publication-ID: forge-446-20260825154038187388-dc43c96ee634

PR #9555, opened 2026-08-27
Exempt chunked PRs from coverage thresholds

### LINKED ISSUE #9554: Disable coverage thresholds for chunked PRs
## Problem

Chunked dynamic-access PRs change a bounded subset of classes, but the review skills apply label-wide coverage percentage gates to library-wide statistics. This makes library size and chunk position affect the verdict even when the chunk covers meaningful dynamic-access calls.

This produced a false failure while reviewing #9537: its 22 covered calls out of 121 library-wide calls fell below the new-library 20% floor.

## Expected behavior

When a PR has the `chunked-dynamic-access` label:

- Skip percentage-based dynamic-access thresholds and regression comparisons for every chunk, including the final chunk.
- Fail the dynamic-access coverage gate only when `dynamicAccess.coveredCalls` is `0`.
- Keep every non-coverage rule and metadata guardrail.

## Scope

Update the contribution contract and all five dynamic-access review skills so they apply the same exception.

kimeta:
Chunked dynamic-access PRs cover a bounded set of classes, while their reported statistics describe the whole library. Applying percentage-based coverage gates therefore makes library size and chunk position affect the review verdict. This raised a false coverage finding on #9537 even though that chunk covered 22 dynamic-access calls.

This change:

- Exempts PRs with the `chunked-dynamic-access` label, including the final chunk, from percentage thresholds and regression comparisons in [§FS-contribution-contract.3](https://github.com/oracle/graalvm-reachability-metadata/blob/codex/disable-chunk-coverage-threshold/docs/functional-spec/contribution-contract.md#3-coverage-and-metadata-gates).
- Applies the same override in all five dynamic-access review skills.
- Keeps the coverage gate blocking when `dynamicAccess.coveredCalls` is `0`.
- Leaves non-coverage rules and metadata guardrails unchanged.

Under this rule, #9537's `22/121` result no longer fails because the chunk has non-zero covered calls.

Validation:

- `grund check`
- `git diff --check origin/master..HEAD`

Fixes #9554

### DIFF
--- a/docs/functional-spec/contribution-contract.md
+++ b/docs/functional-spec/contribution-contract.md
@@ -80,6 +80,13 @@ credible explanation such as a changed upstream API surface.
 | `library-update-request` | Coverage percentage does not drop at all against the previously supported version. |
 | `fixes-javac-fail`, `fixes-java-run-fail`, `fixes-native-image-run-fail` | Coverage percentage does not drop by more than 20 percentage points against the previously tested version. |
 
+Forge-generated PRs whose publication descriptor reports
+`modifiers.chunked_dynamic_access: true` are exempt from every percentage gate
+in this table, including the final chunk. A chunk covers only part of the
+library-wide dynamic-access report (§forge/FS-forge-chunked-dynamic-access), so
+its coverage gate fails only when the reported `dynamicAccess.coveredCalls` is
+zero. This exception does not relax any non-coverage rule or guardrail below.
+
 Two guardrails accompany the gates:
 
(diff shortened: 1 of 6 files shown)


vjovanov (approved):

vjovanov:
Can we summon an agent to report on why we miss dynamic access when the coverage is low?

kimeta:
> Can we summon an agent to report on why we miss dynamic access when the coverage is low?

Yes, I will merge this because chunked PRs are currently wrongly reviewed. I will open PR for detailed agent analysis.

PR #9562, opened 2026-08-27
Chunk optimistic dynamic-access runs after gates

### LINKED ISSUE #9560: Forge: chunk optimistic dynamic-access runs after the native-test gate
Chunking is switched off for every bulk strategy. `prepare_dynamic_access_chunking` (`forge/forge_metadata.py:5064`) returns before it does any work when the selected strategy's workflow is `optimistic_dynamic_access` (`forge/forge_metadata.py:5080`):

```python
if strategy.get("workflow") == "optimistic_dynamic_access":
    # Pure bulk runs have no per-class checkpoints or budgets.
    log_stage("dynamic-access-chunking",
              f"Chunking disabled for bulk strategy '{strategy_name}'; the bulk engine owns the full report.")
    return None
```

That bail sits ahead of the report refresh and ahead of `_resolve_dynamic_access_exhaust_report`, so a bulk run never reads or writes the coordinate's exhaust report. Five strategies use this workflow, and both issue queues currently default to one of them, so in practice chunking no longer runs for new-library or library-update work at all.

The effect is visible on #5950 (`org.bouncycastle:bc-fips:2.1.2`). The issue carries `chunked-dynamic-access` and has a committed `tests/src/org.bouncycastle/bc-fips/2.1.2/dynamic-access-exhaust-report.json` recording `classThreshold: 15`, 16 `completedClasses` and 14 `exhaustedClasses` — 30 classes already processed by earlier chunked runs. Today's bulk run opened at 35/65 covered with 30 uncovered calls across 24 classes, worked the full report including the 14 classes previous runs had already exhausted, and wrote nothing back. Coverage accounting still comes out right, because a covered class simply drops out of the refreshed report; the chunk bookkeeping is what goes stale.

## The shape

Keep bulk as the engine and put the chunk boundary after it, rather than deciding chunking before the workflow starts.

1. Bulk runs its iteration budget (`max-optimistic-iterations`, currently `3`). The native-test gate keeps firing after every successful iteration, exactly as it does today (`forge/ai_workflows/core/optimistic_dynamic_access_strategy.py:170`). The chunk decision is taken once, after the budget is spent.
2. The count is read from the last successful iteration's report. That report was refreshed immediately before the gate that passed, and an iteration that never reaches the native test resets the tree back to that same checkpoint, so the number is exact and no additional `generateDynamicAccessCoverageReport` is needed.
3. Classes uncovered in the run's baseline report and covered in the final one are recorded as `completedClasses` in the coordinate's exhaust report. Nothing is recorded as exhausted — a bulk pass cannot attribute a failure to a class.
4. `remaining` is the uncovered class count minus the processed set (`completed ∪ skipped ∪ exhausted ∪ failed`) taken from the exhaust report and the continuation marker, which is what `_remaining_uncovered_dynamic_access_classes` (`forge/forge_metadata.py:4955`) already computes. Subtracting the processed set is what lets a coordinate carrying earlier iterative chunk state converge: bc-fips's 14 exhausted classes stop being recounted on every run.
5. The decision, taken once:

| Condition | Result |
| --- | --- |
| Bulk covered no new class | Final chunk. `Fixes: #<issue>`, exhaust report removed from the PR, issue closes |
| `remaining` ≤ `FORGE_DYNAMIC_ACCESS_CHUNK_CLASS_THRESHOLD` (default `15`) | Final chunk, as above |
| `remaining` > threshold | `chunk_ready`. `Refs: #<issue>`, exhaust report committed, `chunked-dynamic-access` reapplied, issue returns to `Todo` |

Publication needs no new plumbing. `--chunked-dynamic-access` is already keyed off the exhaust report file's presence and `--chunk-final` off the status not being `chunk_ready` (`forge/forge_metadata.py:7102`), and `apply_chunked_dynamic_access_completion_follow_up` already reapplies the label on `chunk_ready`.

The dispatcher's bail stays. With the decision downstream, a bulk phase needs no chunk size and the pre-run report refresh would be dead cost. What changes is its reason: the dispatcher does not decide chunking for bulk because the workflow decides it after the gate, not because bulk cannot be chunked.

## Zero yield ends chunking

A bulk pass hands the agent the entire report. If three iterations and a passing gate produce not one newly covered class, the next chunk would run the same prompt against the same report with the same budget, and nothing would make it go differently. `_remaining_uncovered_dynamic_access_classes` subtracts `completed ∪ skipped ∪ exhausted ∪ failed`, and a bulk chunk populates only `completedClasses`, so the skip set is far weaker than in the iterative path and a repeat chunk would work the same classes again.

So a bulk phase that covered no new class publishes what it has as the final chunk rather than opening an unbounded series of empty `Refs:` PRs. This is the cheap form of the pass-yield collapse rule from #9499.

## The composite gets the same boundary

`increase_dynamic_access_coverage` with `primary-workflow: optimistic_dynamic_access` — the optimistic-then-iterative family — takes the boundary at the same place: after the bulk phase and its gate, before class-by-class exploration starts. If `remaining` exceeds the threshold, the run publishes the chunk PR and never enters iterative exploration.

The composite already receives `--chunk-class-count` today, because the bail at `forge_metadata.py:5080` matches on `workflow` and the composite's workflow is `increase_dynamic_access_coverage`. So it already chunks — but only in its iterative half, after the bulk phase has already run unbounded over the full report. The boundary moves ahead of that.

A zero-yield bulk phase in the composite does not publish a chunk and does not end the sequence; it falls through into iterative exploration as it does today. The iterative engine marks every class it abandons as exhausted (`dynamic_access_iterative_strategy.py:587`, `:643`), so a class leaves the remaining set whether it was covered or given up on, and progress is structural. The zero-yield rule exists for pure bulk precisely because pure bulk has no such record.

## Defaults move to the composite

Both queues default to a pure-bulk strategy today, which is why the disabled chunk path is the path every issue takes:

| Constant | Today | After |
| --- | --- | --- |
| `add_new_library_support.py:94` `DEFAULT_STRATEGY_NAME` | `dynamic_access_bulk_pi_gpt-5.6-sol` | `optimistic_dynamic_access_iterative_pi_gpt-5.6-sol` |
| `improve_library_coverage.py:95` `DEFAULT_STRATEGY_NAME` | `library_update_dynamic_access_bulk_pi_gpt-5.6-sol` | `library_update_optimistic_pi_gpt-5.6-sol` |
| `forge_metadata.py:385` `DEFAULT_WORK_QUEUE_STRATEGY_NAME` | `dynamic_access_bulk_pi_gpt-5.6-sol` | `optimistic_dynamic_access_iterative_pi_gpt-5.6-sol` |

With the boundary in place, the composite spends bulk first on the cheap broad pass, publishes chunks while the remainder is large, and only reaches per-class iterative exploration once the remainder has fallen to the threshold — which is the part iterative is good at and the part bulk keeps re-running.

## Spec impact

This does not fit the current spec, so the spec moves first, in its own commit, ahead of any code.

§FS-forge-chunked-dynamic-access requires `forge_metadata.py` to run or refresh the report **before choosing the execution mode**, and to pass a concrete chunk size rather than a generic workflow policy. It gains the bulk case: for a bulk phase the mode is decided by the workflow after the native-test gate, and no chunk size is passed, because a bulk phase works the whole report regardless of chunk size.

§AR-dynamic-access-bulk states that a bulk attempt "does not commit per class and does not record exhausted classes — a bulk attempt is accepted or reset whole". The no-per-class-commit clause and the whole-attempt accept-or-reset clause both stay true. It gains: at a chunk boundary a bulk attempt records the classes it finished, derived by diffing the report against the run's baseline, and never records a class as exhausted.

§AR-chunked-dynamic-access-pr-linking requires every chunk PR to carry the merged exhaust state the next run needs to skip classes already "completed, skipped, exhausted, or failed". It changes from enumerating all four states to requiring the state the producing engine can attribute: iterative carries all four, a bulk chunk carries completed.

Moving only one of the two AR points does not work. Making bulk record all four states would have it assert what it cannot know, and moving only the linking point leaves bulk recording nothing, which is the defect this issue is about.

§AR-dynamic-access-composite gains the boundary between its two phases.

Blast radius, from `grund refs`: `FS-forge-chunked-dynamic-access` 6 citations across 4 files, `AR-dynamic-access-bulk` 7 across 6, `AR-chunked-dynamic-access-pr-linking` 11 across 6.

## Settled

**Where the gate runs.** The gate stays inside the iteration loop, firing after every successful iteration. The chunk decision is taken once, after the whole budget is spent, from the report the last passing gate ran against.

**Which count is compared against the threshold.** Remaining uncovered classes after subtracting the processed set recorded by earlier chunks, not the raw uncovered class count.

**What guarantees progress across chunks.** For pure bulk, the zero-yield rule: a chunk that covers no new class is the final chunk. For the composite, the iterative half's exhaustion record, which already guarantees the remaining set shrinks on every chunk.

kimeta:
Closes #9560

## Put the boundary after optimistic work

Optimistic dynamic-access strategies previously bypassed dispatcher chunking, so the default new-library and library-update queues could repeatedly work the full report without updating the coordinate's exhaust state.

The chunk decision now happens once after every configured optimistic iteration and its native-test gate. The bulk baseline is compared with the last successfully gated report, and classes proven covered are recorded before the composite decides whether iterative exploration is still needed. This makes the configured boundary an invocation-wide quota across both phases, as specified by [§FS-forge-chunked-dynamic-access](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/optimistic-chunk-quota/forge/docs/functional-spec/functional-spec.md).

## Fill only the post-bulk shortfall

With the default boundary of 15:

| Bulk result | Remaining | Outcome |
| ---: | ---: | --- |
| 20 completed | 80 | Publish the bulk chunk; do not enter iterative |
| 10 completed | 90 | Let iterative process 5 terminal classes |
| 20 completed | 10 | Let iterative finish the final 10 classes |
| 0 completed | 90 | Give iterative the full 15-class quota |

The implementation carries the already-gated report into iterative exploration instead of regenerating it. Existing completed, skipped, exhausted, and failed classes are excluded from the remainder, while bulk itself records only classes it can prove completed. Pure-bulk runs keep their separate zero-yield/final-remainder behavior. These responsibilities are captured by [§AR-dynamic-access-bulk](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/optimistic-chunk-quota/forge/docs/architecture/workflows.md), [§AR-dynamic-access-composite](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/optimistic-chunk-quota/forge/docs/architecture/workflows.md), and [§AR-chunked-dynamic-access-pr-linking](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/optimistic-chunk-quota/forge/docs/architecture/workflows.md).

## Route queue defaults through both phases

The new-library, library-update, and generic work-queue defaults now select their optimistic-then-iterative strategies. This keeps the cheap broad pass first while ensuring iterative work fills a shortfall or finishes the last small remainder.

### DIFF
--- a/forge/DEVELOPING.md
+++ b/forge/DEVELOPING.md
@@ -123,7 +123,7 @@ python3 ai_workflows/drivers/add_new_library_support.py \
 ```
 
 Options:
-- `--strategy-name NAME` select a predefined workflow strategy from `strategies/predefined_strategies.json`. Defaults to `dynamic_access_bulk_pi_gpt-5.6-sol` (Pi agent).
+- `--strategy-name NAME` select a predefined workflow strategy from `strategies/predefined_strategies.json`. Defaults to `optimistic_dynamic_access_iterative_pi_gpt-5.6-sol` (Pi agent).
 - `--keep-tests-without-dynamic-access` keeps generated tests for dynamic-access workflows even if no dynamic-access call sites are covered.
 - `-v`, `--verbose` enable verbose agent output.
 
--- a/forge/ai_workflows/core/increase_dynamic_access_coverage_strategy.py
+++ b/forge/ai_workflows/core/increase_dynamic_access_coverage_strategy.py
@@ -53,6 +53,7 @@ class IncreaseDynamicAccessCoverageStrategy(WorkflowStrategy):
     def run(self, agent, **kwargs):
         current_report: DynamicAccessCoverageReport | None = None
         iterative_chunk_count: int | None = None
+        required_iterative_chunk_phase: bool = False
         skip_dynamic_access_phase: bool = False
 
         if self.primary is None:
@@ -92,6 +93,7 @@ class IncreaseDynamicAccessCoverageStrategy(WorkflowStrategy):
                         "classes={budget}".format(budget=iterative_chunk_count)
                     )
                     skip_dynamic_access_phase = iterative_chunk_count == 0
+                    required_iterative_chunk_phase = iterative_chunk_count > 0
 
         library = self.context.get("library") or self.context.get("updated_library")
         da_context = dict(self.context)
@@ -142,7 +144,12 @@ class IncreaseDynamicAccessCoverageStrategy(WorkflowStrategy):
 
         has_issue_requested_metadata = da.has_issue_requested_metadata_context()
         if not phase_ok:
-            if self.primary is None and not has_issue_requested_metadata:
+            if required_iterative_chunk_phase:
+                self._print_message(
+                    "required iterative dynamic-access chunk phase did not succeed"
+                )
+                status = RUN_STATUS_FAILURE
+            elif self.primary is None and not has_issue_requested_metadata:
                 self._print_message(
                     "dynamic-access coverage phase did not succeed and no reporter-requested metadata phase is available"
                 )
@@ -180,7 +187,7 @@ class IncreaseDynamicAccessCoverageStrategy(WorkflowStrategy):
                 if len(result) == 2:
                     return status, iterations
                 return (status, iterations) + result[2:]
-            if status != RUN_STATUS_CHUNK_READY:
+            if status not in {RUN_STATUS_CHUNK_READY, RUN_STATUS_FAILURE}:
                 status = RUN_STATUS_SUCCESS
 
         if self.primary is None:
--- a/forge/ai_workflows/core/increase_dynamic_access_coverage_strategy.py
+++ b/forge/ai_workflows/core/increase_dynamic_access_coverage_strategy.py
@@ -11,6 +11,7 @@ from ai_workflows.core.workflow_strategy import (
     WorkflowStrategy,
 )
 from utility_scripts.continuation_marker import PHASE_FIX, save_phase_update
+from utility_scripts.dynamic_access_report import BulkDynamicAccessProgress, DynamicAccessCoverageReport
 from utility_scripts.java_fix_coverage_follow_up import uncovered_dynamic_access_class_count
 
 
@@ -26,7 +27,8 @@ class IncreaseDynamicAccessCoverageStrategy(WorkflowStrategy):
 
     def __init__(self, strategy_obj: dict, **context):
         super().__init__(strategy_obj, **context)
-        primary_workflow_name = strategy_obj.get("primary-workflow")
+        self.primary_workflow_name: str | None = strategy_obj.get("primary-workflow")
+        self.chunk_class_count: int = int(self.context.get("chunk_class_count") or 0)
         self.reachability_repo_path = self.context["reachability_repo_path"]
         self.library = self.context.get("library") or self.context.get("updated_library")
         self.group, self.artifact, self.version = self.library.split(":")
@@ -35,9 +37,12 @@ class IncreaseDynamicAccessCoverageStrategy(WorkflowStrategy):
         )
         if self.dynamic_access_class_threshold < 0:
             raise ValueError("dynamic_access_class_threshold must be non-negative")
-        if primary_workflow_name:
-            PrimaryClass = WorkflowStrategy.get_class(primary_workflow_name)
-            self.primary = PrimaryClass(strategy_obj, **context)
+        if self.primary_workflow_name:
+            PrimaryClass = WorkflowStrategy.get_class(self.primary_workflow_name)
+            primary_context: dict = dict(context)
+            if self.primary_workflow_name == "optimistic_dynamic_access":
+                primary_context["defer_dynamic_access_chunk_decision"] = True
+            self.primary = PrimaryClass(strategy_obj, **primary_context)
         else:
             self.primary = None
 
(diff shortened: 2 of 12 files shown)


jvukicev (changes requested):
The resumed optimistic path resets an exhausted chunk budget instead of preserving the invocation-wide quota.

In `prepare_dynamic_access_chunking`, `active_chunk_remaining_budget` is applied only when it is greater than zero. When continuation state records `chunkClassCount=5` and `chunkProcessedClassCount=5`, `_continuation_active_chunk_remaining_budget()` correctly returns `0`, but this branch ignores it and returns the default boundary (`15`). I reproduced the result as:

```
remaining_budget=0
returned_boundary=15
```

A retry after consuming its saved budget can therefore perform another bulk/iterative quota, contrary to the invocation-wide boundary specified by this PR. Please handle the zero-budget continuation explicitly and add a regression case for the fully consumed (`5/5`) state.

kimeta:
Addressed in 910995769f. `prepare_dynamic_access_chunking` now applies the continuation budget whenever it is present, including `0`, so a fully consumed `5/5` chunk no longer falls back to the default boundary of `15`. I also added the requested regression case asserting a returned boundary of `0` for the fully consumed state. The focused test module passes all 8 tests (`python3 -m unittest tests.test_workflow_driver_defaults`). Thanks, @jormundur00.

jvukicev (approved):

PR #9566, opened 2026-08-27
Fix ni-run metrics call and spec its test patch

### LINKED ISSUE #9565: Successful native-image-run repairs are discarded by a metrics TypeError
Every native-image-run repair that fully succeeds is discarded.

`fix_ni_run.py:514` calls the shared metrics writer with a `tests_root` argument it does not accept:

```python
run_metrics = metrics_writer.create_java_run_fix_run_metrics_output_json(
    ...
    global_iterations=iterations,
    tests_root=tests_root,        # create_javac_fix_run_metrics_output_json has no such parameter
    strategy_name=args.strategy_name,
)
```

`create_javac_fix_run_metrics_output_json` (`metrics_writer.py:610`, aliased to the java-run name at `:683`) takes `repo_path, package, artifact, previous_library_version, new_library_version, agent, model_name, global_iterations, strategy_name, status` plus four keyword-only extras. `tests_root` is not among them, so the call raises `TypeError`.

The argument was never needed. `collect_base_metrics` derives `tests_root` itself at `metrics_writer.py:370` to count generated LOC, and `java_fail_workflow.py` passes `tests_root` only to `init_agent` (`:578`), never to the metrics writer (`:639`). Only `fix_ni_run` threads it into the metrics call, which is why the javac and java-run queues publish normally and only native-image-run dies.

## Why it costs a whole run

The call sits inside `if succeeded:`, after `finalize_run` has returned success. Reaching it requires a complete repair: seed, native gate, `checkMetadataFiles` recovery, finalization. Every failing run takes the `else:` branch, so the first run that fully succeeds is the one that crashes.

Observed on issue #9255 (`org.junit.platform:junit-platform-suite-api:6.1.0`):

- `fixTestNativeImageRun` produced the seed
- native-test gate `PASSED after 5 cycles`
- `checkMetadataFiles` failed on 67 out-of-scope `TypeReached` entries, two mechanical `allowed-packages` attempts found nothing new, the analysis agent pruned them, and the recheck passed
- `finalize_run` returned success
- `TypeError: create_javac_fix_run_metrics_output_json() got an unexpected keyword argument 'tests_root'`

The verified result — new `6.1.0/reachability-metadata.json`, the index entry, and the test-side metadata — was preserved on `ai/kimeta/human-intervention/issue-9255-fails-native-image-run-org.junit.platform-junit-platform-suite-api-6.1.0-c06e5a66`, the issue was reverted to `Todo`, and no PR was opened.

## Why the tests did not catch it

`tests/test_fix_ni_run.py:93` patches the function with a bare `Mock`:

```python
stack.enter_context(patch.object(
    fix_ni_run.metrics_writer,
    "create_java_run_fix_run_metrics_output_json",
    return_value={},
))
```

Without `autospec`, a `Mock` accepts any keyword argument, so three tests execute this exact line and pass while the real call raises. With `autospec=True` and the argument still present, those three fail with the production error.

## Fix

Drop the argument from the call site, and set `autospec=True` on the patch so the same signature drift fails a test instead of a run.

kimeta:
`fix_ni_run` passed `tests_root` to `create_javac_fix_run_metrics_output_json`, which has no such parameter. The call sits inside `if succeeded:`, after `finalize_run` returns, so it only fires once a native-image-run repair has fully passed — seed, native gate, `checkMetadataFiles` recovery, finalization — and then discards it with a `TypeError`.

Issue #9255 hit this: gate `PASSED after 5 cycles`, `checkMetadataFiles` repaired and passing, finalization successful, then the crash. The verified metadata was parked on a human-intervention branch, the issue went back to `Todo`, and no PR opened.

The argument was never needed. `collect_base_metrics` derives `tests_root` itself (`metrics_writer.py:370`) to count generated LOC, and `java_fail_workflow` passes it only to `init_agent` (`:578`), never to the metrics writer (`:639`). Removing it makes all three fix drivers agree, which is why javac and java-run were unaffected.

`tests/test_fix_ni_run.py` already exercised this line, but patched the function with a bare `Mock` that accepts any keyword argument. `autospec=True` binds the real signature: with the argument still present, three tests fail with the production error; with it removed, the file is green.

Verified locally: `test_fix_ni_run` passes, and the full suite is 717 tests with one pre-existing unrelated failure (`test_validate_no_scaffold_placeholders_rejects_placeholder_text`, which also fails on a clean tree).

Fixes #9565

### DIFF
--- a/forge/ai_workflows/drivers/fix_ni_run.py
+++ b/forge/ai_workflows/drivers/fix_ni_run.py
@@ -520,7 +520,6 @@ def main(argv=None) -> int:
             agent=agent,
             model_name=model_name,
             global_iterations=iterations,
-            tests_root=tests_root,
             strategy_name=args.strategy_name,
             status=finalize_status,
             starting_commit=checkpoint,
--- a/forge/tests/test_fix_ni_run.py
+++ b/forge/tests/test_fix_ni_run.py
@@ -93,6 +93,7 @@ class NativeImageRunDriverTests(unittest.TestCase):
             stack.enter_context(patch.object(
                 fix_ni_run.metrics_writer,
                 "create_java_run_fix_run_metrics_output_json",
+                autospec=True,
                 return_value={},
             ))
             stack.enter_context(patch.object(fix_ni_run.metrics_writer, "write_workflow_run_metrics"))


jvukicev (approved):
The metrics writer call now matches its real signature, and  protects the call site against future signature drift. Focused tests and the current CI checks pass.

PR #9570, opened 2026-08-27
Check artifact repositories before Forge claims issues

### LINKED ISSUE #9569: Check artifact repositories before Forge claims issues
Forge currently validates a TCP connection to `repo.maven.apache.org` at startup, while issue-form validation sends HTTP `HEAD` requests to `repo1.maven.org` and `packages.confluent.io`. A machine can therefore pass host requirements and then repeatedly claim and release issues when the actual artifact repository path is unavailable.

Require issue-work startup to:

- send `HEAD` to every configured base URL in `ARTIFACT_REPOSITORY_URLS`;
- require HTTP 200 from each repository before queue processing starts;
- report the exact URL and HTTP or transport result with an actionable fix;
- keep the existing generic TCP/proxy checks for non-artifact services.

This makes the startup gate prove the same DNS, proxy, TLS, and HTTP path used by issue-form artifact lookup, as required by [§forge/FS-forge-host-requirements](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/functional-spec/functional-spec.md#fs-forge-host-requirements).

kimeta:
Closes #9569

## What this changes

Forge previously proved only that a TCP connection to `repo.maven.apache.org:443` could be opened at startup. Issue-form validation actually uses HTTP `HEAD` against `repo1.maven.org` and `packages.confluent.io`, so a different hostname or an HTTP-level proxy, TLS, rate-limit, or server failure could pass startup and then make Forge claim and release issue after issue.

Issue-work startup now checks every base URL in the shared `ARTIFACT_REPOSITORY_URLS` tuple and requires an exact HTTP 200 before Forge scans or claims queue items. This makes the host gate exercise the same route and request method as artifact lookup, following [§forge/FS-forge-host-requirements](https://github.com/oracle/graalvm-reachability-metadata/blob/codex/forge-repository-http-health/forge/docs/functional-spec/functional-spec.md#fs-forge-host-requirements) and the claim pipeline in [§forge/AR-forge-workflow-pipeline](https://github.com/oracle/graalvm-reachability-metadata/blob/codex/forge-repository-http-health/forge/docs/architecture/architecture.md#ar-forge-workflow-pipeline).

## How repository health is decided

- Normalize each configured repository base to a trailing-slash URL and send `HEAD`.
- Accept only HTTP 200; report other statuses and DNS, proxy, TLS, timeout, or connection failures directly in the host-requirements manifest.
- Block issue work with one concrete remediation when any configured repository is unhealthy.
- Report these checks as not required for modes that do not process issues, while retaining the existing generic network probes for other services.

### DIFF
--- a/forge/docs/architecture/architecture.md
+++ b/forge/docs/architecture/architecture.md
@@ -257,8 +257,10 @@ one manifest:
   `local_repositories/`, and temporary Gradle state.
 - **Network reachability** on 443, through the configured proxy when one is
   set: `github.com`, `api.github.com`, `chatgpt.com`, `services.gradle.org`,
-  `repo.maven.apache.org`, `plugins.gradle.org`, `registry-1.docker.io`,
-  `auth.docker.io`.
+  `plugins.gradle.org`, `registry-1.docker.io`, `auth.docker.io`. Each base URL
+  in `ARTIFACT_REPOSITORY_URLS` is checked separately with HTTP `HEAD` and must
+  return `200`; this currently covers `repo1.maven.org` and
+  `packages.confluent.io`, the exact hosts used by `check_issue_form()`.
 - **Git transport** to the monitored branch from each checkout, and access to
   the Docker daemon.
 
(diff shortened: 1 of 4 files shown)


jvukicev (approved):
The host gate now probes the exact configured artifact repository routes used by issue validation. Unit tests pass, both live repository probes returned HTTP 200, and CI is green.

PR #9561, opened 2026-08-27
[GenAI] Improve coverage for org.bouncycastle:bc-fips:2.1.2 using gpt-5.6-sol

graalvmbot:
## What does this PR do?

Fixes: #5950

This PR improves dynamic-access coverage for org.bouncycastle:bc-fips:2.1.2 by generating additional tests.

Summary:
- Chunked dynamic-access: yes
- Validation commands: `./gradlew test -Pcoordinates=org.bouncycastle:bc-fips:2.1.2`
- Validation result: `success`
- Requested coordinate: `org.bouncycastle:bc-fips:2.1.2`
- Match type: `tested-version`
- Matched metadata version: `2.1.2`
- Matched test version: `2.1.2`
- Resolved metadata version: `2.1.2`
- Resolved test version: `2.1.2`
- Strategy: library_update_dynamic_access_bulk_pi_gpt-5.6-sol
- Agent: pi
- Model: gpt-5.6-sol
- Input tokens: 0
- Cached input tokens: 0
- Output tokens: 0
- Metadata entries (before): 881
- Metadata entries (after): 895
- Test-only metadata entries (before): 70
- Test-only metadata entries (after): 70
- Iterations: 0
- Library coverage percentage: 33.04
- Generated lines of code: 1728
- Tested library lines of code: 0

### Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `b13b34a3f812ad0b66f68f824ef7de1097baed7f`

#### Dynamic access coverage

- Before (org.bouncycastle:bc-fips:2.1.2): 35/65 covered calls (53.85%)
- After (org.bouncycastle:bc-fips:2.1.2): 39/65 covered calls (60.00%)

**Reflection:**
- Before (org.bouncycastle:bc-fips:2.1.2): 32/60 covered calls (53.33%)
- After (org.bouncycastle:bc-fips:2.1.2): 36/60 covered calls (60.00%)

**Resources:**
- Before (org.bouncycastle:bc-fips:2.1.2): 3/5 covered calls (60.00%)
- After (org.bouncycastle:bc-fips:2.1.2): 3/5 covered calls (60.00%)

#### Library coverage

**Instruction:**
- Before (org.bouncycastle:bc-fips:2.1.2): 169189/539526 (31.36%)
- After (org.bouncycastle:bc-fips:2.1.2): 178272/539526 (33.04%)

**Line:**
- Before (org.bouncycastle:bc-fips:2.1.2): N/A
[...]

PR #9572, opened 2026-08-28
Restore isolated agent PR reviews

### LINKED ISSUE #9573: Prevent blind Forge PR approvals
Forge's automated PR reviewer has two related failure modes:

1. **A review can be approved without inspecting the change.** On [PR #9571](https://github.com/oracle/graalvm-reachability-metadata/pull/9571#pullrequestreview-5046660291), enterprise Codex policy substituted a restricted sandbox. The agent said it could not inspect the authoritative snapshot, but Forge still submitted the structured `APPROVE` result because the decision and body were non-empty.
2. **Some diffs cannot become review snapshots.** [PR #9561](https://github.com/oracle/graalvm-reachability-metadata/pull/9561) contained ANSI-colored logs, causing `gh pr diff --patch` to fail. Allowing escape sequences alone would materialize an uncapped patch of more than 100 MB, which is not usable agent context.

The reviewer must follow [§forge/FS-automated-pr-review](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/functional-spec/functional-spec.md#fs-automated-pr-review-automated-pull-request-review):

- run in an isolated review worktree with working access to authenticated PR metadata and checks;
- inspect the label-specific rules and targeted local diffs;
- fail without submitting a GitHub review when inspection, execution, or output validation fails;
- avoid materializing an uncapped full patch as agent context.

kimeta:
Closes #9573

## What this does

Forge's snapshot reviewer could submit a live approval even when the configured Codex process could not read `.forge-review-context.json`; ANSI-bearing or very large diffs could also make snapshot construction fail or produce an unusable payload. After reproducing both paths, we chose the simpler fix: run Pi through Forge's existing analysis-agent runtime inside the detached review worktree instead of adding consumption proofs and patch-budget logic to the snapshot protocol.

That keeps automated review aligned with [§forge/FS-automated-pr-review](https://github.com/oracle/graalvm-reachability-metadata/blob/restore-isolated-pi-pr-reviews/forge/docs/functional-spec/functional-spec.md#fs-automated-pr-review-automated-pull-request-review): inspect the published PR, its checks, and its diff under the label-specific rules before Forge submits a decision.

## How review decisions are produced

- Forge invokes `analysis_agent_run(...)` with a review-specific environment that selects Pi, the configured OpenAI Codex model and provider, and the review thinking level. The shared adapter owns invocation, logs, results, and token accounting.
- The review turn explicitly retains the authenticated, non-interactive `gh` session; ordinary analysis turns continue to have GitHub credential channels removed.
- Pi reads PR metadata and checks through `gh`, then uses targeted local diffs against the fresh `origin/master` checkout. The prompt forbids writes, re-checkout, full-patch output, and GitHub mutation.
- Pi must return either structured `APPROVE`/`REQUEST_CHANGES` output or an `error` object when inspection is unavailable. Forge rejects errors, malformed output, timeouts, failed agent turns, and worktree changes; Forge remains the only process that submits the review.
- The uncapped `gh pr diff --patch` snapshot is removed, so terminal escapes and oversized generated logs no longer become reviewer input.

A live `library-update-request` run against [#9561](https://github.com/oracle/graalvm-reachability-metadata/pull/9561) exercised the isolated review path end to end: Pi inspected the change, Forge submitted the approval, and the normal merge gates completed.


jvukicev (changes requested):
Two trust-boundary issues still allow the blind or unauthorized review behavior this PR is intended to prevent.

1. The Pi review process receives the operator's write-capable GitHub session. `review_pull_request` sets `_FORGE_AGENT_ALLOW_GITHUB_ACCESS=1`, and `agent_process_environment` consequently retains both token variables and `GH_CONFIG_DIR`. The available session has `repo` scope, so Pi can directly submit reviews, comments, merges, or other mutations. The prompt's prohibition is not an enforcement boundary, and the worktree-status comparison cannot detect remote mutations. This is especially unsafe when the agent is running inside and reading an untrusted PR checkout. Please provide a genuinely read-only credential or a constrained read-only GitHub interface rather than exposing the publisher session.

2. `parse_review_decision` does not enforce the documented mutually exclusive result schemas or string types. I reproduced both of these being accepted as approvals:

```
{"error":"Unable to inspect diff","decision":"APPROVE","body":"No blockers."}
{"decision":"APPROVE","body":{"error":"Unable to inspect diff"}}
```

The first preserves a direct form of the blind-approval bug, and the second turns malformed output into a non-empty review body through `str(...)`. Please require either an exact `{ "error": <string> }` failure object or an exact `{ "decision": <string>, "body": <string> }` decision object, reject extra/error fields on decisions, and safely reject non-object JSON. Add regression cases covering mixed error/approval output and non-string fields.

kimeta:
@jormundur00 The previous review agent already ran as trusted automation with the operator GitHub privileges and submitted reviews directly, and we did not have issues with that model. This refactor is not intended to introduce a new trust boundary. There is little value in restricting the agent from using gh and other network-dependent tools during review if Forge later grants equivalent access merely to publish the result.

I have therefore reverted the offline context/JSON decision approach and restored the old review behavior. The agent again inspects the live PR and submits the review itself. The only intended architectural difference is that it is now invoked through the centralized analysis_agent API, which owns agent, provider, model, and thinking-level selection as well as logging, timeouts, errors, and token accounting. Since Forge no longer parses and resubmits a JSON verdict, the result-schema issue from point 2 no longer applies.

jvukicev (changes requested):
There is still one blocking failure-path issue in the direct-review implementation.

`review_pull_request()` treats every zero-exit agent turn as a successful review without verifying that the agent actually submitted a fresh GitHub review (`forge/forge_metadata.py:2939-2961`). I reproduced this with a normal agent result whose response was: “I could not inspect the PR metadata, so I did not submit a review.” The function still returned `True`.

The subsequent reconciliation reads the PR’s aggregate `reviewDecision`, so it can act on a stale approval or stale requested-changes review rather than a verdict produced by this invocation. In the stale-approval case, this can bypass the inspection failure and proceed toward merge.

Please verify that the invocation created a new `APPROVED` or `CHANGES_REQUESTED` review for the expected PR and head commit before returning success or running post-review reconciliation, and add a regression test for a successful agent process that reports it could not inspect and submitted no review.

The previous two findings are otherwise addressed: the trusted GitHub-access model is now explicit, and the structured-result parser has been removed.

kimeta:
> I reproduced this with a normal agent result whose response was: “I could not inspect the PR metadata, so I did not submit a review.” The function still returned `True`.

Which agent and model did you use?

jvukicev (approved):

PR #9584, opened 2026-08-28
Refine repair PR coverage gates

### LINKED ISSUE #9583: Refine repair PR coverage regression gates
Repair PR reviews currently use two noisy proxies for regressions:

- metadata-entry count drops, although generation and test-only classification can legitimately produce smaller metadata while Native Image and dynamic-access coverage remain green;
- percentage-point drops on very small dynamic-access reports, where one or two calls create large percentage swings.

Update the contribution contract and the three `fixes-*` review skills so that:

- metadata-entry counts are telemetry only and never block repair PRs;
- the repair gate is applied independently to overall coverage and every breakdown;
- zero covered calls remain blocking when calls exist;
- equal or increased covered calls, or an equal or increased coverage ratio, pass;
- when both reports contain fewer than 10 calls, a loss of at most two covered calls passes;
- otherwise, a drop greater than 20 percentage points remains blocking;
- every reported regression includes an evidence-backed explanation of why it happened, or explicitly states that the cause could not be established without guessing.

Keep Native Image execution, meaningful-test, scope, and anti-bypass requirements unchanged.

kimeta:
Repair PR reviews currently treat metadata-count shrinkage and percentage swings in tiny dynamic-access reports as regressions. Neither signal necessarily means the repaired version lost meaningful coverage: metadata generation can become smaller, and one or two calls can move a small report by tens of percentage points.

This change makes the repair gate reflect stronger evidence while retaining the Native Image and meaningful-test requirements.

### Changes

- Remove metadata-entry count changes as a blocking signal for `fixes-*` PRs.
- Apply one ordered gate to overall dynamic access and every breakdown.
- Accept equal or increased covered-call counts or coverage ratios.
- Allow small reports with fewer than 10 calls on both versions to lose at most two covered calls.
- Retain the 20-percentage-point limit for larger reports and the zero-covered-call failure.
- Require every reported regression to include an evidence-backed cause analysis, or explicitly state that the cause is unknown.
- Update the `fixes-javac-fail`, `fixes-java-run-fail`, and `fixes-native-image-run-fail` review skills.

### Verification

- Validated all three skill packages with `quick_validate.py`.
- Ran `git diff --check`.

Fixes #9583


jvukicev (changes requested):
The ordered repair gate needs two consistency fixes before this is ready:

1. docs/functional-spec/contribution-contract.md:85 requires checking every breakdown present in either version, but the ordered steps assume both sides provide numeric fields. Breakdown sets genuinely vary across versions, and the generated comparison renders an absent side as N/A, so reviewers cannot deterministically apply the gate. Please define the missing-side semantics explicitly, for example by normalizing an absent breakdown to a no-call 0/0 scope, and mirror that rule in all three review skills.

2. docs/functional-spec/functional-spec.md:237 still describes a bounded drop for the fixes-* labels, while this gate permits an arbitrarily large percentage drop when covered-call counts remain equal or increase. The summaries in skills/README.md:35, :40, and :45 also describe the old behavior. Please update these references to describe or point to the ordered repair gate.

Local validation: all three changed skills pass quick_validate.py; git diff --check passes; GitHub CI is green.

jvukicev (approved):

PR #9652, opened 2026-08-31
Rename bulk workflow and align defaults

### LINKED ISSUE #9651: Rename bulk workflow and align optimistic defaults
## Problem

Forge uses `optimistic_dynamic_access` as the workflow identifier for the bulk-only full-report phase, even though the intended optimistic strategy is the composite that runs bulk first and class-by-class refinement afterward. This makes logs, configuration, prompt names, and documentation ambiguous.

The new-library driver defaults to the optimistic composite, but `do_up_to_date_work.sh` still overrides that default with the bulk-only strategy. Update requests already inherit their optimistic composite default, so the two request queues behave differently.

## Expected

- Reserve **bulk** for the full-report phase.
- Reserve **iterative** for class-by-class exploration.
- Reserve **optimistic** for the bulk-then-iterative composite.
- Rename the bulk workflow to `bulk_dynamic_access`, including its implementation, prompt key, iteration parameter, and log stage.
- Default both new-library and update-request work queues to their optimistic composite strategies.
- Preserve composite strategy names and behavior.

## Validation

Update focused workflow/default tests, validate all predefined strategy bundles and prompt paths, and keep the terminology aligned in the Forge specification and architecture.

kimeta:
The bulk-only workflow is currently named `optimistic_dynamic_access`, while the behavior we call optimistic is the composite that runs bulk first and class-by-class refinement afterward. The `do-work` wrapper also still overrides the new-library driver’s composite default with the bulk-only strategy, unlike update requests.

This change:

- reserves `bulk` for the full-report phase and `optimistic` for the bulk-to-iterative composite;
- renames the bulk workflow, implementation class/module, prompt keys/files, iteration parameter, and log stage;
- keeps the existing optimistic composite strategy names;
- aligns the new-library `do-work` default with the composite strategy already selected by the driver;
- updates the functional specification first, then architecture and regression tests.

The full Forge unit suite ran 772 tests with 771 passing. The remaining existing scaffold-placeholder validator failure reproduces independently of this diff.

Fixes #9651


jvukicev (approved):
Reviewed locally at 3f43e0da10085370356048f049601d24ceba38c7.

No blocking findings. The workflow, implementation class/module, prompt keys and files, iteration parameter, log stage, predefined strategies, documentation, and regression tests are renamed consistently, while the existing optimistic composite strategy names remain stable. The new-library do-work default now matches the driver default.

Validation:
- 66 focused Forge tests passed.
- Full Forge suite ran 772 tests: 771 passed; the sole scaffold-placeholder validator failure reproduces independently on origin/master.
- JSON parsing, Python compilation, and git diff --check passed.
- No stale internal bulk-workflow identifiers remain.
- All applicable GitHub checks are green.

PR #9844, opened 2026-09-04
[Automation] Update supported library versions 2026-09-04T00:30

graalvmbot:
This pull request updates supported versions of the existing libraries in the repo.

Added tested versions:
#### `org.jline:jline-console`
- `4.4.2`
#### `software.amazon.awssdk:s3`
- `2.54.12`


kimeta (approved):
Verified the bulk-update scope, body-to-diff version matches, green CI, and final `validateIndexFiles` against current `origin/master`. The two tested versions are appended to the correct metadata buckets; no URL or unrelated field changes.

Issue #9764, opened 2026-09-03
Update existing library: org.apache.kafka:kafka-streams:4.3.1


jessecoddington:
### Additional context and logs

Added locally to fix one of the bugs with the latest version

```
{
  "reflection": [
    {
      "type": "org.apache.kafka.streams.KafkaStreams",
      "fields": [
        {
          "name": "topologyMetadata"
        }
      ]
    },
    {
      "type": "org.apache.kafka.streams.processor.internals.TopologyMetadata",
      "methods": [
        {
          "name": "sourceTopicsForStore",
          "parameterTypes": [
            "java.lang.String",
            "java.lang.String"
          ]
        }
      ]
    }
  ]
}
```

### GraalVM version

GraalVM 25.0.4.1

### Operating system

macOS

### CPU architecture

AARCH64 (ARM64)

kimeta:
`run failed in claim/route_to_driver()[issue #9764]`

## Human intervention needed

The automated job failed for `org.apache.kafka:kafka-streams:4.3.1`. Preserved work: https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fkimeta%2Fhuman-intervention%2Fissue-9764-library-update-request-org.apache.kafka-kafka-streams-4.3.1-4d1ed3f5

The run is classified as `test_generation_failed`, most likely during `explore/generate_tests()` before Gradle verification. No `4.3.1` test project, metadata, execution metrics, dynamic-access report, or Gradle output was preserved; the branch contains only failure-analysis logs. Additionally, `pi.log` reports `Pi RPC exited before completing prompt (exit code -2)`, and the generation transcript ends during preliminary documentation inspection, so the follow-up diagnosis was also interrupted.

Inspect:
- `human-intervention-logs/org.apache.kafka:kafka-streams:4.3.1/failure-analysis/pi.log`
- `human-intervention-logs/org.apache.kafka:kafka-streams:4.3.1/failure-analysis/pi-generation-2026-09-03T09-18-34-984249Z.log`
- Recover the original run logs, if available, from `logs/org.apache.kafka:kafka-streams:4.3.1/`.

Manual next steps:
1. Check the worker/process logs for the signal or timeout that caused the Pi RPC exit and recover any missing original test-generation transcript.
2. Create the `4.3.1` test project from the existing `3.6.0` suite, adapt it for Kafka Streams 4.3.1, and run `compileTestJava` followed by `javaTest`.
[...]

PR #9863, opened 2026-09-04
Run reporter-requested metadata phase on every driver

### LINKED ISSUE #9862: Run reporter-requested metadata phase on every driver via the analysis agent
## Problem

The reporter-requested metadata phase has two gaps. Both are visible in #9777 (`org.apache.kafka:kafka-streams:4.3.1`, reported in #9764).

**It only exists on one route.** Everything referencing `issue_requested_metadata` is five files, and among drivers only `improve_library_coverage.py`. `library_update_router.py` picks the driver like this:

| condition | driver | requested-metadata phase |
|---|---|---|
| `tests/src/<g>/<a>/<version>/` exists | `improve_library_coverage` | yes |
| missing, baseline compile fails | `fix_javac_fail` | no |
| missing, java test fails | `fix_java_run_fail` | no |
| missing, native test fails | `fix_ni_run` | no |

#9764 asked for kafka-streams 4.3.1. No 4.3.1 suite existed, the 3.6.0 baseline failed to compile (Kafka 4.x dropped the `:test` classifier), so it routed to `fix_javac_fail` and the reporter's request never entered any prompt. #9777 ships neither requested entry:

- `org.apache.kafka.streams.KafkaStreams` field `topologyMetadata` — absent
- `org.apache.kafka.streams.processor.internals.TopologyMetadata` method `sourceTopicsForStore(String, String)` — absent

**Where it does run, success doesn't mean the request was satisfied.** `_run_issue_requested_metadata_phase` (`dynamic_access_iterative_strategy.py:153-224`) succeeds when `./gradlew test` reaches `nativeTest` or passes. It never checks that the traced metadata contains what the reporter asked for. The verification exists only at review time, in `skills/review-library-update-request/SKILL.md` step 5 — which is what produced the `human-intervention` label on #9777.

## Proposed changes

**1. Run the phase through the analysis agent.** Send the `issue-requested-metadata` prompt via `analysis_agent_run` (`ai_workflows/agents/agent_runtime.py`) instead of the strategy's main agent, so backend selection follows the analysis role rather than the strategy's agent field.

**2. Wire it into every driver.** `issue_requested_metadata_context` is already plumbed through `workflow_strategy.py:138`, so `fix_javac_fail`, `fix_java_run_fail`, `fix_ni_run`, `fix_java_fails` and `add_new_library_support` need to pass it and invoke the phase. No driver should be able to close a `library-update-request` without having attempted the reporter's request.

**3. Loosen the phase's rules to a generate-then-fill loop.** Today `forge/prompt_templates/issue_requested_metadata/issue-requested-metadata.md` tells the agent to write only tests and to touch nothing else. Replace with:

- write a test that exercises the requested metadata through a public API path,
- run `./gradlew generateMetadata`,
- if generation does not produce the requested entries, add them by hand,
- `nativeTest` must succeed.

**4. Verify by running the suite.** After the phase, run `./gradlew test -Pcoordinates=<coordinates>` and treat a failure as a phase failure, rather than accepting "reached `nativeTest`" as the success condition.

## Spec decision required before item 3

Item 3 contradicts §FS-test-contract.2.7 (*No hand-written reachability metadata*): "Tests and test authors never generate, write, or modify reachability metadata or Native Image config entries". The rule is cited by four prompt templates:

```
forge/prompt_templates/issue_requested_metadata/issue-requested-metadata.md:23
forge/prompt_templates/persistent/basic_iterative_rules.md:28
forge/prompt_templates/persistent/default_agent_rules.md:6
forge/prompt_templates/persistent/dynamic_access_rules.md:33
```

Two ways forward:

- **Narrow the contract.** Carve out an explicit exception in §FS-test-contract.2.7 for the reporter-requested phase: hand-written entries permitted only for needs inferred from the issue, only after `generateMetadata` failed to trace them, and only with a `typeReached` condition. Lands as its own commit ahead of the prompt change.
- **Keep the contract.** Treat a failure to trace the requested entry as a phase failure and surface it for human intervention, instead of letting the agent write it.

Note that §FS-test-contract.1.5 already makes every inferred need mandatory, so the contract currently requires an outcome the phase has no sanctioned way to reach when tracing misses it. That tension is the real subject of this decision.

kimeta:
Closes #9862

## What this does

The reporter-requested metadata phase existed on exactly one route. `library_update_router.py` sends an issue whose requested-version suite is missing to `fix_javac_fail`, `fix_java_run_fail`, or `fix_ni_run` — none of which ever saw the reporter's request. That is how #9764 (`org.apache.kafka:kafka-streams:4.3.1`) produced #9777: Kafka 4.x dropped the `:test` classifier, the 3.6.0 baseline failed to compile, the run routed to `fix_javac_fail`, and neither requested entry ever reached a prompt. Where the phase *did* run, it succeeded on "reached `nativeTest`" and never checked that the request was actually satisfied — the verification lived only in `skills/review-library-update-request/`, at review time, which is what labelled #9777 `human-intervention`.

Every driver now carries the request, the phase runs on the analysis role, and it succeeds only when the coordinate's suite passes.

## Where the phase now lives

It moved from `DynamicAccessIterativeStrategy` up to `WorkflowStrategy`, so any workflow engine has it, and it carries its own prompt templates — a strategy that never declared them still gets the request attempted. Each driver calls `ensure_issue_requested_metadata_phase()` before finalization; the call is idempotent, so the composite workflow that already ran it does not run it twice.

| driver | before | now |
|---|---|---|
| `improve_library_coverage` | yes | yes |
| `fix_javac_fail` | no | yes |
| `fix_java_run_fail` | no | yes |
| `fix_ni_run` | no | yes |
| `add_new_library_support` | no | yes |

The dispatcher extracts the issue body for all eight driver-invocation sites, not just the library-update coverage route. That reverses "Limit issue body context to library updates" from #6848, which is what the issue asks for.

## Generate, then fill

Tracing is tried before anything is hand-written, and the **engine** owns every Gradle command so §forge/AR-forge-strategy-agent-boundary still holds:

1. analysis agent writes tests that drive the requested access through a public API path
2. engine runs `./gradlew generateMetadata -Pcoordinates=<coordinates>`
3. analysis agent is shown the traced result and hand-writes **only** the requested entries generation missed
4. engine runs `./gradlew test -Pcoordinates=<coordinates>`; the phase succeeds only on a full pass

Step 4 is the real change to the success condition. `nativeTest` failing is no longer "close enough" — it is a phase failure that reverts to the phase checkpoint and records its location.

The prompt runs through `analysis_agent_run` rather than the strategy's own agent field, so backend selection follows the analysis role (§forge/FS-forge-agent-runtime-selection). Strategy persistent instructions are deliberately *not* passed: they forbid the one hand-written entry this phase is allowed to make.

## The spec change

Step 3 contradicted §FS-test-contract.2.7 (*No hand-written reachability metadata*), cited by four prompt templates. The tension was already there: §FS-test-contract.1.5 makes every inferred need mandatory, and §forge/AR-forge-driver-queues.2.1 already said the agent "includes the requested metadata when generation did not already produce it" — so the FS and the AR disagreed with each other. The contract is narrowed rather than kept, and it lands as its own commit (`867afb6`) ahead of the code:

> The single exception is the reporter-requested metadata phase. […] only for needs inferred from the issue, only after generation has been run and missed them, only with a `typeReached` condition naming a type reached before the access occurs; no other condition kind is permitted, and never in place of the test.

Nothing under `src/test/resources/META-INF/native-image` is ever written, and no other phase or author may hand-write metadata.


kimeta (commented):
Found two functional blockers and one accounting defect:

- [P1] `forge/ai_workflows/drivers/fix_ni_run.py:563-569`: a false result from `ensure_issue_requested_metadata_phase()` is only logged. The driver still calls `finalize_run()`, derives success solely from finalization, writes success metrics, and returns 0. A Native Image route can therefore publish and close the issue after the mandatory reporter-requested phase failed. Please make phase failure force the run to fail or skip finalization/publication.

- [P1] `forge/ai_workflows/core/workflow_strategy.py:896-918`: the result of `generateMetadata` is logged but never gated. If the initial generated test does not compile, the fill agent is still allowed to hand-write metadata; the later test-repair loop never reruns `generateMetadata`, so the suite can pass with a manual entry that was never shown to be missed by a successful trace. That violates the narrow generate-then-fill exception. Please repair/retry generation until it succeeds before running the fill turn, and rerun generation after any test repair.

- [P2] `forge/ai_workflows/core/workflow_strategy.py:976-992`: every `analysis_agent_run()` result carries input/output/cache usage, but this method discards it. Durable metrics are later built from the strategy test-agent object, so the two or more new analysis-role turns disappear from token and cost accounting, and the effective analysis selection is not represented. Please accumulate these turn results into the workflow metrics.

Validation: the focused changed-area suite passed 226 tests; `grund check` and `git diff --check` passed. The full Forge suite ran 834 tests and had two failures in unchanged areas (`test_native_test_verification` path rendering and `test_publish_new_library_support` placeholder validation).

kimeta (commented):
Re-review of `a2ad396d9d`:

The two previous P1 findings are fixed. Reporter-requested work is now driver-owned after strategy success, Native Image phase failure blocks finalization, and fill is gated on successful generation. The new regressions cover those paths.

One code finding remains:

- [P2] `forge/ai_workflows/core/workflow_strategy.py:1028-1053`: `analysis_agent_run()` returns input, output, and cached-token usage for every reporter-phase turn, but `_run_issue_requested_metadata_turn()` still discards those counters. Durable run metrics are built from the separate strategy test-agent object, so these analysis turns and their cost remain absent.

The PR is also currently unmergeable because current `master` added dynamic-access failure-location recording in #9840. Conflict resolution in `increase_dynamic_access_coverage_strategy.py` must retain `self._record_dynamic_access_failure()` in both failure branches, and the corresponding test resolution must retain the `failed_run_location()` expectation while also asserting that the reporter phase is not called.

Validation on the current PR head: 229 focused tests passed, `grund check` passed, and `git diff --check origin/master...HEAD` passed. GitHub currently reports `CONFLICTING`; only the failing OCA check is present.

cugurm (approved):

PR #9865, opened 2026-09-04
Improve optimistic dynamic-access handoff

### LINKED ISSUE #9864: Preserve partial bulk coverage in composite runs
## Problem

The optimistic dynamic-access composite preserves a successful bulk result only when bulk fully completes at least one class. It loses partial call-site progress within a class.

Observed for `org.springframework.cloud:spring-cloud-vault-config:5.0.2`:

- bulk reached the native gate and produced `1/2` coverage;
- the remaining call stayed in `VaultConfigDataLoader`;
- iterative refinement exhausted that class without further coverage;
- the composite returned failure because neither iterative `successful_classes` nor bulk `completedClasses` was non-empty.

The generated branch and final report still contain the bulk contribution. Only the terminal success calculation fails to credit it.

## Expected behavior

Compare the bulk baseline and final reports at call-site level. Any positive bulk coverage delta should count as acceptable composite progress, even when no class becomes fully covered. An unchanged bulk report followed by zero-yield iterative refinement must continue to fail.

This extends the whole-class fix from #9776 and aligns partial bulk progress with the existing iterative behavior, which accepts and commits partial class coverage.

## Acceptance criteria

- A bulk change from `0/2` to `1/2`, followed by an exhausted iterative remainder, succeeds.
- A bulk result that remains `1/2`, followed by an exhausted iterative remainder, fails.
- Hard report, test, and native-gate failures remain failures.
- Regression tests cover both positive-delta and zero-delta handoffs.

Relevant contracts: [§forge/FS-forge-chunked-dynamic-access](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/functional-spec/functional-spec.md#fs-forge-chunked-dynamic-access) and [§forge/AR-dynamic-access-composite](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/architecture/workflows.md#ar-dynamic-access-composite-composite-fix-then-explore).

kimeta:
Closes #9864

## What this does

The optimistic dynamic-access composite previously had two avoidable failure and cost cases around its bulk-to-iterative handoff. It lost a call-site gain when the affected class remained partial, and it still spent three broad bulk attempts on reports small enough for targeted iterative work.

The handoff now preserves partial bulk progress and routes small reports directly to iterative generation. This aligns the composite with [§forge/FS-forge-chunked-dynamic-access](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/functional-spec/functional-spec.md#fs-forge-chunked-dynamic-access) and [§forge/AR-dynamic-access-composite](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/architecture/workflows.md#ar-dynamic-access-composite-composite-fix-then-explore).

## How progress crosses the handoff

Bulk records the covered-call delta between its baseline and last gated report. Iterative refinement consults that delta only at natural phase completion, so `0/2 → 1/2` remains successful while `1/2 → 1/2` and hard report, test, or native-gate failures still fail.

## How small reports are routed

The six optimistic strategy configurations set `bulk-min-uncovered-classes` to `5`. Their composite prechecks the dynamic-access report and:

- sends `0–4` uncovered classes directly to iterative refinement;
- runs bulk first for `5+` classes;
- reuses the prechecked report as the bulk baseline, avoiding a duplicate refresh;
- retains the existing bulk fallback for unavailable or empty reports.

Pure-bulk strategies are unchanged.

### DIFF
--- a/forge/utility_scripts/dynamic_access_report.py
+++ b/forge/utility_scripts/dynamic_access_report.py
@@ -87,13 +87,14 @@ class BulkDynamicAccessProgress:
     """Post-gate class progress from one bulk phase.
 
     The workflow compares its baseline with the last successfully gated report
-    and carries that exact report into the composite boundary
+    and carries that exact report and covered-call gain into the composite boundary
     (§AR-dynamic-access-bulk).
     """
 
     final_report: DynamicAccessCoverageReport
     completed_classes: tuple[str, ...]
     remaining_classes: tuple[str, ...]
+    covered_call_gain: int
 
 
 def compute_bulk_dynamic_access_progress(
@@ -123,7 +124,13 @@ def compute_bulk_dynamic_access_progress(
             and class_coverage.class_name not in processed_classes
         )
     )
-    return BulkDynamicAccessProgress(final_report, completed_classes, remaining_classes)
+    covered_call_gain: int = max(final_report.covered_calls - initial_report.covered_calls, 0)
+    return BulkDynamicAccessProgress(
+        final_report,
+        completed_classes,
+        remaining_classes,
+        covered_call_gain,
+    )
 
 
 @dataclass(frozen=True)
--- a/forge/utility_scripts/dynamic_access_report.py
+++ b/forge/utility_scripts/dynamic_access_report.py
@@ -97,6 +97,26 @@ class BulkDynamicAccessProgress:
     covered_call_gain: int
 
 
+def remaining_uncovered_classes(
+        report: DynamicAccessCoverageReport,
+        processed_classes: set[str],
+) -> tuple[str, ...]:
+    """Return uncovered classes no phase has processed yet.
+
+    The remainder a phase can still take is the report's uncovered classes
+    minus the exhaust report and continuation marker's processed set
+    (§AR-dynamic-access-composite).
+    """
+    return tuple(
+        class_coverage.class_name
+        for class_coverage in report.classes
+        if (
+            class_coverage.uncovered_calls > 0
+            and class_coverage.class_name not in processed_classes
+        )
+    )
+
+
 def compute_bulk_dynamic_access_progress(
         initial_report: DynamicAccessCoverageReport,
         final_report: DynamicAccessCoverageReport,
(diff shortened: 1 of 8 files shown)


forge/utility_scripts/dynamic_access_report.py
```diff
@@ -115,15 +136,14 @@ def compute_bulk_dynamic_access_progress(
...
     ))
-    remaining_classes: tuple[str, ...] = tuple(
-        class_coverage.class_name
-        for class_coverage in final_report.classes
-        if (
-            class_coverage.uncovered_calls > 0
-            and class_coverage.class_name not in processed_classes
-        )
+    remaining_classes: tuple[str, ...] = remaining_uncovered_classes(final_report, processed_classes)
+    covered_call_gain: int = max(final_report.covered_calls - initial_report.covered_calls, 0)
```
kimeta, line 140: Retracted: I interpreted progress as covering only call sites that were already present and uncovered in the baseline. This workflow measures total covered calls, so a change from 0 to 1 is genuine net coverage progress even if generated tests expand the reachable call-site set. The proposed aggregate delta matches the PR architecture’s “total covered-call gain” semantics; this is not a blocking correctness issue.

kimeta (commented):
I found one blocking correctness issue inline. The branch also needs rebasing onto current `master`: #9863 moved reporter-requested metadata handling into `WorkflowStrategy`, so conflict resolution should retain that ownership rather than restore the removed iterative-only phase. Validation on the PR head: 39 targeted tests passed and `grund check` passed; the full 843-test Forge suite had one unrelated failure that reproduces on `master`.

kimeta (commented):
Correction to my prior review: I have retracted the inline P1. The aggregate `covered_calls` delta correctly represents net coverage progress for this workflow; a transition from 0 to 1 is improvement even when generated tests expand the reachable call-site set. I found no remaining code-correctness issue on the PR head. The branch still needs rebasing onto current `master` to resolve the #9863 conflicts.

cugurm (approved):

PR #10024, opened 2026-09-13
Scope benchmark descriptors and pin a cell's analysis agent

kimeta:
Benchmark cells for one library collided on a single descriptor path, ran their
metadata repairs on whatever agent the launching machine happened to export, and
lost long builds to backgrounded Gradle. This fixes those, and the path filter
that silently kept the first fix from working.

## Descriptor path is scoped by publication

A coordinate carries one result per executed cell, so
`stats/<group>/<artifact>/<version>/forge-publication.json` named the same file
for every run of that library. The second benchmark PR to be opened conflicted
with the first on a file neither had any reason to touch. Benchmark descriptors
now nest under the publication ID; issue-driven publications keep the flat path,
and `publisher.py` validates the shape that matches the task type.

`forge-branch-ready.yml` selects publication branches on that path, and its
filter carried three wildcards. The nested descriptor has four segments, so the
workflow never started, `forge-open-pr.yml` never received its `workflow_run`,
and the branch was pushed without a pull request ever being opened - with no
failing job to show for it, because no job ran. The filter now lists both depths.
This was caught on run `20260912T221227Z-l4-claude-code-opus-5-high-c43e4e9c08`,
whose branch is pushed and still unpublished.

## A cell's repairs run on the cell's agent

`analysis_agent_run` resolves its backend from `FORGE_ANALYSIS_*`, which nothing
set, so repairs ran on the launching machine's ambient default. A cell whose
coverage agent is one model while its metadata repairs are performed by another
measures two agents and attributes the result to one. The launcher now pins the
analysis role to the cell's own family, model, and thinking level, and sends a
provider only to backends for which one means something. The executable is left
unset so each backend's registered default applies.

## Agents wait for their builds

Headless invocations ended their turn to wait on a backgrounded Gradle task.
Ending the turn is the process exiting, so the work lost its supervisor with no
later turn to wake up in; one run stalled three times this way. The task template
now requires foreground execution with a sufficient timeout, artifact
confirmation before reporting success, and the state output written before the
final message.

## Provider prefix

The Rhei target string carried a provider prefix for backends that authenticate
themselves, which rejected the model.

## Verification

- `grund check`: success
- `forge/tests`: 940 tests, same 5 failures and 5 errors as `master` (937 tests);
  no new failure. The pre-existing ones are unrelated to this branch.
- Path filter checked against GitHub glob semantics for both descriptor shapes.


kimeta:
Superseded by #10027, which carries these seven commits plus the workflow-hardening series built on top of them (gate-backed native-metadata prep, program-state API inventory and publication, cache-write token accounting, and the benchmark results union-merge).

PR #10034, opened 2026-09-14
Remove stray benchmark publication descriptors

kimeta:
## What this does

Every PR that touches a `stats/<g>/<a>/<v>/stats.json` currently fails the **Validate library stats JSON** check against a clean merge with master: benchmark runs for `com.h2database:h2:2.1.210` committed forge publication descriptors under `stats/com.h2database/h2/2.1.210/forge-benchmark-<timestamp>-<hash>/`, a depth the stats validator (§forge/AR-actions-publication) does not accept. The benchmark PRs themselves merged green because the check's change filter only watches `stats/*/*/*/stats.json` — validation never ran for them, and it now fires for every unrelated PR instead. This removes the two stray descriptor directories; `./gradlew validateLibraryStats` passes on this branch. The benchmark results in `code-coverage-benchmarks/` are untouched.

## Where the stray files came from

h2 2.1.210 already carries a canonical `stats/com.h2database/h2/2.1.210/forge-publication.json` from its coverage work, so the benchmark publisher could not write to the canonical path and nested its descriptor into a unique `forge-benchmark-*` subdirectory to avoid clobbering it. That collision workaround lands outside the layout the validator recognizes — the publisher's own tests expect the canonical depth-4 path.

## Open questions

1. **The publisher will recreate these on the next benchmark run against an already-published coordinate** — it should either skip committing the descriptor when the canonical path is occupied or fold the provenance into the `code-coverage-benchmarks/` record. Recommend a follow-up forge issue rather than widening the validator.


kimeta:
Superseded by #10037: the publication-scoped descriptor path is required by the spec (§forge/FS-code-coverage-benchmarking.3), so the validator was taught to recognize it rather than deleting the files. This branch is also stale — master now carries four nested descriptors, not two.

PR #9983, opened 2026-09-10
[Benchmark] Record com.h2database:h2:2.1.210 result (gpt-5.6-sol, high)

graalvmbot:
## Code coverage benchmark result

- Run: `20260909T005701Z-l4-pi-gpt-5.6-sol-high-c2686ca021`
- Coordinate: `com.h2database:h2:2.1.210`
- Status: `success`
- Agent: `pi`
- Configured model: `gpt-5.6-sol`
- Observed model: `openai-codex/gpt-5.6-sol`
- Thinking level: `high`
- Benchmark suite commit: `92a2a4fa60b2d6532fa533f2d4f8f795dd28a1cb`
- Runner commit: `5ec491a17e6c740a39552b05373b16316642e085`

## Result

- Covered methods: 2,688 → 11,019
- Methods gained: 8,331
- Coverage gained: 69.76pp
- Method universe: 11,942
- Input tokens: 6,459,340
- Cached input tokens: 288,110,976
- Output tokens: 890,191
Forge-Publication-ID: forge-benchmark-20260910015334-b7cfcebd730e

PR #10038, opened 2026-09-15
Re-anchor descriptor base on benchmark conflict refresh

### LINKED ISSUE #10036: Benchmark conflict refresh leaves descriptor base_commit stale, failing Branch Ready
## What happens

The review worker's benchmark conflict refresh (#10027, #10025) union-merges the per-coordinate results file and pushes a merge commit to the publication head. The merge brings the base branch's descriptors under `stats/**` into the head, so the push re-triggers `Forge Branch Ready` on the refreshed commit.

The branch's own descriptor still records the `base_commit` the run originally rebased onto. `_validate_benchmark_result_publication` requires the head's results list to equal the list at that `base_commit` plus the descriptor's one result — but the base branch gained entries since `base_commit`, which is the very reason the refresh ran. The union-merged list therefore never matches, and Branch Ready fails on every refreshed benchmark head:

```
Benchmark result list must be the sorted base list plus the descriptor result
```

The failure is transient noise, not a merge blocker — Branch Ready is not a required check — but it marks a correctly refreshed publication red and hides real validation failures behind expected ones.

## Verified

PRs #10018 and #9983 were refreshed manually with the union merge plus one extra step the worker does not perform: setting the descriptor's `base_commit` to the merged base SHA inside the same merge commit. Branch Ready revalidated both heads green, and both merged cleanly.

## Fix options

1. The conflict refresh updates the publication descriptor's `base_commit` to the merged base branch SHA and stages it with the resolved results file. Matches what the manual refreshes did; keeps the validator strict.
2. The validator computes the base list from `merge-base(master, head)` instead of the descriptor's `base_commit` for benchmark-result publications.

Option 1 keeps the descriptor truthful about what the head was validated against and needs no trusted-side change.

graalvmbot:
Fixes #10036.

## What this does

The conflict refresh from #10027 union-merges the per-coordinate benchmark results file but leaves the head's publication descriptor pointing at the base commit the run originally rebased onto. Branch Ready validates the result list as *the descriptor's base list plus this run's one entry* — and the entries the refresh just merged in guarantee that comparison fails, so every worker-refreshed benchmark head reports a red readiness run. #10018 and #9983 merged cleanly only because their refreshes were done by hand with the base commit rewritten.

This makes the worker's refresh do the same: after resolving the results file by keyed union, it rewrites the head's one descriptor's `base_commit` to the base commit it merged in and stages it inside the same merge commit ([§forge/FS-automated-pr-review](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto-benchmark-refresh-base-commit/forge/docs/functional-spec/publication.md)). A head that does not carry exactly one descriptor escalates to human intervention, unchanged.

## Interim, by design

This keeps the committed descriptor honest instead of questioning its existence. The durable direction is descriptor-free benchmark publication: the result entry already lives keyed by run ID under `code-coverage-benchmarks/`, so readiness validation can check the appended entry straight from the merge-base diff — no base-commit anchor left to go stale. That is a spec change to `§forge/FS-code-coverage-benchmarking.3` and lands separately; this code is deleted wholesale when it does.


kimeta:
Closing as won't do. Superseded by #10043, which removed the benchmark publication descriptor entirely and validates from the merge-base diff, so there is no `base_commit` left to re-anchor. #10036 is fixed at the root and this interim repair path no longer applies.

PR #10086, opened 2026-09-16
Keep benchmark input tokens disjoint from cached input

kimeta:
## What this does

`_phase_tokens` published Rhei's `tokens.input.total` verbatim as `input`. Rhei
reports that total **inclusive** of the cached classes on newer runtimes and
**exclusive** of them on older ones, so `input` silently means two different
things depending on when a run was recorded — while the spec requires one
meaning:

> Cached input must remain separate from ordinary input.
>
> — [§forge/FS-code-coverage-benchmarking.4](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto/forge/docs/functional-spec/benchmarking.md)

This subtracts the cached classes from a total that contains them, so ordinary
input stays disjoint from `cachedInputRead` and `cachedInputWrite` in every
record.

The drift is not theoretical. Across the twelve runs recorded for
`com.h2database:h2:2.1.210`, the six older entries report `cachedInputRead`
larger than `input` — impossible if the cached reads were inside that total —
and the six newer ones report it smaller:

| Recorded | `input` | `cachedInputRead` | Reads as |
|---|---:|---:|---|
| sol, high, guided | 6,459,340 | 288,110,976 | exclusive |
| sol, medium, guided | 52,311,342 | 49,524,224 | inclusive |

Comparing those two `input` values as-is suggests the guided arm spent 8x fewer
input tokens. On the disjoint basis it is 6,459,340 against 2,787,118 — the
opposite direction.

## How the two shapes are told apart

The invocation record carries no runtime version, so the arithmetic decides: a
total at least as large as its cached classes contains them and is restated; a
smaller total already excludes them and is published as recorded. That keeps
both runtimes correct without pinning a version, and leaves `cachedInputRead`,
`cachedInputWrite`, and `output` untouched — only `input` ever changed meaning.

## A second, separate gap this does not close

`cachedInputWrite` only exists on records written after #10027, and those are
all `pi` runs, where it is legitimately `0` — codex has no separate cache-write
class. The two `claude-code` runs, which do have cache writes, predate the
field, so **no published record has ever carried a non-zero cache write**.

The preserved Rhei accounting still has the numbers, and they are not small:

| Run | published `input` | `cache_write` on disk | fresh input |
|---|---:|---:|---:|
| opus-5, medium | 6,142 | 6,030,864 | 6,037,006 |
| opus-5, high | 7,850 | 7,830,250 | 7,838,100 |
| sol, high (pi) | 6,459,340 | 0 | 6,459,340 |

Anthropic books a turn's new content as `cache_creation` where codex books it as
plain `input`, so the published figures understate those two runs by roughly
1000x on the input side — and cache writes are the *premium* input class
(1.25x base at 5-minute TTL, 2x at 1-hour), not a discounted one.

## Open questions

1. **Restating the two `claude-code` records.** The cache writes above are
    recoverable from
    `forge/local_repositories/code_coverage_benchmarks/<runId>/code-coverage-99000/runtime/accounting/invocations/`,
    so this needs a data edit, not a re-run. I would do it as a follow-up rather
    than here — it has to work with the run-ID conflict rule, and keeping it
    separate lets the behavior fix land on its own.
2. **Is subtracting `cache_write` from an inclusive total right?** It is
    subtracted here for symmetry with `cached_read`, but no post-#10027
    `claude-code` run exists yet, so the inclusive shape has never been observed
    with a non-zero cache write. In the one pre-#10027 Anthropic sample,
    `input.total` (432) excludes both classes. The `total >= cached` guard means
    a wrong guess here cannot corrupt an exclusive record, but it should be
    confirmed against a real inclusive Anthropic record before anyone leans on
    that column.


kimeta:
Superseded by #10177, which carries this commit rebased onto master.

The change had to move: master's `faad895095` split `code_coverage_benchmark.py`, so `_phase_tokens` and the new `_ordinary_input` now live in `code_coverage_benchmark_metrics.py`, and the test moved to `test_code_coverage_benchmark_runs.py` alongside the rest of `CodeCoverageBenchmarkMetricsTests`. #10177 also folds in #10155, since both rewrite the same `total()` path, and it carries this description forward including the deferred restatements.

Closing rather than merging because this branch is `mm/auto`, which misses the `kimeta:mm/auto-*` allowlist in `MAINTAINER_PR_BRANCHES` by the trailing hyphen — so `maintainer-open-pr.yml` never opened it as graalvmbot and its author could not review it.

Issue #10090, opened 2026-09-17
[bug] Forge finalization fails on a stale Gradle daemon: checkMetadataFiles cannot load com/networknt/schema/JsonSchema


cugurm:
## Summary

A Forge run failed issue #10073 (`io.micronaut.security:micronaut-security-oauth2:5.3.2`) in finalization because `checkMetadataFiles` could not even create its task:

```
Execution failed for task ':checkMetadataFiles'.
> Could not create task ':checkMetadataFiles_io.micronaut.security_micronaut-security-oauth2_5.3.2_83444056673848'.
   > Could not create task of type 'MetadataFilesCheckerTask'.
      > Could not generate a decorated class for type MetadataFilesCheckerTask.
         > com/networknt/schema/JsonSchema
```

The failure is instant (~600 ms), happened 5/5 times (three repair-agent attempts could not fix it because the validation never reached the metadata), and the issue was moved to `human-intervention` with the work preserved on
[`ai/cugurm/human-intervention/issue-10073-library-new-request-io.micronaut.security-micronaut-security-oauth2-5.3.2-7b73a58a`](https://github.com/oracle/graalvm-reachability-metadata/tree/ai%2Fcugurm%2Fhuman-intervention%2Fissue-10073-library-new-request-io.micronaut.security-micronaut-security-oauth2-5.3.2-7b73a58a).

## Root cause: stale Gradle daemon, not the branch

- The same `checkMetadataFiles` gate passed for issue #10076 about an hour earlier in the same cycle, on the same master commit (`012c7cf9634f`).
- Reproduced on the preserved branch with Forge's Gradle user home (`/tmp/metadata-forge-gradle/<hash>`): the daemon-backed run fails identically; `./gradlew --no-daemon checkMetadataFiles -Pcoordinates=io.micronaut.security:micronaut-security-oauth2:5.3.2` on the same branch and caches **passes**.
- `com.networknt:json-schema-validator:1.5.1` is a declared `implementation` dependency of `tests/tck-build-logic`. Every Forge worktree rebuilds that included build, and a long-lived daemon (five were alive in the Forge Gradle home) that already had the previous build-logic classloader ends up with an incomplete classpath for the decorated task class.

So the generated tests and metadata on the preserved branch were never actually validated; the human-intervention verdict is a false negative caused by daemon state.

## Suggested fixes

[...]

PR #10098, opened 2026-09-18
Auto PR: forge-recycle-gradle-daemons

graalvmbot:
## Commits

- Recycle Gradle daemons per claimed issue and retry class-loading gate failures

---

Authored by @cugurm on `mm/auto-forge-recycle-gradle-daemons`, opened by `graalvmbot` so the author can review it.


cugurm (approved):
Reviewed the diff: the claim-time `--stop` is scoped to Forge's shared Gradle user home (keyed on the git common dir, so the operator's own daemons are untouched) and is best-effort; the finalization retry keys on the exact stale-daemon signature reproduced in #10090 and reruns exactly once. Spec points extended rather than widened, tests do not launch Gradle, CI green.

Issue #10099, opened 2026-09-18
[bug] Forge crashes at queue processing: module 'concurrent' has no attribute 'futures' in dispatcher/issue_processing.py


cugurm:
## Summary

Every `do-work.sh` cycle on current master (`187b66c187`) dies ~8 s after "Processing configured work queues via forge_metadata", before any issue is claimed:

```
  File "forge/dispatcher/issue_processing.py", line 270, in process_issues_with_label
    executor = concurrent.futures.ThreadPoolExecutor(max_workers=parallelism)
AttributeError: module 'concurrent' has no attribute 'futures'
```

## Cause

#10088 (`faad895095`, "Adopt fissile file-size budgets for agent-facing infrastructure") moved the issue-queue loop out of `forge_metadata.py` into `forge/dispatcher/issue_processing.py`. That module does `import concurrent` (line 10) and then uses `concurrent.futures.Future`, `ThreadPoolExecutor`, `wait` and `FIRST_COMPLETED`. `import concurrent` does not load the `concurrent.futures` submodule; it only worked before because another module in the same process imported `concurrent.futures`, and after the split nothing under `forge/` does (`grep -rn "^import concurrent.futures\|^from concurrent" forge --include=*.py` finds nothing outside tests). Reproduced on a fresh interpreter: `import dispatcher.issue_processing; hasattr(concurrent, "futures")` → `False`.

The test suite did not catch it because pytest (or a test module) imports `concurrent.futures` earlier in the process.

## Fix

`import concurrent.futures` in `forge/dispatcher/issue_processing.py`, plus a regression test that imports the module in a subprocess (fresh interpreter) and resolves `concurrent.futures.ThreadPoolExecutor` through it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10100, opened 2026-09-18
Auto PR: forge-import-concurrent-futures

graalvmbot:
## Commits

- Import concurrent.futures in the Forge issue queue loop

---

Authored by @cugurm on `mm/auto-forge-import-concurrent-futures`, opened by `graalvmbot` so the author can review it.


cugurm (approved):
One-line import fix for the crash in #10099, verified red/green with the new fresh-interpreter regression test; CI green.

PR #10088, opened 2026-09-16
Adopt fissile file-size budgets for agent-facing infrastructure

graalvmbot:
Closes #10087.

Agents pay for oversized files twice: reading costs tokens, and a small edit in a monolithic module drags the whole file into context. This PR adopts [fissile](https://github.com/agent-grounds/fissile) to cap file sizes across the hand-written infrastructure — `forge/`, `tests/tck-build-logic/`, and the docs — and brings every file in that scope under budget or under a recorded, justified exception. Shipped metadata and per-library test projects are deliberately out of scope.

### File splits

All splits are pure refactors: code moved verbatim along responsibility seams, imports and test patch targets retargeted, no behavior change.

| Area | Before | After |
|---|---|---|
| 21 forge Python modules (550–2,000 lines) | monoliths | split into seam-aligned sibling modules |
| 8 TCK Java classes and tests | up to 1,180 lines | split, public APIs unchanged |
| `forge/forge_metadata.py` | 9,223 lines | 513-line CLI shim over a new `forge/dispatcher/` package (34 modules), declared in [§forge/AR-forge-dispatcher-decomposition](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto-fissile-adoption/forge/docs/architecture/dispatcher-package.md) |
| `forge/tests/test_forge_metadata.py` | 4,495 lines | 15 per-module test files; test class and method names preserved |

### Exceptions

`docs/file-size-human-exceptions.toml` gains structural entries for the two TCK Gradle script plugins (splitting a precompiled script plugin over shared script state is not provably semantics-preserving) and deferred entries for three forge spec documents, each naming the condition that retires it. Entries retired by the splits are removed.

### Configuration and guidance

`.agents/fissile.toml` scopes the budgets (source 450/700 lines soft/hard, docs 250/500) and `AGENTS.md` instructs agents to run `fissile check --staged`; the pre-commit hook remains a per-machine `fissile init` step.

### Verification

- forge suite: 955 tests, failure set identical to master's pre-existing baseline after every commit
- `./gradlew -p tests/tck-build-logic test`, scoped `checkstyle` and `checkMetadataFiles`: green
- `grund check`: success after every commit
- `fissile audit`: exits clean — zero hard findings in scope

### Review revision

[...]

Issue #10048, opened 2026-09-16
[Automation] javac compile fails for org.xerial.snappy:snappy-java:1.1.0.1


graalvmbot:
## javac compile on ubuntu-latest / JDK 25 / mode current-defaults

Failure kind: javac compile
OS: ubuntu-latest
JDK: 25
Native-image mode: current-defaults
Reproducer: `env GVM_TCK_LV=1.1.0.1 GVM_TCK_NATIVE_IMAGE_MODE=current-defaults ./gradlew clean compileTestJava -Pcoordinates=org.xerial.snappy:snappy-java:1.0.4.1`
Runner log: https://github.com/oracle/graalvm-reachability-metadata/actions/runs/35039064414
Last 300 lines of the log:
```console
  "1.1.2.4",
  "1.1.2.5",
  "1.1.2.6",
  "1.1.2",
  "1.1.3-M1",
  "1.1.3-M2",
  "1.1.4",
  "1.1.7.1",
  "1.1.7.2",
  "1.1.7.3",
  "1.1.7.4",
  "1.1.7.5",
  "1.1.7.6",
  "1.1.7.7",
  "1.1.7.8",
  "1.1.7",
  "1.1.8.1"
]
Parsed version count: 30
Parsed versions: 1.1.0.1 1.1.0 1.1.1.1 1.1.1.2 1.1.1.3 1.1.1.4 1.1.1.5 1.1.1.6 1.1.1.7 1.1.1 1.1.2.1 1.1.2.2 1.1.2.3 1.1.2.4 1.1.2.5 1.1.2.6 1.1.2 1.1.3-M1 1.1.3-M2 1.1.4 1.1.7.1 1.1.7.2 1.1.7.3 1.1.7.4 1.1.7.5 1.1.7.6 1.1.7.7 1.1.7.8 1.1.7 1.1.8.1
Timeout per Gradle invocation: 5m
========================================================================================
 Testing org.xerial.snappy:snappy-java:1.0.4.1:1.1.0.1
========================================================================================
========================================================================================
 org.xerial.snappy:snappy-java:1.0.4.1:1.1.0.1 stage native-image run
========================================================================================
Starting stage 'native-image run' for 1.1.0.1 at 2026-09-16T00:16:56Z
[...]

PR #10105, opened 2026-09-18
Auto PR: harness-index-update-utils

graalvmbot:
## Commits

- Call the moved index update helpers from the metadata index Gradle tasks

---

Authored by @cugurm on `mm/auto-harness-index-update-utils`, opened by `graalvmbot` so the author can review it.


cugurm (approved):
Two Groovy call sites retargeted to the class the helpers moved to in #10088, unused import dropped; verified by running addLibraryAsLatestMetadataIndexJson on the exact coordinates that failed (#10048) and checking the index.json result. Full build-logic CI matrix green.

PR #10114, opened 2026-09-18
JNA: add reflection metadata for the com.sun.jna.ptr types direct mapping instantiates

ArloL:
## What does this PR do?

Registers the no-arg constructor of the eight `com.sun.jna.ptr` by-reference types, which direct mapping instantiates reflectively.

`Native.register` builds a default value for every `NativeMapped` parameter through `NativeMappedConverter`, and every `ByReference` is a `PointerType`, hence a `NativeMapped`. #9121 registered `NativeLong` for that call site; without the rest, direct-mapping a function with an out parameter fails at `NativeMappedConverter.defaultValue` with `NoSuchMethodException: com.sun.jna.ptr.IntByReference.<init>()`.

All eight entries are needed. `DirectMappedCLibrary` gains one `memcpy` overload per type — `memcpy` takes `void *`, so a pointer to a single value of any of them is an honest call — and removing any single entry fails the test with that type's own `NoSuchMethodException`.

`stats.json` was regenerated on linux-x86-64. Dynamic access is unchanged at 35 of 48; library coverage carries two deltas:

| | instructions | lines | methods |
| --- | --- | --- | --- |
| this change | +51 | +11 | 0 |
| platform drift | +17 | +7 | +1 |

The committed numbers came from aarch64 macOS (#9121), which can no longer produce them: `FunctionNativeCoverageTest` opens `/com/sun/jna/linux-x86-64/libjnidispatch.so`, so `codeCoverageTest` fails off linux-x86-64.

The `future-defaults-all` lane fails for JNA on master too (#7741, `Bulk queries can only be set with 'name'`).

[`DirectMappingTest`](tests/src/net.java.dev.jna/jna/5.8.0/src/test/java/net_java_dev_jna/jna/DirectMappingTest.java) direct-maps three libc functions on statically known arguments: `atol` and `labs`, both pre-existing, and `memcpy`, which copies one value's worth of bytes between two JNA-allocated buffers ([`DirectMappedCLibrary`](tests/src/net.java.dev.jna/jna/5.8.0/src/test/java/net_java_dev_jna/jna/DirectMappedCLibrary.java)).

https://claude.ai/code/session_01BrmdJ4eL3RkWyzfwvn6o3a


kimeta (approved):
Verified locally: full JVM + native test suite passes on the PR branch, and removing any single ptr entry reproduces the predicted NoSuchMethodException from NativeMappedConverter.defaultValue. Metadata is additive only and coverage improves.

PR #10129, opened 2026-09-20
Give every Forge worktree its own Gradle user home with a shared read-only dependency cache

graalvmbot:
## Problem

All Forge worktrees of a checkout shared one Gradle user home and therefore one daemon registry. Any worktree could reuse a daemon that still held another worktree's build logic (the stale-daemon failures behind #10090/#10098), and the `gradlew --stop` that #10098 runs on every new workspace stops **every** daemon of the checkout — busy ones included, as seen in run logs on 2026-09-20 (`Daemon is stopping immediately` on a daemon mid-build).

## Change

Bound Gradle daemons by worktree, keep dependencies shared — using only documented Gradle mechanisms:

- **One Gradle user home per worktree**: a linked worktree builds in `<checkout home>/worktree-homes/<name>-<hash>`; its daemons serve it alone and are stopped and deleted with the worktree (`remove_worktree`, waiting for the daemon JVM to exit before deleting the home). The checkout itself keeps today's long-lived home.
- **Shared read-only dependency cache** (`GRADLE_RO_DEP_CACHE`): as a Forge process creates its first worktree, that worktree's build warms the checkout home once (`gradlew help --no-daemon`), and the home's `caches/modules-2` is published to `<checkout home>/ro-dep-cache` — artifacts hard-linked, indexes copied, lock/gc files left out, swapped in whole so no worktree reads a half-written copy. Worktrees download only what the copy lacks.
- **Orphan cleanup**: worktree homes whose worktree no longer exists are stopped and removed before publishing.
- The per-workspace `--stop` from #10098 is dropped; the stale-daemon rerun of the finalization gates stays, now scoped to the worktree's own daemons.
- Warming and publishing are best effort: a failure is reported and worktrees merely download more.

Spec first: §forge/FS-forge-run-requirements.4 is rewritten in its own commit; `forge/README.md` and `publication.md` follow.

## Verification

- `forge` unit tests: new `tests/test_gradle_dependency_cache.py` (11 tests), updated `test_gradle_environment.py` / `test_git_worktree_regressions.py`; the full suite has the same 13 pre-existing failures as `master`, nothing new.
- On a Forge host with Gradle 9.1: a fresh worktree resolves `gradlew help --offline` with **zero** downloads into its own home; its daemon does not appear in the checkout home's registry; removing the worktree leaves no home or daemon behind; a second worktree in the same process skips the preparation. Warm + publish took ~6–8 s on a warm checkout home.
- `grund check`, `fissile check --staged`, `pylint` clean.

## Cost

Each worktree pays one daemon start and its own transform/classloader caches (~20–30 s on its first build) instead of reusing a checkout-wide daemon; library-specific dependencies not in the published copy are downloaded once per worktree.


kimeta (approved):
Spec updated first; verified end to end on a Forge host (isolated daemons, zero downloads through the read-only cache, clean removal).

PR #10123, opened 2026-09-19
Auto PR: revert-spring-aop-6.1.7

graalvmbot:
## What this does

Reverts #10119 (merge commit `2c7336aecd`). Its `metadata/org.springframework/spring-aop/6.1.7/reachability-metadata.json` makes every job of the **Test affected Spring AOT smoke tests (4.1.x)** matrix (119 jobs) fail at `nativeCompile`:

```
Error: Type not found during analysis: HotSpotType<Lorg/springframework/aop/aspectj/AspectJMethodBeforeAdvice;, resolved>
##[error] Test :boot:actuator-webmvc failed, but it is PASSING upstream (success). This is a new regression!
```

The matrix had already failed on the PR head when Forge approved it and enabled auto-merge; see #10122 for the Forge-side defect. Reverting restores master's smoke tests; issue #9398 (`java run fails for spring-aop:6.1.7`) should be reopened so the fix is regenerated with the repaired metadata.

## Commits

- Revert "[GenAI] Test fix for org.springframework:spring-aop:6.1.7 using gpt-5.6-sol (#10119)"

Issue #10122, opened 2026-09-19
[bug] Forge approved and auto-merged PR #10119 while its Spring AOT smoke-test matrix (119 jobs) was already failing; CI repair result was lost to a descriptor error


cugurm:
## Summary

PR #10119 (`[GenAI] Test fix for org.springframework:spring-aop:6.1.7`) was merged to master on 2026-09-19 at 20:46 UTC although the **Test affected Spring AOT smoke tests (4.1.x)** matrix on its head `c68505328b` had finished with **119 failures** between 19:54 and 20:02 UTC. Every job fails identically at `nativeCompile`:

```
Error: Type not found during analysis: HotSpotType<Lorg/springframework/aop/aspectj/AspectJMethodBeforeAdvice;, resolved>
##[error] Test :boot:actuator-webmvc failed, but it is PASSING upstream (success). This is a new regression!
```

The only metadata change in the PR is `metadata/org.springframework/spring-aop/6.1.7/reachability-metadata.json`, so master's Spring AOT smoke tests are now red.

## What Forge did (cycle log, `do-work.sh --once --review-limit 4`)

```
[Approving descriptor-validated PR #10119 at c68505328b...; no semantic review agent launched.]
[Enabled auto-merge for approved PR #10119 at c68505328b...]
[Starting immediate contribution-scoped CI diagnosis for PR #10119.]
[CI repair session 71ce02b6dec34872; log: logs/org.springframework:spring-aop:6.1.7/pr-ci-repair/pi.log]
...
ERROR: Failed descriptor-driven processing for PR #10119: <ValidationError: "'' should be non-empty">
ERROR: Pull request processing failed for pull request(s): [10119]
```

Two defects compound here:

1. **Approval + auto-merge before the smoke matrix is considered.** The review step approves on descriptor validation and index checks alone and enables auto-merge immediately; because the Spring AOT smoke matrix is not a required check, GitHub merged as soon as the required checks were green — even though the matrix had *already* failed 40 minutes earlier. Forge itself noticed the failing CI right after ("immediate contribution-scoped CI diagnosis") — at that point auto-merge should have been disabled, or approval should have been withheld until the diagnosis concluded.
[...]

PR #10145, opened 2026-09-21
Gate approval and auto-merge on the whole rollup

graalvmbot:
## What this does

Forge decides *what* is mergeable; GitHub decides *when*. Today the second half
is delegated: `_process_descriptor_pull_request` approves the validated head and
enables auto-merge **before** it looks at CI, on the assumption that "GitHub
queues or merges only after every required CI and repository merge gate is
successful" — which holds only if the required-status-check set covers every
gate. It does not. `master` requires 5 contexts, and the Spring AOT smoke
aggregator is not among them, so #10119 merged at 20:46 with 119 smoke jobs
already red since 20:02 (#10122, reverted in #10123). Forge had read that
failure itself: the cycle log shows `Enabled auto-merge` and then
`Starting immediate contribution-scoped CI diagnosis` on the same head.

This makes Forge the whole gate. Nothing is approved, armed, or spent above the
check-rollup read:

| Rollup at the exact head | Before | After |
|---|---|---|
| Failed | approve, **arm**, then diagnose | disarm, then diagnose |
| Pending | approve, arm, index validation, wait | defer — unapproved, unarmed, nothing spent |
| Green | approve, arm | index validation, approve, arm |
| Conflicting head | approve, arm, then refresh | refresh only; the push earns a new approval |

Arming on a rollup Forge has personally seen green is safe whatever the
repository's required-check set happens to be, so §root/AR-required-status-checks
becomes a second, independent enforcement of the same rule rather than a
prerequisite for correctness.

Two side effects worth naming:

- **Less work per pass.** `validate_pull_request_indexes_before_merge` builds a
  worktree and runs index validation; it ran on every pass over a PR waiting on
  a ~1 h matrix, and now runs once, on a green head.
- **One more pass to merge** a PR that was still pending when reviewed. That
  pass is now a single rollup read.

The failed-CI branch calls `disable_pull_request_auto_merge` rather than
`ensure_pull_request_unapproved`: under the new invariant there is no Forge
approval to dismiss, and the disable is a no-op when `autoMergeRequest` is
`None`, so it costs nothing while still covering a head armed on green by an
earlier pass that later went red on a re-run.


kimeta (approved):
Approving at 898e6bb89ed7759fd3d49dcd503705e780bc5cc0.

The fix is correct and minimal. `approve_pull_request_from_descriptor` and
`enable_pull_request_auto_merge` have exactly one call site each in the whole
tree, and both now sit strictly below the rollup read, so "Forge arms nothing it
has not seen pass" holds globally rather than only on this path.

Verified:

- `statusCheckRollup.state` covers every check run on the commit, not just
  required contexts, and `has_failed_pull_request_ci` reads only `.state` — so
  the 100-context pagination cap does not weaken the gate.
- `autoMergeRequest` is in the GraphQL query and `state_cache` is rebuilt per
  cycle, so the disable short-circuit reads live state rather than a missing
  field.
- The conflict path is a strict improvement: the old code armed auto-merge with
  `--match-head-commit <old sha>` and then pushed a refreshed head, which GitHub
  would invalidate anyway.
- 178 dispatcher tests, 2 errors; the same 2 errors are present on master
  unchanged. `grund check` is clean.

Three non-blocking follow-ups, none of which need to hold this up:

1. The failed-CI branch disarms but leaves the Forge approval standing. The
   stated reason — that there is no approval to dismiss — does not hold in the
   scenario that justifies calling disable in the first place, since approve and
   arm are always set together. `repair_failed_ci_pull_request` covers this on
   the repair and reject paths, but the transient-rerun branch returns early
   without it, so a green rerun produces a second approval review at the same
   SHA. `ensure_pull_request_unapproved` is already imported here and is a
   superset of the current call.
2. `select_review_pull_requests` is unreferenced across the tree and now holds a
   stale second copy of the classification this PR replaces. Worth deleting
   while the contract is fresh.
3. The stage table says "seven stages" and has eight rows; the off-by-one
   predates this change.

Issue #10152, opened 2026-09-21
Benchmark drops all token metrics when one invocation times out


kimeta:
One agent invocation that dies without emitting usage nulls the token metrics
for the whole benchmark run.

`20260919T230609Z-l4-claude-code-opus-5-high-naive-99ad937f6a` is a **successful**
run with 20 cover passes and no token numbers at all:

```json
"total": {"tokens": {"input": null, "cachedInputRead": null,
                     "cachedInputWrite": null, "output": null}}
```

The 20th pass hit the 2h timeout and was killed, so Rhei recorded its tokens as
`"status": "unknown"`. The other 19 passes measured fine — their numbers are on
disk and were discarded.

The cause is `total()` in `forge/benchmarks/code_coverage_benchmark_metrics.py`:

```python
values = [_nested_token(invocation, *keys) for invocation in relevant]
if any(value is None for value in values):
    return None
```

One missing record out of twenty returns `None` for every token class. That
contradicts [§FS-code-coverage-benchmarking.4](https://github.com/oracle/graalvm-reachability-metadata/blob/master/forge/docs/functional-spec/benchmarking.md):
"Metrics that exist before the failure must still be retained."

The killed pass's tokens are genuinely unrecoverable — no usage was ever emitted
— so the fix is to sum the invocations that did report and keep `null` only for
a phase that measured nothing.

Timeouts will recur: in that run, pass durations ran 13s to 2h with two passes
already over 1h30m against the 2h cap.

The run is unusable for cost comparison today; §FS-forge-generation-benchmarking.6
disqualifies a run that "lost token, cost, iteration, LOC, coverage,
dynamic-access, or metadata metrics that the schema can represent."

PR #10155, opened 2026-09-21
Keep measured phase tokens when an invocation has no usage

kimeta:
## What this does

One agent invocation that dies without emitting usage nulled the token metrics
for its whole phase — and, through the phase sums, for the run. `total()` in
`_phase_tokens` treated a single unmeasured invocation as poisoning every
measured one:

```python
if any(value is None for value in values):
    return None
```

Run `20260919T230609Z-l4-claude-code-opus-5-high-naive-99ad937f6a` hit exactly
this: its 20th cover pass was killed at the 2h timeout before any usage was
emitted, so Rhei recorded that invocation's tokens as `"status": "unknown"` and
all four published token totals came out `null` — although the other 19 passes
measured fine and their numbers are on disk. Fixes #10152.

The killed invocation's tokens were never recorded anywhere, so they are
genuinely unrecoverable. The fix sums the invocations that did report and keeps
`null` only for a phase in which nothing measured at all — retaining the
metrics that exist before the failure, as
[§forge/FS-code-coverage-benchmarking.4](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto-keep-measured-tokens/forge/docs/functional-spec/benchmarking.md)
requires.

## Why it stacks on #10086

Both changes rewrite the same `total()` / `_phase_tokens` path, and this one
composes with the disjoint-input restatement: the partial sums feed
`_ordinary_input` the same way complete sums do, and its `total >= cached`
guard already tolerates token classes summed over different invocation
subsets.

## What this does not do

The published record for the affected run stays `null`: a written record is
immutable for its runId
([§forge/FS-code-coverage-benchmarking.3](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto-keep-measured-tokens/forge/docs/functional-spec/benchmarking.md)).
Restating it from the preserved workspace is a data edit under the run-ID
conflict rule, same as the restatement already deferred from #10086.


kimeta:
Superseded by #10177, which carries this commit alongside #10086's, both rebased onto master.

This PR's first commit — the test-file split — is gone: master's `faad895095` had already cut the same seam, moving `CodeCoverageBenchmarkLifecycleTests` and `CodeCoverageBenchmarkMetricsTests` into `test_code_coverage_benchmark_runs.py`, so after the rebase it was pure duplication. The fix itself now applies to `total()` in `code_coverage_benchmark_metrics.py`, and `benchmark._collect_result` became the public `benchmark.collect_result` in that same split.

#10152 stays open until the fix lands in #10177.

PR #10157, opened 2026-09-22
[GenAI] Add support for io.micronaut:micronaut-http-client:5.1.15 using gpt-5.6-sol

graalvmbot:
## What does this PR do?

Fixes: #10140

This PR introduces tests and metadata for io.micronaut:micronaut-http-client:5.1.15, enabling support for this library.

Summary:
- Strategy: optimistic_dynamic_access_iterative_pi_gpt-5.6-sol
- Agent: pi
- Model: gpt-5.6-sol
- Input tokens: 218365
- Cached input tokens: 3646848
- Output tokens: 25069
- Metadata entries: 103
- Test-only metadata entries: 19
- Iterations: 9
- Library coverage percentage: 34.98
- Generated lines of code: 290
- Tested library lines of code: 1301

### Forge

- Forge monitored branch: `origin/master`
- Forge branch: `master`
- Forge commit hash: `d6c776fdfa805b9e417995bcf5719ca4fd7be8c4`

Stats from `stats/<groupId>/<artifactId>/<metadata-version>/stats.json`:

Dynamic access coverage:
- Overall: 0/0 covered calls (100.00%)

Library coverage:
- Instruction: 5775/16669 (34.65%)
- Line: 1300/3719 (34.96%)
- Method: 331/912 (36.29%)

### Local CI Verification

- Status: `success`
- Commands run: 2
- Fixup attempts: 0

## Local Agent Review

- Decision: `approved`
- Model: `gpt-5.6-sol`
- Review session: `389693261d604a2e`

Reviewed the single-coordinate diff and local evidence under the library-new-request rules. The contribution stays within the allowed io.micronaut:micronaut-http-client:5.1.15 metadata, stats, and test paths; uses only reachability-metadata.json; and has a separate-package, public-API test suite with bounded timeouts, no scaffold placeholder, native-image skip, swallowed failure, version-pinned assertion, visibility bypass, compatibility override, or Native Image flag. The reported dynamic-access result is the valid 0/0 no-calls case, so percentage and entry-count mismatch gates do not apply. Index validation evidence passed, and the stable finalization pass completed all three native lanes, metadata splitting and validation, style checks, generated-test checks, and statistics generation without further publishable changes.

**Finding — Generated library statistics were stale**

[...]

PR #10175, opened 2026-09-22
Re-poll unsettled PR mergeability before classification

### LINKED ISSUE #10174: Dispatcher approves conflicting PRs when mergeability reads UNKNOWN
GitHub computes a pull request's `mergeable` field lazily: every push to the base branch invalidates it for all open PRs, and the first query afterwards returns `UNKNOWN` while a background job recomputes it. The dispatcher treats only the literal `CONFLICTING` as conflicting, so an `UNKNOWN` read falls through the not-conflicting path.

Because the review pass merges PRs while classifying the rest of the queue, it reliably queries mergeability seconds after moving the base — exactly when the answer is `UNKNOWN`. A conflicting PR with a green rollup is then approved and armed for auto-merge instead of being conflict-refreshed, and auto-merge waits forever on a `DIRTY` head. The PR starves until a later pass happens to read the settled state.

Observed on #10157: approved and armed at 2026-09-22 11:33:13 UTC, between the merges of #10170 (11:32:57) and #10169 (11:34:08), while the PR had been continuously conflicting since creation. #10146 collected three approvals the same way over ~21 hours.

The fix is to treat `UNKNOWN` as an unanswered question: re-poll the state fetch with bounded backoff until GitHub settles the answer, and only then classify the head as conflicting or not.

kimeta:
Closes #10174

## What this does

GitHub computes `mergeable` lazily: any base-branch push resets it to `UNKNOWN` for every open PR until a background job — triggered by the next read — settles it. The dispatcher classifies a head as conflicting only on the literal `CONFLICTING`, so a read taken seconds after the pass itself merged a sibling sees `UNKNOWN`, skips conflict refresh, and (with a green rollup) approves and arms auto-merge on a head that is actually `DIRTY`. Auto-merge then waits forever; the PR starves until a later pass happens to read the settled state. #10157 was approved and armed this way at 11:33:13 UTC, between two sibling merges at 11:32:57 and 11:34:08, while continuously conflicting since creation.

`get_pull_request_state` now treats `UNKNOWN` as an unanswered question and re-polls with bounded backoff before returning, so every downstream decision — conflict refresh before CI, approval only on a settled clean head — runs on a real answer. The decision ladder itself is unchanged, per [§forge/FS-automated-pr-review](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/retry-unknown-mergeability/forge/docs/functional-spec/publication.md), which this PR extends with the corresponding sentence.

## How the re-poll works

```
fetch            → settled? return   (common case: one fetch, no sleep)
sleep 2s → fetch → settled? return
sleep 4s → fetch → settled? return
sleep 8s → fetch → return as-is
```

A settled answer costs exactly one fetch, identical to today. A still-`UNKNOWN` answer after the last delay falls through to the existing behavior, so the worst case is 14 s and three extra reads for one PR.


kimeta:
Superseded by #10176, which carries the same commits plus the fix for approving a head whose mergeability never settled. Reopened from mm/auto-retry-unknown-mergeability so graalvmbot is the author and I stay eligible to review.

PR #10177, opened 2026-09-22
Auto PR: benchmark-token-accounting

### LINKED ISSUE #10152: Benchmark drops all token metrics when one invocation times out

graalvmbot:
## What this does

Two defects in `_phase_tokens` corrupt published benchmark token metrics in
opposite ways: one makes `input` mean two different things across records, the
other throws away a whole phase's accounting over a single missing record. Both
sit on the same `total()` path, so they land together.

Supersedes #10086 and #10155, which were stacked on `mm/auto` and predate
master's `faad895095` source split; the change now lives in
`code_coverage_benchmark_metrics.py`, where that commit moved `_phase_tokens`.

## 1. Ordinary input must stay disjoint from cached input

`_phase_tokens` published Rhei's `tokens.input.total` verbatim as `input`. Rhei
reports that total **inclusive** of the cached classes on newer runtimes and
**exclusive** of them on older ones, so `input` silently means two different
things depending on when a run was recorded — while the spec requires one
meaning:

> Cached input must remain separate from ordinary input.
>
> — [§forge/FS-code-coverage-benchmarking.4](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto-benchmark-token-accounting/forge/docs/functional-spec/benchmarking.md)

`_ordinary_input` subtracts the cached classes from a total that contains them,
so ordinary input stays disjoint from `cachedInputRead` and `cachedInputWrite`
in every record.

The drift is not theoretical. Across the twelve runs recorded for
`com.h2database:h2:2.1.210`, the six older entries report `cachedInputRead`
larger than `input` — impossible if the cached reads were inside that total —
and the six newer ones report it smaller:

| Recorded | `input` | `cachedInputRead` | Reads as |
|---|---:|---:|---|
| sol, high, guided | 6,459,340 | 288,110,976 | exclusive |
| sol, medium, guided | 52,311,342 | 49,524,224 | inclusive |

Comparing those two `input` values as-is suggests the guided arm spent 8x fewer
input tokens. On the disjoint basis it is 6,459,340 against 2,787,118 — the
opposite direction.

### How the two shapes are told apart

The invocation record carries no runtime version, so the arithmetic decides: a
total at least as large as its cached classes contains them and is restated; a
smaller total already excludes them and is published as recorded. That keeps
both runtimes correct without pinning a version, and leaves `cachedInputRead`,
`cachedInputWrite`, and `output` untouched — only `input` ever changed meaning.

## 2. A phase must keep the tokens it did measure

One agent invocation that dies without emitting usage nulled the token metrics
for its whole phase — and, through the phase sums, for the run. `total()`
treated a single unmeasured invocation as poisoning every measured one:

```python
if any(value is None for value in values):
    return None
```

Run `20260919T230609Z-l4-claude-code-opus-5-high-naive-99ad937f6a` hit exactly
this: its 20th cover pass was killed at the 2h timeout before any usage was
emitted, so Rhei recorded that invocation's tokens as `"status": "unknown"` and
all four published token totals came out `null` — although the other 19 passes
measured fine and their numbers are on disk. Fixes #10152.

The killed invocation's tokens were never recorded anywhere, so they are
genuinely unrecoverable. `total()` now sums the invocations that did report and
keeps `null` only for a phase in which nothing measured at all — retaining the
metrics that exist before the failure, as
[§forge/FS-code-coverage-benchmarking.4](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto-benchmark-token-accounting/forge/docs/functional-spec/benchmarking.md)
requires.

## How the two compose

The partial sums feed `_ordinary_input` the same way complete sums do, and its
`total >= cached` guard already tolerates token classes summed over different
invocation subsets. A phase that measured nothing still yields `null` for every
class, and `_ordinary_input` passes `None` straight through.

## What this does not do

Neither fix restates a record that is already published. A written record is
immutable for its runId
([§forge/FS-code-coverage-benchmarking.3](https://github.com/oracle/graalvm-reachability-metadata/blob/mm/auto-benchmark-token-accounting/forge/docs/functional-spec/benchmarking.md)),
so correcting the existing entries is a data edit under the run-ID conflict
rule, deliberately kept out of a behavior fix. Two restatements are now
outstanding:

1. **The run from §2 stays `null`.** Its other 19 passes are on disk in the
   preserved workspace and are recoverable.
2. **No published record has ever carried a non-zero cache write.**
   `cachedInputWrite` only exists on records written after #10027, and those are
   all `pi` runs, where it is legitimately `0` — codex has no separate
   cache-write class. The two `claude-code` runs, which do have cache writes,
   predate the field.

The preserved Rhei accounting still has those numbers, and they are not small:

| Run | published `input` | `cache_write` on disk | fresh input |
|---|---:|---:|---:|
| opus-5, medium | 6,142 | 6,030,864 | 6,037,006 |
| opus-5, high | 7,850 | 7,830,250 | 7,838,100 |
| sol, high (pi) | 6,459,340 | 0 | 6,459,340 |

Anthropic books a turn's new content as `cache_creation` where codex books it as
plain `input`, so the published figures understate those two runs by roughly
1000x on the input side — and cache writes are the *premium* input class
(1.25x base at 5-minute TTL, 2x at 1-hour), not a discounted one. They are
recoverable from
`forge/local_repositories/code_coverage_benchmarks/<runId>/code-coverage-99000/runtime/accounting/invocations/`.

## Open question

**Is subtracting `cache_write` from an inclusive total right?** It is subtracted
here for symmetry with `cached_read`, but no post-#10027 `claude-code` run
exists yet, so the inclusive shape has never been observed with a non-zero cache
write. In the one pre-#10027 Anthropic sample, `input.total` (432) excludes both
classes. The `total >= cached` guard means a wrong guess here cannot corrupt an
exclusive record, but it should be confirmed against a real inclusive Anthropic
record before anyone leans on that column.

## Commits

- Keep benchmark input tokens disjoint from cached input
- Keep measured phase tokens when an invocation has no usage

---

Authored by @kimeta on `mm/auto-benchmark-token-accounting`, opened by
`graalvmbot` so the author can review it.


kimeta (approved):
Both commits are the rebase of #10086 and #10155 onto master, relocated into `code_coverage_benchmark_metrics.py` after the `faad895095` source split. Forge benchmark suite passes (42 tests), `grund check` and `fissile check --staged` are clean.

